PR #4, opened 2019-10-12
dts: msm8916-samsung: Add J5 2016 J510FN

Kiciuk:


dts/msm8916-samsung.dts
```diff
@@ -23,4 +23,11 @@
 		compatible = "samsung,a5u-eur", "qcom,msm8916", "lk2nd,device";
 		lk2nd,match-bootloader = "A500FU*";
 	};
+	
+	j5xnlte-eur {
```
Minecrell: Where does the `n` in `j5xnlte` come from? Is that just in the Android codename? Because in downstream, the device tree is called `msm8916-sec-j5xlte-eur-r04.dts` as far as I can tell (without `n`)

Minecrell:
FYI, I merged https://github.com/msm8916-mainline/lk2nd/pull/6 which added the same device (now with working USB?).

PR #7, opened 2019-12-01
dts: Add Galaxy Tab A 8.0 (SM-T357W)

Armen138:
Added dts file for Galaxy Tab A 8.0, Canadian LTE version SM-T357W

### DIFF
--- a/dts/msm8916-samsung-r06.dts
+++ b/dts/msm8916-samsung-r06.dts
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+/include/ "msm8916.dtsi"
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <206 0>;
+	qcom,board-id = <0xCE08FF01 6>;
+
+	gt58ltebmc {
+		model = "Samsung Galaxy Tab A 8.0 (CAN)";
+		compatible = "samsung,gt58ltebmc", "qcom,msm8916", "lk2nd,device";
+		lk2nd,match-bootloader = "T357W*";
+	};
+};
--- a/dts/rules.mk
+++ b/dts/rules.mk
@@ -7,5 +7,6 @@ DTBS += \
 	$(LOCAL_DIR)/msm8916-motorola-harpia.dtb \
 	$(LOCAL_DIR)/msm8916-samsung-r01.dtb \
 	$(LOCAL_DIR)/msm8916-samsung-r03.dtb \
-	$(LOCAL_DIR)/msm8916-samsung-r04.dtb
+	$(LOCAL_DIR)/msm8916-samsung-r04.dtb \
+	$(LOCAL_DIR)/msm8916-samsung-r06.dtb
 endif


dts/msm8916-samsung-r06.dts
```diff
@@ -0,0 +1,22 @@
...
+/include/ "msm8916.dtsi"
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <206 0>;
+	qcom,board-id = <0xCE08FF01 6>;
+
+	gt58ltebmc {
+		model = "Samsung Galaxy Tab A 8.0";
+        compatible = "samsung,gt58ltebmc", "qcom,msm8916-mtp", "qcom,msm8916", "qcom,mtp", "lk2nd,device";
```
Minecrell: the indentation looks weird here - please make sure you are using tabs.

dts/msm8916-samsung-r06.dts
```diff
@@ -0,0 +1,22 @@
...
+		lk2nd,match-bootloader = "T35*";
+
+		samsung,muic-reset {
```
Minecrell: I don't think you need this workaround for your device.

dts/msm8916-samsung-r06.dts
```diff
@@ -0,0 +1,22 @@
...
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <206 0>;
+	qcom,board-id = <0xCE08FF01 6>;
+
+	gt58ltebmc {
+		model = "Samsung Galaxy Tab A 8.0";
+        compatible = "samsung,gt58ltebmc", "qcom,msm8916-mtp", "qcom,msm8916", "qcom,mtp", "lk2nd,device";
+		lk2nd,match-bootloader = "T35*";
```
Minecrell: Please make this as strict as possible i.e. `T357W*` if that works. lk2nd displays the full bootloader string in the fastboot menu, so you can check it there.

dts/msm8916-samsung-r06.dts
```diff
@@ -0,0 +1,22 @@
...
+
+/include/ "msm8916.dtsi"
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <206 0>;
+	qcom,board-id = <0xCE08FF01 6>;
+
+	gt58ltebmc {
+		model = "Samsung Galaxy Tab A 8.0";
```
Minecrell: Indicating that this is the Canadian variant might be helpul. So either something like 

```suggestion
		model = "Samsung Galaxy Tab A 8.0 (CAN)";
```

or 

```suggestion
		model = "Samsung Galaxy Tab A 8.0 (SM-T357W)";
```

Just make sure that it fits on one line in the fastboot screen.

dts/msm8916-samsung-r06.dts
```diff
@@ -0,0 +1,17 @@
...
+/include/ "msm8916.dtsi"
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <206 0>;
+	qcom,board-id = <0xCE08FF01 6>;
+
+	gt58ltebmc {
+		model = "Samsung Galaxy Tab A 8.0 (CAN)";
+		compatible = "samsung,gt58ltebmc", "qcom,msm8916-mtp", "qcom,msm8916", "qcom,mtp", "lk2nd,device";
```
Minecrell: Like the `model`, this does not actually need to match what is used in downstream. Therefore you definitely don't need the `mtp` compatibles, because your tablet actually has nothing in common with MTP (a reference development board with MSM8916)

```suggestion
		compatible = "samsung,gt58ltebmc", "qcom,msm8916", "lk2nd,device";
```

Not sure about `samsung,gt58ltebmc`. I guess I would have called it `samsung,gt58lte-can` or `samsung,gt58lte-canbmc` like the downstream device tree (`msm8916-sec-gt58lte-canbmc-r06.dts`). `gt58ltebmc` is the Android codename as far as I can tell so that would be fine too.

We have no obligation to follow any particular name here, so it's up to you which one you would prefer using to refer to your device in the future. You can also keep it as is if you want.

PR #10, opened 2020-02-13
lk2nd: Write unitinfo to smem for motorola devices

Sorixelle:


app/aboot/lk2nd-device.c
```diff
@@ -118,13 +134,19 @@ static int lk2nd_find_device_offset(const void *fdt)
...
 	if (offset < 0) {
 		dprintf(CRITICAL, "Failed to find matching lk2nd,device node: %d\n", offset);
 		return;
 	}
 
 	lk2nd_samsung_muic_reset(fdt, offset);
 
+	fdt_getprop(fdt, offset, "lk2nd,motorola-unitinfo", &len);
+	if (len > 0) {
+		lk2nd_motorola_smem_write_unit_info(fdt);
```
Minecrell: Can you check the prop inside `lk2nd_motorola_smem_write_unit_info` (just like `lk2nd_samsung_muic_reset` does)?

app/aboot/lk2nd-motorola.h
```diff
@@ -0,0 +1,26 @@
+#ifndef _LK2ND_MOTOROLA
+#define _LK2ND_MOTOROLA
+
+#define MMI_UNIT_INFO_VER 3
```
Minecrell: I think it would be better to use the header as-is from https://github.com/msm8916-mainline/android_kernel_qcom_msm8916/blob/harpia/drivers/soc/qcom/mmi-unit-info.h, without modifications.
That preserves the license header. You could consider removing `mmi_set_pureason` from the header, but on the other hand it shouldn't hurt if you keep it in there (unused).

dts/carrier-channel-ids.dtsi
```diff
@@ -0,0 +1,273 @@
+/* Copyright (c) 2013, The Linux Foundation. All rights reserved.
```
Minecrell: Rename this file to something like `(msm8916-)?motorola-carrier-channel-ids.dtsi`

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,63 @@
...
+	memset(unit_info, 0, 1024);
+
+	int chosen_offset = fdt_path_offset(fdt, "/chosen");
+	const uint32_t *system_rev = fdt_getprop(fdt, chosen_offset, "linux,hwrev", NULL);
+	const uint32_t *system_serial_low = fdt_getprop(fdt, chosen_offset, "linux,seriallow", NULL);
+	const uint32_t *system_serial_high = fdt_getprop(fdt, chosen_offset, "linux,serialhigh", NULL);
+	const uint32_t *powerup_reason = fdt_getprop(fdt, chosen_offset, "mmi,powerup_reason", NULL);
+
+	int len = 0;
+	const char *baseband = fdt_getprop(fdt, chosen_offset, "linux,hwrev", &len);
```
Minecrell: Oops! Look at the property name :)

app/aboot/lk2nd-device.h
```diff
@@ -10,6 +10,13 @@ struct lk2nd_device {
 	const char *cmdline;
 	const char *bootloader;
 	const char *model;
+
+	// Properties used for motorola unit info
+	const char *barcode;
```
Minecrell:
```suggestion
	const char *serialno;
```

app/aboot/lk2nd-device.c
```diff
@@ -54,6 +54,22 @@ static void parse_boot_args(void)
...
+		}
+		if ((val = strpresuf(arg, "androidboot.device="))) {
+			lk2nd_dev.device = strdup(val);
+		}
+		if ((val = strpresuf(arg, "androidboot.radio="))) {
+			lk2nd_dev.radio = atoui(val);
+			lk2nd_dev.radio_str = "";
+			if (lk2nd_dev.radio == 0) {
+				lk2nd_dev.radio_str = strdup(val);
+			}
```
Minecrell: The integer stuff should be handled in `lk2nd-motorola.c`, here you should just set it to `lk2nd_dev.radio` as-is (as string).

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,63 @@
...
+#include <debug.h>
+#include <libfdt.h>
+#include <malloc.h>
+#include <string.h>
+
+#include "lk2nd-device.h"
+#include "lk2nd-motorola.h"
+#include "smem.h"
+
+static struct mmi_unit_info *unit_info;
```
Minecrell: No need to have this as a global variable.

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,63 @@
...
+
+void lk2nd_motorola_smem_write_unit_info(const void *fdt)
+{
+	int ret = 0;
+	unit_info = malloc(1024);
+	memset(unit_info, 0, 1024);
```
Minecrell: Use a define like in the kernel driver:

```c
#define SMEM_KERNEL_RESERVE_SIZE 1024
```

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,63 @@
...
+		unit_info->device,
+		unit_info->radio,
+		unit_info->radio_str,
+		unit_info->system_rev,
+		unit_info->system_serial_high, unit_info->system_serial_low,
+		unit_info->machine, unit_info->barcode,
+		unit_info->baseband, unit_info->carrier,
+		unit_info->powerup_reason);
+
+	ret = smem_alloc_write_entry(0x86, unit_info, 1024);
```
Minecrell:
```c
#define SMEM_ID_VENDOR0		0x86
#define SMEM_KERNEL_RESERVE	SMEM_ID_VENDOR0
```

(to make it clear which smem "slot" is used)

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,63 @@
...
+	if (ret != 0) {
+		dprintf(INFO, "Failed to write motorola unit info to smem: %d\n", ret);
+        return;
+	}
+
+	dprintf(INFO, "Wrote motorola unit info\n");
```
Minecrell: Look at the formatting of the return ^.

Also leaking memory :(

```c
free(unit_info);
```

app/aboot/lk2nd-device.h
```diff
@@ -10,6 +10,12 @@ struct lk2nd_device {
 	const char *cmdline;
 	const char *bootloader;
 	const char *model;
+
+	// Properties used for motorola unit info
```
Minecrell: now it's time to remove this comment and to move these properties directly below bootloader. This is not supposed to be too Motorola-specific.

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,77 @@
...
+	const char *baseband = fdt_getprop(fdt, chosen_offset, "mmi,baseband", &len);
+	if (len == -1) {
+		baseband = "";
+	}
+
+	unit_info->version = MMI_UNIT_INFO_VER;
+	unit_info->system_rev = fdt32_to_cpu(*system_rev);
+	unit_info->system_serial_low = fdt32_to_cpu(*system_serial_low);
+	unit_info->system_serial_high = fdt32_to_cpu(*system_serial_high);
+	strlcpy(unit_info->machine, "", MACHINE_MAX_LEN);
```
Minecrell: unneeded, `unit_info` is zeroed, so this will be a no-op.

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,77 @@
...
+	strlcpy(unit_info->barcode, lk2nd_dev.serialno, BARCODE_MAX_LEN);
+	strlcpy(unit_info->baseband, baseband, BASEBAND_MAX_LEN);
+	strlcpy(unit_info->carrier, lk2nd_dev.carrier, CARRIER_MAX_LEN);
+	strlcpy(unit_info->device, lk2nd_dev.device, DEVICE_MAX_LEN);
```
Minecrell: all of these are somewhat unsafe, it will crash if they are not set on command line.

Maybe something like:

```c
if (lk2nd_dev.serialno)
	strlcpy(unit_info->barcode, lk2nd_dev.serialno, BARCODE_MAX_LEN);
```

etc.

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,77 @@
...
+		unit_info->device,
+		unit_info->radio,
+		unit_info->radio_str,
+		unit_info->system_rev,
+		unit_info->system_serial_high, unit_info->system_serial_low,
+		unit_info->machine, unit_info->barcode,
+		unit_info->baseband, unit_info->carrier,
+		unit_info->powerup_reason);
+
+	int ret = 0;
```
Minecrell: lk kind of follows ancient C code standards where all variables are declared on top of the method. (like in the kernel)

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,77 @@
...
+
+	const char *baseband = fdt_getprop(fdt, chosen_offset, "mmi,baseband", &len);
+	if (len == -1) {
+		baseband = "";
+	}
+
+	unit_info->version = MMI_UNIT_INFO_VER;
+	unit_info->system_rev = fdt32_to_cpu(*system_rev);
+	unit_info->system_serial_low = fdt32_to_cpu(*system_serial_low);
+	unit_info->system_serial_high = fdt32_to_cpu(*system_serial_high);
```
Minecrell: This will crash if the properties are missing in the device tree. Here a helper method might be useful:

```c
static inline void fdt_getprop_u32(const void *fdt, int offset, const char *name, uint32_t *val) {
	int len;
	const uint32_t *prop = fdt_getprop(fdt, chosen_offset, "linux,hwrev", &len);
	if (len == sizeof(*val))
		*val = fdt32_to_cpu(*prop);
}
```

```c
fdt_getprop_u32(fdt, chosen_offset, "linux,hwrev", &unit_info->system_rev);
```

etc

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,77 @@
...
+	fdt_getprop(fdt, offset, "lk2nd,motorola-unitinfo", &len);
+	if (len == 0) {
+		return;
+	}
+
+	struct mmi_unit_info *unit_info;
+	unit_info = malloc(SMEM_KERNEL_RESERVE_SIZE);
+	memset(unit_info, 0, SMEM_KERNEL_RESERVE_SIZE);
+
+	int chosen_offset = fdt_path_offset(fdt, "/chosen");
```
Minecrell: Check that `chosen_offset >= 0`, otherwise log an error and return.

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,77 @@
...
+	unit_info->system_serial_high = fdt32_to_cpu(*system_serial_high);
+	strlcpy(unit_info->machine, "", MACHINE_MAX_LEN);
+	strlcpy(unit_info->barcode, lk2nd_dev.serialno, BARCODE_MAX_LEN);
+	strlcpy(unit_info->baseband, baseband, BASEBAND_MAX_LEN);
+	strlcpy(unit_info->carrier, lk2nd_dev.carrier, CARRIER_MAX_LEN);
+	strlcpy(unit_info->device, lk2nd_dev.device, DEVICE_MAX_LEN);
+	unit_info->powerup_reason = fdt32_to_cpu(*powerup_reason);
+
+	const uint32_t radio = atoui(lk2nd_dev.radio);
+	unit_info->radio = radio;
```
Minecrell: can write this directly to `unit_info->radio`, no need for the variable

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,88 @@
...
+		return;
+	}
+
+	struct mmi_unit_info *unit_info;
+	unit_info = malloc(SMEM_KERNEL_RESERVE_SIZE);
+	memset(unit_info, 0, SMEM_KERNEL_RESERVE_SIZE);
+
+	int chosen_offset = fdt_path_offset(fdt, "/chosen");
+	if (chosen_offset < 0) {
+		dprintf(INFO, "Failed to get /chosen node in device tree");
```
Minecrell:
```suggestion
		dprintf(CRITICAL, "Failed to get /chosen node in device tree\n");
```

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,88 @@
...
+
+void lk2nd_motorola_smem_write_unit_info(const void *fdt, int offset)
+{
+	int len, ret = 0;
+	fdt_getprop(fdt, offset, "lk2nd,motorola-unitinfo", &len);
+	if (len == 0) {
+		return;
+	}
+
+	struct mmi_unit_info *unit_info;
```
Minecrell: move to top

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,88 @@
...
+	unit_info = malloc(SMEM_KERNEL_RESERVE_SIZE);
+	memset(unit_info, 0, SMEM_KERNEL_RESERVE_SIZE);
+
+	int chosen_offset = fdt_path_offset(fdt, "/chosen");
```
Minecrell: move declaration to top

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,88 @@
...
+		return;
+	}
+
+	unit_info->version = MMI_UNIT_INFO_VER;
+
+	const char *baseband = fdt_getprop(fdt, chosen_offset, "mmi,baseband", &len);
+	if (len < 0) {
+		baseband = "";
+	}
+	strlcpy(unit_info->baseband, baseband, BASEBAND_MAX_LEN);
```
Minecrell: Move declaration of `baseband` to top. Also, `unit_info->baseband` is empty by default since it's zeroed. Therefore you can change the if condition to `len > 0` and move the `strlcpy` into the if statement.

Also, we may read beyond the end of the property if the string is not null terminated for stupid reasons. I think you can avoid that using:

```c
strlcpy(unit_info->baseband, baseband, MIN(len, BASEBAND_MAX_LEN));
```

(note the `MIN`)

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,88 @@
...
+		strlcpy(unit_info->carrier, lk2nd_dev.carrier, CARRIER_MAX_LEN);
+	if (lk2nd_dev.device)
+		strlcpy(unit_info->device, lk2nd_dev.device, DEVICE_MAX_LEN);
+
+	fdt_getprop_u32(fdt, chosen_offset, "linux,hwrev", &unit_info->system_rev);
+	fdt_getprop_u32(fdt, chosen_offset, "linux,seriallow", &unit_info->system_serial_low);
+	fdt_getprop_u32(fdt, chosen_offset, "linux,serialhigh", &unit_info->system_serial_high);
+	fdt_getprop_u32(fdt, chosen_offset, "mmi,powerup_reason", &unit_info->powerup_reason);
+
+	unit_info->radio = atoui(lk2nd_dev.radio);
```
Minecrell: `if (lk2nd_dev.radio) {` ... (null check)

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,88 @@
...
+	if (unit_info->radio == 0) {
```
Minecrell: If I'm reading the downstream code correctly the string is always copied, so you could avoid this if statement: https://github.com/msm8916-mainline/android_kernel_qcom_msm8916/blob/fc0452207ebb9db007ac5b08239a2a352ac27a16/drivers/soc/qcom/mmi-unit-info.c#L70-L76

app/aboot/lk2nd-motorola.c
Minecrell: this still exists. also

`unit_info = malloc(SMEM_KERNEL_RESERVE_SIZE);` below this.

Then you don't need to bother with freeing `unit_info`.

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,89 @@
...
+#define SMEM_KERNEL_RESERVE      SMEM_ID_VENDOR0
+#define SMEM_KERNEL_RESERVE_SIZE 1024
+
+static void fdt_getprop_u32(const void *fdt, int offset, const char *name, uint32_t *val) {
+	int len;
+	const uint32_t *prop = fdt_getprop(fdt, offset, name, &len);
+	if (len == sizeof(*val))
+		*val = fdt32_to_cpu(*prop);
+	else
+		dprintf(CRITICAL, "Failed to get prop %s in device tree", name);
```
Minecrell:
```suggestion
		dprintf(CRITICAL, "Failed to get prop %s in device tree\n", name);
```

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,89 @@
...
+void lk2nd_motorola_smem_write_unit_info(const void *fdt, int offset)
+{
+	int chosen_offset, len, ret = 0;
+	const char *baseband;
+	struct mmi_unit_info *unit_info;
+
+	fdt_getprop(fdt, offset, "lk2nd,motorola-unitinfo", &len);
+	if (len == 0) {
+		return;
+	}
```
Minecrell: either use brackets for simple if statements or don't, but not both in one method please :)

app/aboot/lk2nd-motorola.c
```diff
@@ -0,0 +1,89 @@
...
+
+	fdt_getprop_u32(fdt, chosen_offset, "linux,hwrev", &unit_info->system_rev);
+	fdt_getprop_u32(fdt, chosen_offset, "linux,seriallow", &unit_info->system_serial_low);
+	fdt_getprop_u32(fdt, chosen_offset, "linux,serialhigh", &unit_info->system_serial_high);
+	fdt_getprop_u32(fdt, chosen_offset, "mmi,powerup_reason", &unit_info->powerup_reason);
+
+	if (lk2nd_dev.radio) {
+		unit_info->radio = atoui(lk2nd_dev.radio);
+		strlcpy(unit_info->radio_str, lk2nd_dev.radio, RADIO_MAX_LEN);
+	}
```
Minecrell: the order here is a bit random, maybe do

- copy from lk2nd_dev stuff
- fdt_getprop_u32
- baseband

PR #12, opened 2020-02-17
target: msm8916: Fix rules.mk

Roker2:
Last file need to be without \, because it is last.

### DIFF
--- a/target/msm8916/rules.mk
+++ b/target/msm8916/rules.mk
@@ -32,7 +32,7 @@ DEFINES += \
 
 OBJS += \
 	$(LOCAL_DIR)/init.o \
-	$(LOCAL_DIR)/meminfo.o \
+	$(LOCAL_DIR)/meminfo.o
 
 ifneq ($(DISPLAY_USE_CONTINUOUS_SPLASH),1)
 OBJS += \


Roker2:
I port your commits to nougat bootloader for msm8953 and find this mistake.

Minecrell:
Hi @Roker2! Thanks for the PR!

This change should not be necessary:
  - The backslash (`\`) is followed by an empty line. This is a common approach to avoid having to change more than a line when you want to append another file to the end of the list.
  - You should not use `target/msm8916` for another SoC, it's only for MSM8916.

Can you explain which error you get exactly?

Roker2:
Now I didn't try to build msm8953 nougat bootloader with your commits.
And if you think, that additional backslash is not mistake... Okay, no problems.

I have one question.
Will the code for msm8916 work for msm8953?

https://github.com/Roker2/kernel_lk/commit/31becefc948ae9a56de1cddf94dcedbef8d50e89
https://github.com/Roker2/kernel_lk/commit/2c6acb8fe99174aedcccfb4cb88bb832cf8f39c1

Junak:
@Roker2 i also did re-base lk2nd for msm8953 [here](https://github.com/Junak/lk2nd).
> Roker2/kernel_lk@31becef

This commit only disables display. You may want to add the rest from original commit. Also some panels require more [code](https://github.com/Junak/lk2nd/commit/f4f4241271a46d7033f6bf5640541d4e8b98538f) to trigger update. 
> Roker2/kernel_lk@2c6acb8

It should work (even without it).

Roker2:
@Junak, did you try to build your fork and boot it?

Junak:
Yes

Roker2:
@Junak, I can't to build it for mido.
Log
https://gist.github.com/Roker2/e1177861ac4d33641740c233568ad8ba
dts
https://gist.github.com/Roker2/3cc90acb1c339dc0496cb1f6b9d98ce4

Roker2:
Okay, I booted Mainline bootloader by @Junak. What is the next step?
Can I adapt mido 3.18 kernel for mainline bootloader? If yes, how?

Roker2:
Okay, adapting is uneeded, it can our kernel without changes.
It mean, that mido kernel developers can go to try to port 4.9 kernel!

Minecrell:
I think we agree that the changes in this PR seem to be not necessary, so I will close this PR. Thanks again!

PR #14, opened 2020-04-11
Add msm8939-based Alcatel Idol 3 (5.5) (6045)

vknecht:


dts/msm8939-alcatel-idol3.dts
```diff
@@ -0,0 +1,13 @@
...
+/dts-v1/;
+
+/include/ "msm8916.dtsi"
+
+/ {
+	qcom,msm-id = <239 0>;
+	qcom,board-id = <8 0>;
+
+	model = "Alcatel OneTouch Idol 3 (5.5)";
+	compatible = "alcatel,idol3", "qcom,msm8939", "lk2nd,device";
```
Minecrell, line 12: Mentioned that in the chat already: I think we should match cmdline here like for some other devices since this is just `msm8939-mtp` actually.

vknecht:
Added a match on r63315 panel, although one or more other panel models can be used on this device...

Minecrell:
Applied with small change to commit summary (added `dts:` prefix). Thank you!

PR #17, opened 2020-04-21
Init Samsung SM-J500FN (Rev. 5)

jugendhacker:


Minecrell:
Applied after squashing commits and slight change of commit summary. Thank you!

PR #19, opened 2020-05-11
Add support for new devices, and fix incorrect bootloader name in j5xlte-eur

zqm64:
Device list:
J3109
J3110
J3119
J5008
J500F
J500H
J5108
J510H
J7008
J700P

The bootloader name in j5xlte-eur should be J510F*
And the J510FN should be called j5xnlte

PR #20, opened 2020-05-29
dts: Add Samsung Galaxy J5 2015 (SM-J500H) (Rev. 5)

nergzd723:
Samsung Galaxy J5(2015) device port.

Samsung Galaxy SM-J500H has a different board ID and is not detectable by lk2nd. This adds J500H support to lk2nd.


nergzd723:
Duplicate of #19

PR #23, opened 2020-06-20
Fix builds on GCC 10, add gprimeltecan dts

WTechNinja:
I couldn't get it to build for a while but someone mentioned that a recent update to gcc or something changed the default behaviour from "-fcommon" to "-fnocommon", and I might be mistaken, but I think that's what broke builds on this, because when I added -fcommon my builds are working again.

And the other commit here is adding the gprimeltecan file.
Hopefully it all looks good.

### DIFF
--- a/makefile
+++ b/makefile
@@ -52,7 +52,9 @@ ifeq ($(ENABLE_TRUSTZONE),1)
 endif
 
 INCLUDES := -I$(BUILDDIR) -Iinclude
-CFLAGS := -fcommon -O2 -g -fno-builtin -finline -Wno-multichar -Wno-unused-parameter -Wno-unused-function -include $(CONFIGHEADER)
+CFLAGS := -O2 -g -fno-builtin -finline -Wno-multichar -Wno-unused-parameter -Wno-unused-function -include $(CONFIGHEADER)
+# -fcommon is needed to build this using GCC 10
+CFLAGS += -fcommon
 #CFLAGS += -Werror
 ifeq ($(EMMC_BOOT),1)
   CFLAGS += -D_EMMC_BOOT=1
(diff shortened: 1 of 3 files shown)


WTechNinja:
I have confirmed that the lk2nd for gprimeltecan works well. It boots android with no problems, the only odd thing is that it says the bootloader is locked, but you can still install boot images via fastboot. No problems I've seen so far.

makefile
```diff
@@ -52,7 +52,7 @@ ifeq ($(ENABLE_TRUSTZONE),1)
 endif
 
 INCLUDES := -I$(BUILDDIR) -Iinclude
-CFLAGS := -O2 -g -fno-builtin -finline -Wno-multichar -Wno-unused-parameter -Wno-unused-function -include $(CONFIGHEADER)
+CFLAGS := -fcommon -O2 -g -fno-builtin -finline -Wno-multichar -Wno-unused-parameter -Wno-unused-function -include $(CONFIGHEADER)
```
Minecrell, line 55: `-fcommon` seems to be indeed working with gcc 10, good find! (Note: In general it looks like `-fcommon` isn't really something you want, but this is all ugly legacy code from qcom so it's really the best solution for now...)

Can you put it on an extra line like the `-Werror` below and add a short comment that it's to fix build on gcc 10?

WTechNinja:
it turns out the text editor I used to edit the makefile (Atom) removed some random spaces and tabs in empty fields.... Is that worth fixing or will it not impact anything?

Minecrell:
@WTechNinja For most projects I don't want trailing whitespace at all. But for this project I try to keep minimal diff from the original files so I would prefer if you can revert it. Probably easier to edit it with some editor that doesn't strip trailing whitespace.

WTechNinja:
The whitespace has returned, all should be well.

Minecrell:
Squashed your last two commits and pushed to master. Thank you!

PR #24, opened 2020-06-22
Add support to generate odin-compatible flashable archive

zqm64:
Sorry I can't write the corresponding instructions in README, my english is so bad


zqm64:
Although yes I see there's heimdall in README, But I believe Odin should be always better. And heimdall isn't working for me (I have Galaxy J 2015 Devices).

Minecrell:
> Although yes I see there's heimdall in README, But I believe Odin should be always better. And heimdall isn't working for me (I have Galaxy J 2015 Devices).

I never had any problems with it, and I know of people who have successfully used it on at least J5 2015 and J5 2016. Make sure you use the latest version (Heimdall v1.4.2).

I also have to disagree with "Odin should be always better": Odin is a proprietary tool that is only available in leaked form from Samsung. Heimdall is open-source software, which almost certainly makes it a better choice - as long as it works properly.

In general, I'm not opposed to add a build rule for generating TAR archives that can be used with Odin. (Although I do wonder if it would be easier to just do this manually after building lk2nd.)

If you want the build rule for the TAR archive directly in lk2nd, then please make sure it is [build in a reproducible way](https://en.wikipedia.org/wiki/Reproducible_builds), i.e. the resulting tar archive does not contain user names or file modification dates. See e.g. https://reproducible-builds.org/docs/archives/#full-example (`SOURCE_DATE_EPOCH` isn't set for lk2nd, please just set `--mtime="@0"`). Thanks.

zqm64:
> > Although yes I see there's heimdall in README, But I believe Odin should be always better. And heimdall isn't working for me (I have Galaxy J 2015 Devices).
> 
> I never had any problems with it, and I know of people who have successfully used it on at least J5 2015 and J5 2016. Make sure you use the latest version (Heimdall v1.4.2).
> 
> I also have to disagree with "Odin should be always better": Odin is a proprietary tool that is only available in leaked form from Samsung. Heimdall is open-source software, which almost certainly makes it a better choice - as long as it works properly.
> 
> In general, I'm not opposed to add a build rule for generating TAR archives that can be used with Odin. (Although I do wonder if it would be easier to just do this manually after building lk2nd.)
> 
> If you want the build rule for the TAR archive directly in lk2nd, then please make sure it is [build in a reproducible way](https://en.wikipedia.org/wiki/Reproducible_builds), i.e. the resulting tar archive does not contain user names or file modification dates. See e.g. https://reproducible-builds.org/docs/archives/#full-example (`SOURCE_DATE_EPOCH` isn't set for lk2nd, please just set `--mtime="@0"`). Thanks.

Okay It's done. checksums are now the same across every builds.

Here's my heimdall logs:
(Although I'm using 1.4.1-2+b1 version from official debian repo, But I don't think only just 0.0.1 version will make big differences.)
```
Initialising connection...
Detecting device...
Claiming interface...
Setting up interface...

Initialising protocol...
Protocol initialisation successful.

Beginning session...

Some devices may take up to 2 minutes to respond.
Please be patient!

Session begun.

Downloading device's PIT file...
ERROR: Failed to send request to end PIT file transfer!
ERROR: Failed to download PIT file!
Ending session...
ERROR: Failed to send end session packet!
Releasing device interface...
```

Minecrell:
@zqm64 Actually someone recently reported that 1.4.2 works but 1.4.1 doesn't. There are 3 years between those two versions. I suggest you try it :)

zqm64:
> @zqm64 Actually someone recently reported that 1.4.2 works but 1.4.1 doesn't. There are 3 years between those two versions. I suggest you try it :)

BTW May I ask some questions about mainline linux kernel here?
1. Use which toolchain to build?
2. Does fbcon work for now?

Minecrell:
> BTW May I ask some questions about mainline linux kernel here?

Sure. You can also ask them on the [postmarketOS mainline chat channel](https://wiki.postmarketos.org/wiki/Matrix_and_IRC) if you want (available via IRC and Matrix).

>  1. Use which toolchain to build?

Any recent GCC should be fine. You can also use clang if you want.

> 2. Can fbcon (framebuffer console) work with our current framebuffer?

You can easily set up fbcon on the mainline kernel. Not sure what you mean with "current" framebuffer.

> 3. Use USB Gadget Serial instead of USB Network for debugging?

You should be able to configure any kind of USB gadget (serial, ADB, MTP, ...), although I don't know how USB serial works exactly.

> 4. Will it work on MSM8929 soc without big changes?

MSM8929 [seems to be based on MSM8939 with slightly changed CPU clocks](https://github.com/msm8916-mainline/android_kernel_qcom_msm8916/blob/master/arch/arm/boot/dts/qcom/msm8929.dtsi#L14). While MSM8939 is very similar to MSM8916, it still needs more or less completely separate mainline support. There are some efforts to make MSM8939 work on mainline, quite a few things are working, but as far as I know it's not as good as the mainline support for MSM8916 yet.

You could probably get something working with a few limitations though. I recommend starting with a MSM8916 if you have one.

Minecrell:
I pushed it to master. I would still recommend Heimdall if possible, though. (Also: Feel free to ask more questions about the mainline kernel :))

PR #30, opened 2020-09-17
dts: Add support for Lenovo PHAB Plus

TipzTeam:


dts/rules.mk
```diff
@@ -2,7 +2,7 @@ LOCAL_DIR := $(GET_LOCAL_DIR)
 
 ifeq ($(PROJECT), msm8916-secondary)
 DTBS += \
-	$(LOCAL_DIR)/apq8016-samsung-r02.dtb \
+    $(LOCAL_DIR)/apq8016-samsung-r02.dtb \
```
Minecrell, line 5: Revert

dts/msm8939-qrd.dts
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+/include/ "msm8916.dtsi"
+
+/ {
+	qcom,msm-id = <239 0>;
+	qcom,board-id = <0x1000b 5>;
```
Minecrell, line 9: This looks like the file should be called `msm8939-qrd-skuk.dts`

dts/msm8939-qrd.dts
```diff
@@ -0,0 +1,15 @@
...
+
+	lenovo-pb1770m {
+		model = "Lenovo PHAB Plus";
+		compatible = "lenovo-pb1770m", "lenovo-pb1770n", "qcom,msm8939-qrd-skuk", "qcom,msm8939-qrd", "qcom,msm8939", "qcom,qrd", "lk2nd,device";
```
Minecrell, line 13: You don't need all of them. You need to pick one for your device, e.g. `lenovo,pb1770m` or maybe `lenovo-pb1770` if the two are similar enough. Then you should add `qcom,msm8939` and `lk2nd,device`, e.g.

```suggestion
		compatible = "lenovo,pb1770m", "qcom,msm8939", "qcom,qrd", "lk2nd,device";
```

Also, it seems like this is a shared DTS (in other words, there could be other devices with the same `qcom,board-id`.

This means that you need to add something like `lk2nd,match-cmdline` to make sure only your device is matched. Like here:

https://github.com/msm8916-mainline/lk2nd/blob/799945cb0d2871cf404b4a137637f6ab9d052f49/dts/msm8939-mtp.dts#L12-L16

TipzTeam: There's not a "PB1770", but PB1-770M is just the global variant while PB1-770N is the china only one, and only difference afaik is firmware.

dts/rules.mk
```diff
@@ -2,7 +2,7 @@ LOCAL_DIR := $(GET_LOCAL_DIR)
 
 ifeq ($(PROJECT), msm8916-secondary)
 DTBS += \
-	$(LOCAL_DIR)/apq8016-samsung-r02.dtb \
+    	$(LOCAL_DIR)/apq8016-samsung-r02.dtb \
```
Minecrell, line 5: still weird :)

Also, can you squash your commits? Thanks.

TipzTeam:
@Minecrell @Minecrell @Minecrell @Minecrell @Minecrell @Minecrell @Minecrell @Minecrell @Minecrell @Minecrell @Minecrell @Minecrell @Minecrell @Minecrell Meeeeeerrrrrrrgggggggeeeeee

Minecrell:
I did some minor cosmetic fixes and pushed it. Thanks!

TipzTeam:
@Minecrell you should actually specify the support for PB1-770N, because they are the same device with a different model name and not much people know about this.

Minecrell:
@TipzTeam You chose `pb1770m` as compatible instead of `pb1770`, which made me think there are probably some meaningful  (but maybe minor) differences for the two devices. In that case, I would prefer to not list another device until it was really tested.

TipzTeam:
> @TipzTeam You chose `pb1770m` as compatible instead of `pb1770`, which made me think there are probably some meaningful (but maybe minor) differences for the two devices. In that case, I would prefer to not list another device until it was really tested.

I have both devices and I can confirm it works. Also, why pb1770m was chosen as the unified codename is because the first people working on it called it that, so I will just follow all the first people's trend. But actually any codename works for this, this thing never really have a codename from OEM.

Minecrell:
@TipzTeam OK I fixed it. Thank you.

PR #32, opened 2020-10-01
dts: Add Xiaomi Mi 4i (xiaomi-ferrari)

SUNGOLDSV:


dts/msm8939-mtp.dts
```diff
@@ -14,4 +14,9 @@
 		compatible = "alcatel,idol3", "qcom,msm8939", "lk2nd,device";
 		lk2nd,match-cmdline = "* mdss_mdp.panel=1:dsi:0:qcom,mdss_dsi_r63315_1080p_video:1:none";
 	};
+
+	xiaomi-ferrari {
+		model = "Xiaomi Mi 4i";
+		compatible = "xiaomi,ferrari", "qcom,msm8939", "lk2nd,device";
```
Minecrell, line 20: That's not how it works, you need some `lk2nd,match-cmdline` here. Otherwise it will match for **all** devices.

SUNGOLDSV: Fixed

PR #36, opened 2020-10-11
Add support for Asus Zenfone 2 Laser 720p (Z00L)

antony-jr:
This PR adds support to boot lk2nd in Asus Zenfone 2 Laser 720p (Z00L).

### DIFF
--- a/dts/msm8916-asus-z00l.dts
+++ b/dts/msm8916-asus-z00l.dts
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+/include/ "msm8916.dtsi"
+
+/ {
+        qcom,msm-id =   <206 0>,
+			<248 0>,
+			<249 0>,
+			<250 0>;
+	qcom,board-id = <21 0>;
+	model = "Asus Zenfone Laser 2 ZE550KL";
+	compatible = "asus,z00l", "qcom,msm8916", "lk2nd,device";
+};
--- a/dts/msm8916-asus-z00l.dts
+++ b/dts/msm8916-asus-z00l.dts
@@ -5,16 +5,8 @@
 /include/ "msm8916.dtsi"
 
 / {
-        qcom,msm-id =   <206 0>,
-			<248 0>,
-			<249 0>,
-			<250 0>;
+	qcom,msm-id =   <206 0>;
 	qcom,board-id = <21 0>;
-	compatible = "qcom,msm8916-mtp", "qcom,msm8916";
-	
-	asus-z00l {
-		model = "Asus Zenfone 2 Laser (720p)";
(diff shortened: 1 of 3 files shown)


dts/msm8916-asus-z00l.dts
```diff
@@ -0,0 +1,20 @@
...
+
+/dts-v1/;
+
+/include/ "msm8916.dtsi"
+
+/ {
+        qcom,msm-id =   <206 0>,
+			<248 0>,
+			<249 0>,
+			<250 0>;
```
Minecrell, line 11:
```
qcom,msm-id = <206 0>;
```

should be enough

dts/msm8916-asus-z00l.dts
```diff
@@ -0,0 +1,20 @@
...
+	qcom,board-id = <21 0>;
+	compatible = "qcom,msm8916-mtp", "qcom,msm8916";
+	
+	asus-z00l {
```
Minecrell, line 15: Your original changes in the first commit were fine, this special syntax is only needed when multiple devices use one board-id.

Minecrell (changes requested):

dts/msm8916-asus-z00l.dts
antony-jr: See https://github.com/msm8916-mainline/lk2nd/blob/3f6837d645d36e27a4a0a4298a91e52ce6c0febe/dts/msm8939-asus-z00t.dts#L10 , Asus Z00T is very similar to Z00L but uses a different SOC but the board id seems to be similar.

dts/msm8916-asus-z00l.dts
antony-jr: Alright.

dts/msm8916-asus-z00l.dts
Minecrell: https://github.com/msm8916-mainline/lk2nd/commit/1f5a335422866c8f3e952cfce8ebc8083a85a531

Minecrell:
I mentioned it on Matrix already but FWIW:

Feel free to include gpio-keys and wifi (pronto) already here. I would squash that into the initial commit later anyway. :)

antony-jr:
> Feel free to include gpio-keys and wifi (pronto) already here.

I think you confused with the other PR, forgive me for using similar names for the PR. This is lk2nd.

Minecrell:
> > Feel free to include gpio-keys and wifi (pronto) already here.
> 
> I think you confused with the other PR, forgive me for using similar names for the PR. This is lk2nd.

Yeah I already deleted it here and added it to the other one xD

Minecrell:
Removed some whitespace after `qcom,msm-id`, squashed the commits and pushed it. Thanks!

PR #41, opened 2020-10-19
lk2nd: Introduce mainline panel selection 

Minecrell:
Some devices have multiple possible panels. In that case, we need
to load different panel drivers. On downstream this is implemented
directly in the kernel, but this is difficult to implement cleanly
in mainline.

Instead, it's easier to do this as part of lk2nd. The mainline device
tree only contains a generic compatible (e.g. "motorola,harpia-panel")
with all required properties for it (e.g. GPIOs, regulators, ...).
We assume that the panels are similar enough that they all need exactly
the same properties.

The lk2nd device tree then contains a list of possible panels with
their mainline compatibles (e.g. "motorola,harpia-panel-tianma").
We can identify the panel using the "mdss_mdp.panel" parameter passed
by the primary bootloader.

When mainline is booted, we search for the "generic" compatible,
and replace it with the real one. The kernel will then probe the
correct panel. Yay!

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 - Asus Zenfone 2 Laser (1080p) - Z00T
 - Asus Zenfone Max ZC550KL (2016) - Z010D
 - BQ Aquaris X5 (paella, picmt)
+- Lenovo A6000
 - Lenovo A6010
 - Lenovo PHAB Plus - PB1-770M, PB1-770N
 - LG K10 (m216) - K420 (see notes in `dts/msm8916-lg.dts` for now)
--- a/app/aboot/lk2nd-device.c
+++ b/app/aboot/lk2nd-device.c
@@ -116,14 +116,14 @@ static void parse_boot_args(void)
 			parse_arg(aboot, "carrier=", &lk2nd_dev.carrier);
 			parse_arg(aboot, "radio=", &lk2nd_dev.radio);
 		} else {
-			parse_arg(arg, "mdss_mdp.panel=", &lk2nd_dev.panel);
+			parse_arg(arg, "mdss_mdp.panel=", &lk2nd_dev.panel.name);
 		}
 
 		arg = strtok_r(NULL, " ", &saveptr);
 	}
 	free(args);
 
-	lk2nd_dev.panel = parse_panel(lk2nd_dev.panel);
+	lk2nd_dev.panel.name = parse_panel(lk2nd_dev.panel.name);
 }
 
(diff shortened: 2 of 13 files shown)


vknecht:
This mod works for me on idol3.
Would be great if in each panel subsection, in addition to panel's compatible, one could set status for arbitrary nodes.
For example, with panel r63315 I'd like to set synaptics@20 "disabled" and touchscreen@38 "okay", and the reverse when nt35596 panel is detected.

JonnyMe:
I totally missed this branch, is it still a thing? (by reading from mainline channel it is a bit unclear)

Just for the sake of log I would like to point out that `paella` now has multiple panels problem, an [issue](https://gitlab.com/postmarketOS/pmaports/-/issues/928) was opened on pmaports and it came out it has a different panel in `mdss_mdp.panel`. I'll make a branch for the kernel on my local repo too for testing once I have the confirmation that the generated panel driver for `truly_otm1288a` is working.

Minecrell:
> I totally missed this branch, is it still a thing? (by reading from mainline channel it is a bit unclear)

Yep, I just was not able to find the time to finish it up :/

> Just for the sake of log I would like to point out that paella now has multiple panels problem, an issue was opened on pmaports and it came out it has a different panel in mdss_mdp.panel. I'll make a branch for the kernel on my local repo too for testing once I have the confirmation that the generated panel driver for truly_otm1288a is working.

Actually I made this because someone showed up with the truly_otm1288a panel in October. :) It works fine with a separate panel driver.

JonnyMe:
> Actually I made this because someone showed up with the truly_otm1288a panel in October. :) It works fine with a separate panel driver.

I'll search for that panel driver then if it's already working :-) however I tested this branch and edited current paella panel and dts on a local msm8916/linux branch and it works like a charm!

> Yep, I just was not able to find the time to finish it up :/

I know something about not having lot of time :-)

Minecrell:
> I'll search for that panel driver then if it's already working :-) however I tested this branch and edited current paella panel and dts on a local msm8916/linux branch and it works like a charm!

I think I didn't push it yet, just built a boot image for someone to test. This one I think:

https://matrix.org/_matrix/media/r0/download/matrix.org/vZZYhMWwTVtYWaPvputbjyuf/paella-modem-otm1288a.boot.img

JonnyMe:
Oh great! Thank you!

Minecrell:
@JonnyMe FYI, added the otm1288a panel driver in https://github.com/msm8916-mainline/linux-panel-drivers/commit/aab93d3636861e06d6813ebc6ac377add325f689 + https://github.com/msm8916-mainline/linux/commit/df9d8963939be438fcbd5608c24e718bae3a3142 to force it into video mode. :)

PR #48, opened 2021-01-18
dts: msm8916: wt865x8: use common compatible for panels

jenneron:


Minecrell:
I guess I should have thought of this before making the msm8916-mainline/linux tag. Oh well. (Seems better to change it for a6010 now already so we don't break panel later...)

Minecrell:
But eh I don't want to also make a new lk2nd release too right now. I guess I'll just leave this open until the `wt86518` changes are complete. Thanks!

Minecrell (approved):

PR #49, opened 2021-01-23
dts: msm8939: Add more board-ids for Huawei Honor 5X

wegank:
This commit adds support for all Huawei Honor 5X variants supported by LineageOS, which are

-  KIW-AL{00,10,20}
-  KIW-L{05,21,22,23,24}
-  KIW-CL00
-  KIW-TL00H
-  KIW-UL00


Minecrell:
Usually it doesn't really work that way because then lk2nd has no way to detect the actual board-id (which would be needed when booting the downstream kernel in case the dtbs are different there). Can you post the lk log to check this?

```
fastboot oem lk_log && fastboot get_staged output.txt
```

wegank:
Should be working fine this time. I didn't add all the variants, just the one I'm using.

Minecrell:
Still need the lk_log to tell you :)

```
fastboot oem lk_log && fastboot get_staged output.txt
```

wegank:
[output.txt](https://github.com/msm8916-mainline/lk2nd/files/5861967/output.txt)

Minecrell:
Pushed a couple of changes to simplify this. Can you test this and post the lk_log again?

Also, please check if it's enough to have the `<XXXX 4>` board-ids listed. Usually the last number is just used as a hardware revision so `<XXXX 0>` might not even exist in practice.

wegank:
[output.txt](https://github.com/msm8916-mainline/lk2nd/files/5862110/output.txt)

wegank:
Yes, it's enough to have the `<XXXX 4>` board-ids listed. Tested on downstream and mainline.

PR #53, opened 2021-02-03
scripts/dtbTool: on alpine needs to be libdft.so.1 (added the .1)

craigcomstock:
In order to build on alpine I have to make this change. Not sure if it breaks other system though. :(


minlexx:
dtc-dev provides libfdt.so https://pkgs.alpinelinux.org/contents?branch=edge&name=dtc-dev&arch=x86_64&repo=main

so this is not needed. just install -dev package

PR #55, opened 2021-02-21
Add support for Samsung Galaxy S5

minlexx:
Proofpic [lk2nd-klte](https://user-images.githubusercontent.com/13084776/108623860-552bb480-7452-11eb-809f-711ed4024757.jpg)

Does [lk_log](https://paste.sr.ht/~minlexx/ee247cd5a31d974f06e5bd8c707cf0af823c85af) look fine?


dts/msm8974/msm8974-samsung-klte.dts
```diff
@@ -0,0 +1,20 @@
...
+
+/include/ "skeleton.dtsi"
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+
+	// qcom,board-id = <0xCE08FF01 8>;
+	// ^^ no qcom,board-id in downstream DTB; downstream DTB has only this:
+	qcom,msm-id = <0xC208FF01 14 0x10000>;
+	// model = "SAMSUNG K PROJECT REV14";
```
Minecrell, line 13: I think that's fine, that's the older way to specify those. Can you remove the comments here?

dts/msm8974/msm8974-samsung-klte.dts
```diff
@@ -0,0 +1,20 @@
...
+
+	klte {
+		model = "Samsung Galaxy S5 (SM-G900F)";
+		compatible = "samsung,klte", "qcom,msm8974", "lk2nd,device";
+		lk2nd,match-cmdline = "* samsung.hardware=SM-G900F *";
```
Minecrell, line 18: Based on your log `lk2nd,match-bootloader = "G900F*";` should also work, I think that's a bit nicer for consistency with the msm8916 Samsung devices.

project/msm8974.mk
```diff
@@ -17,6 +17,7 @@ ENABLE_SDHCI_SUPPORT := 1
 ENABLE_USB30_SUPPORT := 1
 
 #DEFINES += WITH_DEBUG_DCC=1
+DEFINES += WITH_DEBUG_LOG_BUF=1
```
Minecrell, line 20: This should be in `msm8974-secondary.mk` (I try to avoid modifying the original make files).

minlexx: B-but.. this is same as in msm8916.mk currently https://github.com/msm8916-mainline/lk2nd/blob/master/project/msm8916.mk#L34

Minecrell: Yeah but for some reason Qualcomm added it there :D

dts/msm8974/msm8974-samsung-klte.dts
minlexx: :ok:

dts/msm8974/msm8974-samsung-klte.dts
minlexx: It works

project/msm8974.mk
minlexx: :ok:

minlexx:
Reworded last commit ro have `project: ` prefix like previous one for the same file

minlexx:
After last commit 
```
	dprintf(INFO, "DEBUG: overwriting entry with (*0xfc4ab000) = 0 (force PSHOLD reset)\n");
	mac = (unsigned char *)entry;
	*(unsigned int *)(mac + 0)  = 0xe3a03a0b;  // mov     r3, #0xb000
	*(unsigned int *)(mac + 4)  = 0xe34f3c4a;  // movt    r3, #0xfc4a // => r3 == 0xfc4ab000
	*(unsigned int *)(mac + 8)  = 0xe3a02000;  // mov     r2, #0
	*(unsigned int *)(mac + 12) = 0xe5832000;  // str     r2, [r3]  // => *r3 = 0
	dprintf(INFO, "DEBUG: this line should not be printed, as board is already rebooting");
```
last I see in UART before reset is 
```
[410] DEBUG: Dumping some kernel info before entering critical section and platform_uninit stuff:
[420] DEBUG: IS_ARM64: false
[420] DEBUG: ===== 1st 128 bytes of kernel entry address:
[430]  000: 0000A0E1 0000A0E1 0000A0E1 0000A0E1
[430]  016: 0000A0E1 0000A0E1 0000A0E1 0000A0E1
[440]  032: 050000EA 18286F01 00000000 F08A7800
[440]  048: 01020304 45454545 D0630000 00900FE1
[440]  064: 6E0D00EB 0170A0E1 0280A0E1 00200FE1
[450]  080: 030012E3 0100001A 1700A0E3 563412EF
[450]  096: 00000FE1 1A0020E2 1F0010E3 1F00C0E3
[460]  112: D30080E3 0400001A 010C80E3 0CE08FE2
[460] DEBUG: overwriting entry with (*0xfc4ab000) = 0 (force PSHOLD reset)
[470] DEBUG: this line should not be printed, as board is already 
```
so that confirms that in theory kernel should be booting..?

minlexx:
This is now ready I think.

Reason for it not working before are pmOS's default msm8974 kernel config that had `CONFIG_ATAGS` and `CONFIG_ARM_APPENDED_DTB`. `CONFIG_ATAGS` is just useless, but `CONFIG_ARM_APPENDED_DTB` is important one.  With `CONFIG_ARM_APPENDED_DTB` disabled I can successfully boot kernel with lk2nd with both "fastboot flash boot" and "fastboot boot". :tada: 

Theory is that `CONFIG_ARM_APPENDED_DTB` forces kernel to use appended DTB blob which was not fixed by bootloader to have correct address and size of memory location in memory node and kernel fails to boot.

Minecrell:
I opened #60 to implement proper support for `qcom,msm-id` so that downstream kernels for klte could also be booted through lk2nd in the future.

Thanks for all the time you spent to make this work properly, @minlexx!

Rongronggg9:
Hi. Thanks for all your efforts!

Galaxy S5 (Qualcomm) has multiple variants and mine is [`kltechnduo` (G9006W - that's mine, G9008W)](https://wiki.lineageos.org/devices/kltechnduo/).
Even for `klte`, there are still [a lot of sub-variants (G900F/M/R4/R7/T/V/W8)](https://wiki.lineageos.org/devices/klte/).

Unfortunately, after flashing `lk2nd`, it can't boot at all and make my device immediately fall back to the Odin mode.
However, the kernel for `klte` from pmOS works (if I directly flash it to `BOOT`).

Currently, I do not have a UART cable so that's all I can confirm.

Would changing this line make it works?
https://github.com/msm8916-mainline/lk2nd/blob/eda99658e1c9cae78a01d6a7eca5d994462232b6/dts/msm8974/msm8974-samsung-klte.dts#L14

minlexx:
probably, what's in your stock cmdline?

Rongronggg9:
> probably, what's in your stock cmdline?

Now I've located the problem, it is the revision id.
The max revision id of `kltechn` (`kltechnduo` is mostly the same) is `10` and that should be the retail revision. ([ref](https://github.com/LineageOS/android_kernel_samsung_msm8974/tree/lineage-18.1/arch/arm/boot/dts/msm8974pro))



From [specdevice](https://specdevice.com/showspec.php?id=c712-4732-ffff-ffffe5b57a23) we can verify that.

```
Processor : ARMv7 Processor rev 1 (v7l)
processor : 0
BogoMIPS : 38.00
processor : 1
BogoMIPS : 38.00
processor : 2
BogoMIPS : 38.00
processor : 3
BogoMIPS : 38.00
Features : swp half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32
CPU implementer : 0x51
CPU architecture: 7
CPU variant : 0x2
CPU part : 0x06f
CPU revision : 1
Hardware : Qualcomm MSM8974PRO-AC
Revision : 000a
Serial : 00004a77000096fb
```

So changing the revision id to `10` works like a charm!
```diff
--- a/dts/msm8974/msm8974-samsung-klte.dts
+++ b/dts/msm8974/msm8974-samsung-klte.dts
@@ -6,7 +6,7 @@

 / {
        // This is used by the bootloader to find the correct DTB
-       qcom,msm-id = <0xC208FF01 14 0x10000>;
+       qcom,msm-id = <0xC208FF01 10 0x10000>;

        klte {
                model = "Samsung Galaxy S5 (SM-G900F)";
```



Later I may open a PR for `kltechn` (compatible with `kltechnduo`, of course).

PR #58, opened 2021-02-27
scripts: fix shebang

qbladea:
Due to the strange shebang in the script, the patche-shebang failed
when it was built on the Guix/Nix package manager. I fixed it.

### DIFF
--- a/scripts/dtbTool
+++ b/scripts/dtbTool
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#!/usr/bin/env python
 #
 #  Copyright (c) 2015-2016, The Linux Foundation. All rights reserved.
 #
--- a/scripts/mkbootimg
+++ b/scripts/mkbootimg
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#!/usr/bin/env python
 #
 #  Copyright (c) 2013, The Linux Foundation. All rights reserved.
 #


Minecrell:
Hi @qbladea. I'm a bit confused about this PR. What is `patche-shebang` and what exactly are you trying to do here?

qbladea:
Since Guix/Nix does not use paths such as /usr/bin/python /bin/sh /usr/bin/awk to place the interpreter, patch-shebang is change it to

``` 
/xxx/store/...../bin/python
/xxx/store/...../bin/sh
/xxx/store/...../bin/awk
```

But this project uses a strange way of writing, which makes it unable to match.

Minecrell:
@qbladea But what exactly are you trying to do here? Package lk2nd as Guix/Nix/... package? How is that useful?

qbladea:
I am trying to use Guix to generate the boot.img and system.img of Redmi 2.

I did not do Nix, but the packaging work after patch-shebang repair will be very simple. Those who are interested can use it in NixOS-Mobile.

Minecrell:
I'm still a bit confused why this is necessary. Perhaps `patch-shebang` should be fixed to parse these shebangs with space as well since I don't think they are wrong per-se. Anyway, it does look a bit ugly so I will merge your change for `mkbootimg` and `dtbTool`. `test-aboot.py` is unused as far as I can see so I would rather leave it alone.

PR #59, opened 2021-03-03
lk2nd: Add smb1360 battery detection for l8910

JonnyMe:


lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -15,21 +22,105 @@ static const struct smb1360_battery batteryB = {
...
+	profile = smb1360_check_batt_profile(fdt, offset, connected_rid);
 
-	return &batteryA; // TODO
+	switch (profile) {
+	case SMB1360_BATTERY_PROFILE_A:
+		return &batteryA;
+	case SMB1360_BATTERY_PROFILE_B:
+		return &batteryB;
+	case SMB1360_BATTERY_PROFILE_KEEP:
+		return &batteryK;
```
Minecrell, line 122: IMO we should return `NULL` ("battery detection failed") in this case. This will mean that smb1360 will stay disabled since we don't know what to do in that case.

lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -15,21 +22,105 @@ static const struct smb1360_battery batteryB = {
...
+	status_t ret;
+
+	int i, len, timeout, loaded_profile, new_profile;
+	uint8_t val = 0;
+	uint32_t profile_rid[BATTERY_PROFILE_MAX], rid10;
+
+	prop = fdt_getprop(fdt, offset, "qcom,profile-a-rid-kohm", &len);
+	if (len < 0)
+		return -EINVAL;
+	profile_rid[0] = fdt32_to_cpu(*prop);
```
Minecrell, line 46: maybe make this normal variables now and call it

```suggestion
	profile_a_rid = fdt32_to_cpu(*prop);
```

lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -15,21 +22,105 @@ static const struct smb1360_battery batteryB = {
...
+
+	volt = pm8x41_adc_channel_read(VADC_BAT_CHAN_ID);
+	batt_id_uv = volt * 1000;
+
+	if (batt_id_uv == 0)
+		return -EINVAL;
+
+	denom_dividend = vref * 1000000LL;
+	denom_quotient = batt_id_uv;
+	denom = (abs64(denom_dividend) / abs64(denom_quotient)) - 1000000LL;
```
Minecrell, line 92: Hmm, downstream uses `denom = div64_s64(vref * 1000000LL, batt_id_uv) - 1000000LL;`, so a 64-bit signed division. Where does the abs64 come from?

JonnyMe: I was looking into `div64_s64` to see how it performed the division and saw it there, https://elixir.bootlin.com/linux/v4.9/source/lib/div64.c#L163. By checking the results, downstream returns a value for `connected_rid` of 100. Without `abs` I get -99, with `abs` I get 99. I guess the difference of 1 in result is given by some round. What do you think?

Minecrell: If I'm reading this correctly `div64_s64` does an unsigned 64-bit division with abs first and then in the `t = (dividend ^ divisor) >> 63;` line computes the sign separately and adds it back. So effectively the result should be the one without abs(). It seems like something else is odd here...

lk2nd/smb1360/smb1360-qcom.c
JonnyMe: By looking at downstream, in case the loop in `smb1360_check_batt_profile` reaches its maximum without having found a match it returns 0, meaning it leaves things unchanged. If we return `NULL` here it won't enable the driver at all. If you prefer so for safety reasons I'll change it to `NULL`. Let me know :-)

lk2nd/smb1360/smb1360-qcom.c
JonnyMe: Sure, I'll change it now

lk2nd/smb1360/smb1360-qcom.c
JonnyMe: Ohh...I'll rework on it then, thank you

lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -15,21 +22,105 @@ static const struct smb1360_battery batteryB = {
...
+static uint32_t smb1360_parse_batt_id(const void *fdt, int offset)
+{
+	const uint32_t *prop;
+	int64_t denom, batt_id_uv;
+	int64_t denom_dividend, denom_quotient, connected_rid_dividend, connected_rid_quotient;
+	int32_t len, volt, rpull = 0, vref = 0;
+	uint32_t connected_rid;
+
+	prop = fdt_getprop(fdt, offset, "qcom,batt-id-vref-uv", &len);
+	if (len == sizeof(vref))
```
Minecrell, line 75:
```suggestion
	if (len != sizeof(vref))
```

lk2nd/smb1360/smb1360-qcom.c
Minecrell: Yep actually I do prefer that I think.

lk2nd/smb1360/smb1360-qcom.c
JonnyMe: I saw that re-reading it, thank you :-)

lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -2,9 +2,15 @@
 
 #include <debug.h>
 #include <libfdt.h>
+#include <pm8x41.h>
 #include <pm8x41_adc.h>
+#include <pm8x41_hw.h>
+#include <platform/timer.h>
```
Minecrell, line 8: all of these are unused I suspect

lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -15,21 +21,97 @@ static const struct smb1360_battery batteryB = {
...
+static uint32_t smb1360_parse_batt_id(const void *fdt, int offset)
+{
+	const uint32_t *prop;
+	int64_t denom, batt_id_uv;
+	int32_t len, rpull = 0, vref = 0;
+	uint32_t connected_rid;
+
+	prop = fdt_getprop(fdt, offset, "qcom,batt-id-vref-uv", &len);
+	if (len != sizeof(vref))
+		return -EINVAL;
```
Minecrell, line 70: I think we can just take `return 0` as invalid value here and remove the define.

lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -15,21 +21,97 @@ static const struct smb1360_battery batteryB = {
...
+
+	if ((profile_rid_a - connected_rid) < rid10) {
+		new_profile = SMB1360_BATTERY_PROFILE_A;
+	} else if ((profile_rid_b - connected_rid) < rid10) {
+		new_profile = SMB1360_BATTERY_PROFILE_B;
+	} else {
+		return -EINVAL;
+	}
+
+	return new_profile;
```
Minecrell, line 58: and here we could do

```c
	if ((profile_rid_a - connected_rid) < rid10) {
		return &profileA;
	} else if ((profile_rid_b - connected_rid) < rid10) {
		return &profileB;
	} else {
		return NULL;
	}
```

to avoid the `-EINVAL` and the switch-case in smb1360_qcom_detect_battery

lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -15,21 +21,97 @@ static const struct smb1360_battery batteryB = {
...
+
+	if (batt_id_uv == 0)
+		return -EINVAL;
+
+	denom = (vref * 1000000LL / batt_id_uv) - 1000000LL;
+	if (denom == 0)
+		return -EINVAL;
+
+	connected_rid = (rpull * 1000000LL + (denom / 2)) / denom;
+
```
Minecrell, line 88: would be nice to log both batt-id-uv and connected_rid here using something like

```c
dprintf(SPEW, "batt-id voltage: %lld, connected_rid: %d\n", batt_id_uv, connected_rid);

File: lk2nd/smb1360/smb1360-qcom.c [hunk 9]
```diff
@@ -15,21 +16,79 @@ static const struct smb1360_battery batteryB = {
...
 
+static const struct smb1360_battery *smb1360_check_batt_profile(const void *fdt, int offset,
+								uint32_t connected_rid)
+{
+	const uint32_t *prop;
+	int len, loaded_profile;
+	uint32_t profile_rid_a, profile_rid_b, rid10;
+
+	prop = fdt_getprop(fdt, offset, "qcom,profile-a-rid-kohm", &len);
+	if (len < 0)
```
Minecrell, line 27:
```suggestion
	if (len != sizeof(profile_rid_a))
```

lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -15,21 +16,79 @@ static const struct smb1360_battery batteryB = {
...
 
-	// TODO: Read qcom,batt-id-vref-uv etc and do stuff
-
-	volt = pm8x41_adc_channel_read(VADC_BAT_CHAN_ID);
-	dprintf(INFO, "batt_id_v = %u\n", volt);
-	// Note: batt_id_v seems close but also slightly different from mainline/downstream
-	// I wonder where all the different values come from
+	connected_rid = smb1360_parse_batt_id(fdt, offset);
+	if (connected_rid == 0)
+		return NULL;
```
Minecrell, line 91: maybe `dprintf(CRITICAL, "...");` something here that detecting connected_rid failed

lk2nd/smb1360/smb1360-qcom.c
```diff
@@ -15,21 +16,79 @@ static const struct smb1360_battery batteryB = {
...
+	profile_rid_b = fdt32_to_cpu(*prop);
+
+	rid10 = connected_rid / 10;
+
+	if ((profile_rid_a - connected_rid) < rid10) {
+		return &batteryA;
+	} else if ((profile_rid_b - connected_rid) < rid10) {
+		return &batteryB;
+	} else {
+		return NULL;
```
Minecrell, line 43: and here something like

```c
dprintf(CRITICAL, "connected_rid: %d does not match profile A (%d) or profile B (%d)\n",
connected_rid, profile_rid_a, profile_rid_b);
```

Issue #37, opened 2020-10-11
Add option to set GPIO for volume down button


Minecrell:
z00l/a51f/m216 all use a GPIO for volume down, not the resin node. This needs to be made configurable somehow.

antony-jr:
For solution we can always erase the boot partition with stock fastboot and then flash lk2nd again which worked pretty good up until now for me. But you can always just boot the kernel directly instead of flashing. If you are going to use it as daily driver then flash it.

So if anyone looking for solution, do the following,
```
 $ # Get your phone into stock fastboot
 $ fastboot erase boot # Erase boot partition
 $ fastboot flash boot lk2nd.img # flash lk2nd again without the kernel
 $ fastboot reboot
```

PR #63, opened 2021-03-29
Add support for LG Leon LTE (c50)

jenneron:
Device does not have fastboot and ability to boot into recovery with keys combination itself.
It uses both volume buttons as gpios, volume down is 107 and volume up is 108. Usually 107 is volume up, so lk2nd reacts on volume down key as volume up and doesn't react on volume up key at all. I can boot into recovery via lk2nd menu and use fastboot boot, but after flashing boot.img there will no ability to enter lk2nd fastboot.
For now i hardcoded gpios in target/msm8916/init.c, so PR is draft.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 - Lenovo A6010
 - Lenovo PHAB Plus - PB1-770M, PB1-770N
 - LG K10 (m216) - K420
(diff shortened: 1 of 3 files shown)


Minecrell:
I merged the first commit because I think it's fine. The problem with the buttons is essentially reported already as https://github.com/msm8916-mainline/lk2nd/issues/37. Would be nice to see a proper solution for it (GPIO configurable via device tree I guess). That's better for a separate PR though.

PR #71, opened 2021-05-02
Handle custom keys in lk2nd

### LINKED ISSUE #37: Add option to set GPIO for volume down button

TravMurav:
Add `lk2nd,keys` property that adds/overrides default keys used by lk2nd. Fixes #37 

The new property is an array of 4 value records:
```
#include <lk2nd.h>
// < KEY_CODE GPIO PULL ACTIVE >
lk2nd,keys = 
        <KEY_VOLDOWN LK2ND_RESIN>;
        <KEY_VOLUP 107 GPIO_PULL_UP GPIO_ACTIVE_LOW>;
        <KEY_HOME 109 GPIO_PULL_UP GPIO_ACTIVE_LOW>;
```
(`LK2ND_RESIN` is a macro for 3 magic values (0 0 0) to use resin)
Keycodes can be duplicated.
KEY_HOME keycode is used as an extra keycode that leads into fastboot mode.
If there is no record for some keycode, the default handler is used.

Also:
 - Show version of lk2nd on the screen (to make remote troubleshooting easier and have less "is this lk2nd? questions")
 - Add sm-t555


target/msm8916/init.c
```diff
@@ -196,6 +254,10 @@ void target_init(void)
 
 	spmi_init(PMIC_ARB_CHANNEL_NUM, PMIC_ARB_OWNER_ID);
 
+#if WITH_LK2ND
+	lk2nd_init();
+#endif
```
Minecrell, line 259: I wonder if we can put this in some LK core file instead to avoid patching every SoC? Maybe directly before `target_init()`.

lk2nd/lk2nd-device.c
```diff
@@ -275,6 +275,35 @@ static void lk2nd_parse_device_node(const void *fdt)
...
+	val = fdt_getprop(fdt, offset, "lk2nd,keys", &len);
+	if (len > 0 && len % 16 == 0) {
+		len /= 16;
+		/* sentinel: key=0 */
+		map = calloc(len + 1, sizeof(struct lk2nd_keymap));
+		for (int i = 0; i < len; i++) {
+			map[i].key = fdt32_to_cpu(val[i*4]);
+			map[i].gpio = fdt32_to_cpu(val[i*4 + 1]);
+			map[i].pull = fdt32_to_cpu(val[i*4 + 2]);
+			map[i].active = fdt32_to_cpu(val[i*4 + 3]);
```
Minecrell, line 302: Can we put pull + active into one "flags" value maybe? Just to make it a bit more like Linux. Also, the uint32 is probably more than large enough

lk2nd/lk2nd-device.c
```diff
@@ -275,6 +275,35 @@ static void lk2nd_parse_device_node(const void *fdt)
...
+		return NULL; // Will be reported later again. Hopefully.
+
+	offset = lk2nd_find_device_offset(fdt);
+	if (offset < 0)
+		return NULL;
+
+	/* FIXME: Assuming that all keys are pulled UP */
+	val = fdt_getprop(fdt, offset, "lk2nd,keys", &len);
+	if (len > 0 && len % 16 == 0) {
+		len /= 16;
```
Minecrell, line 295: 4 * sizeof(uint32_t)

lk2nd/lk2nd-device.c
TravMurav: I was trying to avoid moving bits back and forth here and reuse the defines from gpio.h that lk uses. Can probably shift active into the second byte and keep pull as is I guess

make/build.mk
```diff
@@ -50,7 +50,8 @@ $(OUTELF_STRIP): $(OUTELF)
 $(BUILDDIR)/%.dtb: %.dts
 	@$(MKDIR)
 	@echo compiling $<
-	$(NOECHO)dtc -O dtb -i dts -o $@ $<
+	$(NOECHO)cpp -nostdinc -I dts -I arch -undef -x assembler-with-cpp $< $@.prep
```
Minecrell, line 53: Is an include from arch actually used anywhere?

lk2nd/lk2nd-device.c
```diff
@@ -275,6 +275,34 @@ static void lk2nd_parse_device_node(const void *fdt)
 #endif
 }
 
+struct lk2nd_keymap* lk2nd_fdt_parse_early_keys(void)
```
Minecrell, line 278: Can we make this not early(TM) somehow? Would prefer not calling `lk2nd_find_device_offset` more often than necessary. I thought that's why you moved `lk2nd_init()`. You should parse the `struct lk2nd_keymap` into `struct lk2nd_device` and then use it from there.

target/msm8974/init.c
```diff
@@ -149,15 +150,85 @@ uint32_t target_volume_down()
...
+	return 0;
+}
+
+int target_volume_up() {
+    return target_key_pressed(KEY_VOLUMEUP);
+}
+
+int target_volume_down() {
+    return target_key_pressed(KEY_VOLUMEDOWN);
+}
```
Minecrell, line 214: I wonder if we can (and if it's worth) de-duplicating this. It seems fairly similar for both 8916+8974.

make/build.mk
TravMurav: Oh I guess I missed it when I copied the line xD

target/msm8974/init.c
Minecrell: Something like `target_display_cont_splash_mdp5.c` maybe. Make a `target_keys_lk2nd.c`, define `target_volume_up/down` there and then guard the original definitions with `#if !WITH_LK2ND`. Would this work?

lk2nd/lk2nd-device.c
```diff
@@ -275,6 +275,34 @@ static void lk2nd_parse_device_node(const void *fdt)
...
+		return NULL; // Will be reported later again. Hopefully.
+
+	offset = lk2nd_find_device_offset(fdt);
+	if (offset < 0)
+		return NULL;
+
+	val = fdt_getprop(fdt, offset, "lk2nd,keys", &len);
+	if (len > 0 && len % (3 * sizeof(uint32_t)) == 0) {
+		len /= (3 * sizeof(uint32_t));
+		/* sentinel: key=0 */
```
Minecrell, line 295: What is this supposed to tell me?

lk2nd/lk2nd-device.c
```diff
@@ -275,6 +275,34 @@ static void lk2nd_parse_device_node(const void *fdt)
...
+	if (!fdt || dev_tree_check_header(fdt))
+		return NULL; // Will be reported later again. Hopefully.
+
+	offset = lk2nd_find_device_offset(fdt);
+	if (offset < 0)
+		return NULL;
+
+	val = fdt_getprop(fdt, offset, "lk2nd,keys", &len);
+	if (len > 0 && len % (3 * sizeof(uint32_t)) == 0) {
+		len /= (3 * sizeof(uint32_t));
```
Minecrell, line 294: Maybe `#define KEY_SIZE (3 * sizeof(uint32_t))`

target/msm8974/init.c
TravMurav: Then how would I reuse those methods for, e.g. fallback?
Like this?
```
#if LK2ND
int foo_old()
#else
int foo()
#endif
{
...
```

TravMurav: And then IMO 8916/8974 are pretty different: 8916 is gpio + resin and 8974 is pmic_gpio and (resin/resin_v2)

lk2nd/lk2nd-device.c
TravMurav: Probably a very bad attempt to say
` /* last element indicates end of the array with key=0 */ `

target/msm8974/init.c
Minecrell: Are the buttons even necessarily on PMIC on 8974? Sounds kind of like the typical unfortunate lk2nd assumption that makes one wish we had invested time in something more flexible than lk2nd (with proper device trees) :D

TravMurav: I think fp2 needs pmic button looking at the mainline dt, and looking at a couple more + the fact that the target uses that makes me assume that *most* of 8974 use pmic gpio for the buttons (e.g.  klte uses 3 for home...)

Minecrell: Maybe we should make this configurable though while we're at it? The PMIC GPIOs could also be used on 8916 theoretically and the other way around for 8974. The way to access SoC/PMIC GPIOs is pretty much the same for 8916/8974 I think?

TravMurav: Hm, do you suggest something like 
```
#define PM_GPIO(x) (x | (1 << 16))
```
and check for the flag in the target handler?

Minecrell: Well I don't know, perhaps we might as well define a dummy

```
/ {
    msmgpio: gpio {
        phandle = <0xMAGIC1>;
    };
    pmic_gpio: pmic-gpio {
        phandle = <0xMAGIC2>;
    };
}
```
and then properly do

```
something-gpios = <&msmgpio 42 GPIO_ACTIVE_HIGH>;
```

where the first uint32_t will end up as either 0xMAGIC1 or 0xMAGIC2.

Or we abandon lk2nd and do this properly with U-Boot.

minlexx: > Or we abandon lk2nd and do this properly with U-Boot.

You're going to start a revolution with statements like this :fire:

make/build.mk
```diff
@@ -50,7 +50,8 @@ $(OUTELF_STRIP): $(OUTELF)
 $(BUILDDIR)/%.dtb: %.dts
 	@$(MKDIR)
 	@echo compiling $<
-	$(NOECHO)dtc -O dtb -i dts -o $@ $<
+	$(NOECHO)cpp -nostdinc -I dts -undef -x assembler-with-cpp $< $@.prep
```
Minecrell, line 53:
```suggestion
	$(NOECHO)$(TOOLCHAIN_PREFIX)cpp -nostdinc -I dts -undef -x assembler-with-cpp $< $@.prep
```
?

target/target_keys_lk2nd.c
```diff
@@ -0,0 +1,109 @@
...
+	keys_init();
+	target_init_keys();
+
+	if(target_volume_down())
+		keys_post_event(KEY_VOLUMEDOWN, 1);
+
+	if(target_volume_up())
+		keys_post_event(KEY_VOLUMEUP, 1);
+
+	if(target_key_pressed(KEY_HOME))
```
Minecrell, line 105:
```suggestion
	if (target_key_pressed(KEY_HOME))
```

PR #69, opened 2021-04-28
msm8974: Add support for Fairphone 2

z3ntu:
~~Volume down button doesn't work - #37~~

`fastboot boot` lk2nd from lk2nd doesn't work (booting TWRP works fine)

booting mainline also seems to have some problems iirc?

The rest works fine :)

PR #73, opened 2021-05-14
Add SONY Rhine and pm8941 SPMI regulators

konradybcio:


konradybcio:
Sadly no display, nobody cared enough to enable display in the original BL, but hey, at least the eMMC isn't suicidal..

dts/msm8974/msm8974-sony-xperia-rhine.dts
```diff
@@ -0,0 +1,16 @@
...
+/include/ "skeleton.dtsi"
+
+/ {
+	model = "SONY Xperia Z1 / Z1 Compact / Z Ultra";
+	compatible = "somc,rhine", "qcom,msm8974", "lk2nd,device";
+	// thanks for giving us device-unique IDs, oh wait..
+	lk2nd,match-bootloader = "*Rhine*";
+
+	// You are ID 8, you are ID 8, EVERYBODY IS ID 8!
+	qcom,msm-id = <0 0>; /* Rhine bootloader doesn't care but dtbTool apparently does */
```
Minecrell, line 14: Would prefer if you add some useful msm-id here if bootloader doesn't care.

dts/msm8974/msm8974-sony-xperia-rhine.dts
```diff
@@ -0,0 +1,16 @@
...
+	qcom,board-id = <8 0>;
```
Minecrell, line 15: I guess you will need to coordinate with @z3ntu (#69) and add a generic dts with subnodes for msm8974-mtp I guess.

z3ntu:
This was my solution which also covers pm8841 and pma8084 https://github.com/z3ntu/lk2nd/commit/5e147f7056ee0b080a5e224c68de6927a212288a

minlexx:
WHy not have separate structs for each?

z3ntu:
If you mean pma8084 vs pm8941, you'd need to query the pmic id from smem and choose the correct struct based on that - which is more complicated.

minlexx:
lk already does this (parse smem etc)

TravMurav:
Sorry, this PR got a bit stale...
Lately we were trying to refactor lk2nd so I'm not sure if this change is still relevant thus I'm closing it. If you want to respin it, please feel free to reopen or re-target to the `experimental-tmp2` branch.

PR #76, opened 2021-06-03
dts: Add support for LYF Wind 1 (lyf-ls5010)

SUNGOLDSV:


dts/msm8916/msm8916-qrd5-v1.dts
```diff
@@ -0,0 +1,15 @@
...
+
+#include "skeleton.dtsi"
+
+/ {
+	compatible = "qcom,msm8916-v1-qrd/5-v1", "qcom,msm8916";
+
+	lyf-ls5010 {
+		model = "LYF Wind 1 (lyf-ls5010)";
+		compatible = "lyf,ls5010", "qcom,msm8916", "lk2nd,device";
+		lk2nd,match-cmdline = "* mdss_mdp.panel=1:dsi:0:qcom,mdss_dsi_hx8394_d_hd_video_truly:1:none";
```
Minecrell, line 13: I think it's better to prefer for adding panel support here and add a list of all possible panels, and then use `lk2nd,match-panel;`. See e.g. `msm8916-qrd7+12-v1.dts` for an example.

The available panels are those you can find in the decompiled dts below `qcom,mdss_mdp@1a00000` (excluding the simulator ones), i.e. `qcom,mdss_dsi_hx8394_d_hd_video_truly`, `qcom,mdss_dsi_hx8394_d_hd_video_holitech`, `qcom,mdss_dsi_hx8394_f_hd_video_bitland`, `qcom,mdss_dsi_kd050m7_hd_video_kd`. And you would translate them to a compatible like

`lyf,truly-hx8394`, or `lyf,kd-kd050m7`, idk, something like that

Minecrell (changes requested):

Minecrell (approved):

PR #78, opened 2021-06-10
dts: htc-m8qlul: Add support for HTC One M8s

Newbytee:
This adds support for htc-m8qlul (or htc-m8ql_ul), also known as HTC One M8s.

This doesn't actually work! But it does. You need to disable all other dts and then it works. If you don't, the device will go to its own fastboot instead of lk2nd after you boot. Minecrell suggested we merge it like this and do something more proper when/if I get mainline Linux booting on this thing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 - Asus Zenfone Max ZC550KL (2016) - Z010D
 - BQ Aquaris X5 - paella, picmt
 - DragonBoard 410c - apq8016-sbc
+- HTC One M8s - m8qlul (quirky - see comment in `dts/msm8916/msm8939-htc-m8qlul.dts`)
 - Huawei Ascend G7 - G7-L01
 - Huawei Honor 5X - kiwi
 - Lenovo A6000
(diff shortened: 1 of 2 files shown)


dts/msm8916/rules.mk
```diff
@@ -1,5 +1,9 @@
 LOCAL_DIR := $(GET_LOCAL_DIR)
 
+# To build for htc-m8qlul, comment out all dtbs and add
+# $(LOCAL_DIR)/msm8939-htc-m8qlul.dtb
+# m8qlul does not work with all dtbs enabled.
```
Minecrell, line 5: Can you put the comment in the dts file (and adjust reference in README). Seems a bit better there.

README.md
```diff
@@ -28,6 +28,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 - Asus Zenfone Max ZC550KL (2016) - Z010D
 - BQ Aquaris X5 - paella, picmt
 - DragonBoard 410c - apq8016-sbc
+- HTC One M8s - m8qlul (quirky - see comment in makefile)
```
Minecrell, line 31:
```suggestion
- HTC One M8s - m8qlul (quirky - see comment in `dts/msm8916/msm8939-htc-m8qlul.dts`)
```

Minecrell (approved):

PR #81, opened 2021-06-27
Add low effort LG D855 dts

swiftgeek:
What works:
* Volume buttons as up/down
What does not work:
* Recovery mode

I didn't bother to add more variants either, just one I have in hand.


Minecrell:
Pushed with some minor modifications, thanks!

PR #83, opened 2021-07-06
dts: sm-g5108q: Add support for the Samsung Galaxy Core Max

mxpro1996:
Work well on this msm8916 device.
IMG_20210706_090833


Minecrell:
Moved the comment to above the `g5108q {` block and pushed it, thanks!

PR #91, opened 2021-09-06
dts: alcatel-idol3: Use panel selection

vknecht:


TravMurav (approved):

dts/msm8916/msm8939-mtp.dts
```diff
@@ -12,7 +12,24 @@
...
+		lk2nd,match-panel;
+
+		panel {
+			compatible = "alcatel,idol3-panel";
+
+			qcom,mdss_dsi_jdi_1080p_video {
+				compatible = "alcatel,idol3-panel-jdi";
+			};
+			qcom,mdss_dsi_nt35596_1080p_video {
+				compatible = "alcatel,idol3-panel-nt35596";
```
Minecrell, line 24: Do you have this panel? Not sure if we should support it, perhaps this is just some early prototype?

vknecht: My devices have nt35596 (not v03) and r63315, and saw no evidence of others being used.
Only those two have generated drivers... Should the others be removed ?

Minecrell: The jdi one looks like an exact copy from the qcom tree, it probably doesn't actually exist. I'd say we can remove that one. Not sure about the v3, you can keep it or remove it. Toss a coin maybe :D

vknecht: Ok I removed the jdi one

Minecrell (approved):

PR #93, opened 2021-09-13
Cherry-pick « Add "fastboot oem pstore" command » by Junak

vknecht:
Requires addition in device' dts eg.
lk2nd,pstore = <0x9ff00000 0x00100000>;

Couldn't check it really works yet, so it's a draft for now...

### DIFF
--- a/app/aboot/aboot.c
+++ b/app/aboot/aboot.c
@@ -791,7 +791,7 @@ unsigned char* generate_mac_address()
 
 	return mac;
 }
-
+extern void lk2nd_clear_pstore();
 typedef void entry_func_ptr(unsigned, unsigned, unsigned*);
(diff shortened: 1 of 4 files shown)


TravMurav:
I *really* don't like how this patch looks: the pstore stuff is just smeared across lk2nd code, that can be done much better IMO 

Is there any data format spec for ramoops/pstore? Dumping the data is easy but how easy it is to decode it nicely? (tbh implementing this was kinda on my todo list for low priority things in lk2nd)

TravMurav:
This was left as a draft for quite a while now and I think while the idea is good, it needs more thought to be implemented properly. I have left some notes for self to look into it for the lk2nd refactor branch.

PR #94, opened 2021-09-15
dts: huawei-kiwi: add modem to config

### LINKED ISSUE #75: kiwi doesn't connect to mobile network
When booting through lk2nd, my huawei-kiwi (Honor 5X, KIW-L21) doesn't connect to the mobile network anymore in my latest LineageOS system ("No service"). Everything is working whenever I boot the device from the stock bootloader.

To illustrate this further, here's my sequence of commands (but of course it is the same result when flashing lk2nd to boot instead of booting it directly):

```
$ adb reboot bootloader
$ fastboot boot build-msm8916-secondary/lk2nd.img # or an image from the releases download
Downloading 'boot.img'                             OKAY [  0.012s]
booting                                            OKAY [  0.002s]
Finished. Total time: 0.015s
$ fastboot boot ~/boot.img # original LineageOS boot image
Downloading 'boot.img'                             OKAY [  0.969s]
booting                                            OKAY [  0.102s]
Finished. Total time: 1.092s

# phone doesn't connect to mobile network now

$ adb reboot bootloader
$ fastboot boot ~/boot.img
Downloading 'boot.img'                             OKAY [  0.958s]
booting                                            OKAY [  0.017s]
Finished. Total time: 0.986s

# phone now connects to mobile network again
```

Hoping that it helps, here is an [lk_log.txt](https://github.com/msm8916-mainline/lk2nd/files/6478980/lk_log.txt).

fynngodau:
This fixes #75 like @Newbytee suggested in https://github.com/msm8916-mainline/lk2nd/issues/75#issuecomment-920028380.

I don't know how this works but it resolves the issue where the device didn't connect to mobile network after booting :upside_down_face: :tada:

### DIFF
--- a/dts/msm8916/msm8939-huawei-kiwi.dts
+++ b/dts/msm8916/msm8939-huawei-kiwi.dts
@@ -17,5 +17,6 @@
 	model = "Huawei Honor 5X";
 	compatible = "huawei,kiwi", "qcom,msm8939", "lk2nd,device";
 
+	// fix modem not connecting to cellular network
 	huawei,modem_id = <0xFFB40105 0x0>;
 };
--- a/dts/msm8916/msm8939-huawei-kiwi.dts
+++ b/dts/msm8916/msm8939-huawei-kiwi.dts
@@ -16,4 +16,6 @@
 
 	model = "Huawei Honor 5X";
 	compatible = "huawei,kiwi", "qcom,msm8939", "lk2nd,device";
+
+	huawei,modem_id = <0xFFB40105 0x0>;
 };


Newbytee:
Maybe add a comment explaining why it's there in case someone uses this as reference for something else in the future?

fynngodau:
Sure!

dts/msm8916/msm8939-huawei-kiwi.dts
```diff
@@ -16,4 +16,7 @@
 
 	model = "Huawei Honor 5X";
 	compatible = "huawei,kiwi", "qcom,msm8939", "lk2nd,device";
+
+	// fix modem not connecting to cellular network
```
Newbytee, line 20: I'd probably word it like this, but I'm not saying it strictly is better.

```suggestion
	// Modem doesn't come up without this for unknown reasons

Comment (TravMurav) [reply dts/msm8916/msm8939-huawei-kiwi.dts:20]: +1 for longer explanation but I'd rather like to see a link to downstream kernel sources/dts that lists this value (if such source tree is even available) or maybe an explanation how one could extract these from the decompiled dtb.

I'm not strictly opposed to linking information sources there

Comment (Newbytee) [reply dts/msm8916/msm8939-huawei-kiwi.dts:20]: How about something like this: https://github.com/LineageOS/android_kernel_huawei_kiwi/search?q=modem_id

File: dts/msm8916/msm8939-huawei-kiwi.dts [hunk 2]
```diff
@@ -16,4 +16,7 @@
...
+	huawei,modem_id = <0xFFB40105 0x0>;
```
Minecrell, line 21: The downstream kernel seems to have several different of these "modem_id"s so I'm not sure if we should just hardcode one particular one here. Might need to split this up in multiple device trees. I think someone had an opinion on that in the postmarketOS mainline chat but iirc they don't have a GitHub account...

PR #92, opened 2021-09-08
dts: Add LG Google Nexus 5 D820 and D821 (hammerhead)

wonderfulShrineMaidenOfParadise:
A possible solution to [pmaports#1227
](https://gitlab.com/postmarketOS/pmaports/-/issues/1227)
hh
pmOS and TWRP are able to boot on ``androidboot.hardware.ddr=elpida`` devices with this dts. But not on ``androidboot.hardware.ddr=hynix`` devices.

Closes #96

PR #95, opened 2021-09-15
dts: Fix broken lk2nd-msm8974-appended.dtb.img

wonderfulShrineMaidenOfParadise:
``lk2nd-appended-dtb.img`` doesn't boot with ``msm8974-lge-d855.dtb`` enabled.
``fastboot boot lk2nd-appended-dtb.img`` gets stuck and it crashes after ``fastboot flash boot lk2nd-appended-dtb.img`` and booting from hammerhead.
``compatible = "lge,d855", "qcom,msm8974", "lk2nd,device";`` is known to be what breaks ``lk2nd-msm8974-appended-dtb.img``.
So I tried to make this file in a common format.

### DIFF
--- a/dts/msm8974/msm8974-lge-d855.dts
+++ b/dts/msm8974/msm8974-lge-d855.dts
@@ -9,11 +9,13 @@
 	// This is used by the bootloader to find the correct DTB
 	qcom,msm-id = <194 118 0x10001>;
 
-	model = "LG G3 (D855)";
-	compatible = "lge,d855", "qcom,msm8974", "lk2nd,device";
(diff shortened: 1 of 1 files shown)


wonderfulShrineMaidenOfParadise:
``lk2nd-msm8974-appended.dtb.img`` is confirmed broken more than this patch can fix. See https://github.com/msm8916-mainline/lk2nd/pull/92#issuecomment-933279977 for details.
However there is no device depending on ``lk2nd-msm8974-appended.dtb.img`` for now. Closing.

PR #97, opened 2021-09-19
dts: alcatel-idol3: Add smb1360 battery detection

vknecht:


vknecht:
Closing, since no special battery detection or operation is needed in bootloader,
the smb1360 can be enabled directly in device' dts.

PR #99, opened 2021-09-19
dts: Add Samssung A5 (SM-A500H)

AriaMoradi:
This is based on: https://tinhte.vn/thread/huong-dan-chung-cai-postmarketos-len-samsung-galaxy-3g-sm-a500h-ds-a53g-aarch64-kernel-5-13.3402304/

I read the translated version by google translate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 - Motorola Moto G (2015) - osprey
 - Motorola Moto G4 Play - harpia
 - Samsung Galaxy A3 (2015) - SM-A300F, SM-A300FU
-- Samsung Galaxy A5 (2015) - SM-A500F, SM-A500FU, SM-A500YZ
+- Samsung Galaxy A5 (2015) - SM-A500F, SM-A500FU, SM-A500YZ, SM-A500H
 - Samsung Galaxy Core Max - SM-G5108Q (quirky - see comment in `dts/msm8916/msm8916-samsung-r08.dts`)
 - Samsung Galaxy Core Prime LTE - SM-G360F
 - Samsung Galaxy Grand Prime - SM-G530W
--- a/dts/msm8916/msm8216-samsung-r08.dts
+++ b/dts/msm8916/msm8216-samsung-r08.dts
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <248 0>;
+	qcom,board-id = <0xF808FF01 8>;
+
+	a53g {
+		model = "Samsung Galaxy A5 (SM-A500H)";
+		compatible = "samsung,a53g", "qcom,msm8916", "lk2nd,device";
+		lk2nd,match-bootloader = "A500H*";
+		lk2nd,smd-rpm-hack-opening;
+	};
+};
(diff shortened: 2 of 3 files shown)


dts/msm8916/msm8916-samsung-r08.dts
```diff
@@ -16,6 +16,19 @@
 		lk2nd,match-bootloader = "A500F*";
 	};
 
+	a53g {
+		model = "Samsung Galaxy A5 (SM-A500H)";
+		compatible = "samsung,a53g", "qcom,msm8916", "lk2nd,device";
+		lk2nd,match-bootloader = "A500H*";
+		lk2nd,smd-rpm-hack-opening;
```
TravMurav, line 23: Are you sure that you need this? Can you test if the device works without the rpm-hack? I'd like to avoid enabling this without a reason. (make sure to drop it in both nodes)

dts/msm8916/msm8916-samsung-r08.dts
```diff
@@ -16,6 +16,19 @@
...
+
+		samsung,muic-reset {
```
TravMurav, line 25: Also check if this is required and either drop or add to both nodes

TravMurav (changes requested):
Also please reword the commit to be similar with other commits introducing the new device

AriaMoradi:
@Baonks81 this is the code I extracted from your article, what do you think?

AriaMoradi:
I did some testing and here are my observations:
1- Without the `msm8916-samsung-r08.dts` record the model number is recognized in lk2nd and the device boots successfully(So it's probably not necessary?)
2- Without both the device model doesn't get recognized and linux fails to boot and I get a black screen(or the panel and touch keys don't turn on?)
3- `lk2nd,smd-rpm-hack-opening;` is necessary, without it I get a black screen like No. 2
4- About `muic-reset` it wasn't in `msm8216-samsung-r05.dts` in the first place, also the device uses a mainline/close to mainline kernel so I don't think it has a `CONFIG_MUIC_SUPPORT_RUSTPROOF` config so not necessary?

dts/msm8916/msm8216-samsung-r05.dts
```diff
@@ -19,4 +19,14 @@
...
 		};
 	};
+
+	a53g {
+		model = "Samsung Galaxy A5 (SM-A500H)";
+		compatible = "samsung,a53g", "qcom,msm8916", "lk2nd,device";
+		lk2nd,match-bootloader = "A500H*";
+		lk2nd,smd-rpm-hack-opening;
+
+		qcom,msm-id = <206 0>;
```
TravMurav, line 29: If, you're dropping the node from 8916 file, drop the override as well (Thinking about it, yes you should only need this node unless the bootloader is funny)

Also please squash both commits into a single one

TravMurav (changes requested):

TravMurav:
> I did some testing

1. Yes, overriding id's is usually not necessary (especially for mainline kernel, it's mostly done for compatibility with downstream)
2,3. both fail because rpm hack is missing then
4. RUSTPROOF is some downstream quirk that is not present on mainline and we have to handle it in lk2nd. You don't need it if usb works for you in lk2nd/linux

AriaMoradi:
After this change, lk2nd shows new values: `DTB - <248 0xf808ff01 0 0x5>`, before it was what we were setting before: `DTB - <206 0xce08ff01 0 0x8>`

AriaMoradi:
Fixed everything and squashed all commits.

TravMurav (approved):

TravMurav:
Fixed a small typo and merged. Thanks!

PR #106, opened 2021-10-07
arm: include/ops.h: Fix a typo

konradybcio:


TravMurav:
Sorry for letting this hang for so long. Agree with the old comment that it's probably just a droplet in the sea of caf lk but what is fun without fixing those occasionally...

PR #109, opened 2021-10-16
lk1st: msm8916: Add r69431_720p_video_panel support [DONTMERGE]

fxsheep:
This commits allow lk1st to initialize the panel. Panel auto-selection is broken, so needs manual override.
I think we need a better way to detect(select) the panel since qcom's lk isn't designed to support various devices via a single binary.


fxsheep:
> Can you be a bit more specific on which device you're using this?

Xiaomi wt88047

Minecrell:
The annoying part is that panel selection seems quite device-specific, different approaches are used to do that (based on logs I already saw probing some GPIOs, reading some magic DSI registers and more). I'm not sure I really want to deal with all that. :(

fxsheep:
> The annoying part is that panel selection seems quite device-specific, different approaches are used to do that (based on logs I already saw probing some GPIOs, reading some magic DSI registers and more). I'm not sure I really want to deal with all that. :(

Yeah, but at least we can split them to per-device. After device level, it's a question left to device maintainers, not project level )

Whatever approaches there are, we can find them all in stock aboot image. That needs some RE but it's not a big deal compared to manually reversing the display init sequences, anyways.

fxsheep:
And after all lk1st is optional - not every device has the luck to use arbitrary-signed aboot.

Minecrell:
> Yeah, but at least we can split them to per-device. After device level, it's a question left to device maintainers, not project level )

There is not a big number of "device maintainers" that we could bother here. But yes, it should be made per-device most likely by appending one of the existing device trees somehow, which already have a list of panels too. (see `dts/msm8916/msm8916-qrd7+12-v1.dts`).

> That needs some RE but it's not a big deal compared to manually reversing the display init sequences, anyways.

I would appreciate having a proof of concept for panel detection even if it works only for one particular device.

Note that unlike the panel detection there is no need to RE the display init sequences, you can just take them from the device tree on Linux. Basically, just need to generate the LK header in https://github.com/msm8916-mainline/linux-mdss-dsi-panel-driver-generator

fxsheep:
> I would appreciate having a proof of concept for panel detection even if it works only for one particular device.
I'll replicate wt88047's aboot panel detection when time allows.
> Basically, just need to generate the LK header in https://github.com/msm8916-mainline/linux-mdss-dsi-panel-driver-generator
Ah nice to hear that ) This does remind me of the EFIDroid's DTB based display driver.

stephan-gh:
I extended the panel driver generator to also produce a header file for LK, and opened https://github.com/msm8916-mainline/lk2nd/pull/139 as a replacement for this PR. There is no automatic detection of panels at the moment but the panel can be selected on the make command line.

@fxsheep Can you test #139 if you're still interested in this?

PR #114, opened 2021-10-31
dts: msm8226: Add Huawei Watch (sturgeon)

z3ntu:
Props to @MagneFire for testing

image

Text on screen is flipped vertically, maybe something like this needs to be triggered? https://github.com/msm8916-mainline/lk2nd/commit/64eeff989c028601e96a97cb1902766ba64e1bca Not really too important though imo

### DIFF
--- a/dts/msm8226/apq8026-huawei-sturgeon.dts
+++ b/dts/msm8226/apq8026-huawei-sturgeon.dts
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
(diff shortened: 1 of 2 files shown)


TravMurav (approved):

Minecrell:
> Text on screen is flipped vertically, maybe something like this needs to be triggered? 64eeff9 Not really too important though imo

We're not touching the display configuration with the "continuous splash" feature, we just take it over from the previous bootloader. The funny thing is that this means that they had to modify the splash image to be vertically flipped (so it shows up correctly).

I guess we could touch the display configuration to fix this but maybe that's something for later...

PR #115, opened 2021-11-01
Introduce Multi-DTB container support

TravMurav:
lk2nd already implements a lot of checks to definitively detect which specific device model it runs on.
This means that lk2nd could also use those checks to choose the DTB from a container instead of loading the single appended one. Implementing this would allow us to make more generic system images that could boot on all devices within a supported platform (platforms?).

The proposed approach is to use FDT based container for DTB images: This allows re-using libfdt to just select a node within the container by it's `compatible`. Then install this container as appended-dtb to re-use most of the boot path and just insert a check for unpacking the container.

### DIFF
--- a/platform/msm_shared/dev_tree.c
+++ b/platform/msm_shared/dev_tree.c
@@ -39,6 +39,10 @@
 #include <kernel/thread.h>
 #include <lk2nd.h>
 
+#if WITH_LK2ND
+#include <grove.h>
+#endif
+
 struct dt_entry_v1
 {
 	uint32_t platform_id;
@@ -471,6 +475,8 @@ void *dev_tree_appended(void *kernel, uint32_t kernel_size, uint32_t dtb_offset,
 	struct dt_entry_node *dt_node_tmp1 = NULL;
 	struct dt_entry_node *dt_node_tmp2 = NULL;
 	unsigned dtb_count = 0;
+	int ret = 0;
+
 
 	/* Initialize the dtb entry node*/
 	dt_entry_queue = (struct dt_entry_node *)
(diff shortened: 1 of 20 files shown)


platform/msm_shared/dev_tree.c
```diff
@@ -561,7 +567,23 @@ void *dev_tree_appended(void *kernel, uint32_t kernel_size, uint32_t dtb_offset,
 	}
 
 	if(bestmatch_tag) {
-		memcpy(tags, bestmatch_tag, bestmatch_tag_size);
+#if WITH_LK2ND
+		/* If the DTB is a container, unpack it. */
+		/* Most trees < 64k so 128k should be enough. */
+		memcpy(tags + 1024*128, bestmatch_tag, bestmatch_tag_size);
```
minlexx, line 573: Instead of using semi-magical numbers (1024, 128) add some human-readable constant/macro and use it in all places

scripts/forester.py
```diff
@@ -0,0 +1,175 @@
...
+    apq8016-sbc {
+        compatible = "qcom,apq8016-sbc", "qcom,apq8016";
+        compression = "none";
+        dtb = /incbin/("arch/arm64/dts/qcom/apq8016-sbc.dtb");
+    };
+
+    msm8916-samsung-gt58 {
+        compatible = "samsung,gt58", "qcom,msm8916";
+        compression = "gzip";
+        dtb = /incbin/("arch/arm64/dts/qcom/msm8916-samsung-gt58.dtb.gz");
```
minlexx, line 26: is `/incbin/` something standard for fdt..?

TravMurav:
After quite a bit of research and experimentation it's clear that we don't need a yet another standard and I think the scope of lk2nd may change a bit in the future. So I'm obsoleting this idea. Thanks for the input on this!

PR #123, opened 2021-12-27
Improve lk1st support for DB410c

stephan-gh:
I use `make TOOLCHAIN_PREFIX=arm-none-eabi- lk1st-msm8916 DISABLE_EMMC=1 LK1ST_PANEL=ADV7533_1080P_VIDEO_PANEL` to build a pretty nice lk1st for DB410c with display support. :)

See commit log.


platform/msm_shared/i2c_qup.c
```diff
@@ -48,6 +48,11 @@
 #include <platform/timer.h>
 #include <platform/interrupts.h>
 
+#if DEBUGLEVEL > INFO
+#undef DEBUGLEVEL
+#define DEBUGLEVEL INFO
+#endif
```
minlexx, line 54: I don't fully understand this change. What is this about?

target/msm8916/init.c
```diff
@@ -115,6 +115,7 @@ void target_sdc_init()
 	struct mmc_config_data config;
 	struct mmc_config_data sd_config;
 
+#ifndef DISABLE_EMMC
```
TravMurav, line 118: This looks more like "disable first emmc"? but I assumed it disables all mmc's from the name.
`USE_SECONDARY_MMC` maybe?

TravMurav (commented):
Other than a naming nitpick, it looks fine to me

TravMurav (approved):

PR #127, opened 2022-01-02
aboot: fastboot-extra: Add "fastboot oem readl&writel&readb&writeb"

fxsheep:
Add fastboot oem commands to read/write a specified address.

Usage:

```
fastboot oem readl <address>
fastboot oem readb <address>
fastboot oem writel <address> <value>
fastboot oem writeb <address> <value>
```

### DIFF
--- a/app/aboot/fastboot-extra.c
+++ b/app/aboot/fastboot-extra.c
@@ -10,8 +10,81 @@
 #include <reg.h>
 #include <smd.h>
 #include <smem.h>
+#include <stdlib.h>
 #include <target.h>
 #include "fastboot.h"
+
+static bool parse_write_args(const char *arg, uint32_t *addr, uint32_t *value)
+{
+	char *saveptr;
+	char *args = strdup(arg);
+	char *addr_str = strtok_r(args, " ", &saveptr);
+	char *value_str = strtok_r(NULL, " ", &saveptr);
+
+	if (!addr_str || !value_str) {
+		free(args);
+		return false;
+	}
+
+	*addr = atoi(addr_str);
+	*value = atoi(value_str);
+	free(args);
+	return true;
+}
+
+static void cmd_oem_readl(const char *arg, void *data, unsigned sz)
+{
+	char response[MAX_RSP_SIZE];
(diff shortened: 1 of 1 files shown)


app/aboot/fastboot-extra.c
```diff
@@ -1,3 +1,4 @@
+#include <stdlib.h>
```
stephan-gh, line 1: The include list is alphabetically ordered

app/aboot/fastboot-extra.c
```diff
@@ -13,6 +14,87 @@
...
+		}
+		if (arg[i] == ' ' && i <= 10 && address[0] == 0) {
+			strncpy(address, arg, i);
+			addr = atoi(address);
+		}
+		if (arg[i] == ' ' && arg[i + 1] != ' ') {
+			value = atoi(arg + i + 1);
+			break;
+		}
+	}
```
stephan-gh, line 48: Looks like we can share this code between writel and writeb. I'm also a bit confused about how it works... Perhaps we should simply use the standard `strtok` here? Something like this:

```c
static bool parse_write_args(const char *arg, uint32_t *addr, uint32_t *value)
{
	char *saveptr;
	char *args = strdup(arg);
	char *addr_str = strtok_r(args, " ", &saveptr);
	char *value_str = strtok_r(NULL, " ", &saveptr);

	if (!addr_str || !value_str) {
		free(args);
		return false;
	}

	*addr = atoi(addr_str);
	*value = atoi(value_str);
	free(args);
	return true;
}
```

stephan-gh (commented):
Thanks for the PR! Just two small comments.

app/aboot/fastboot-extra.c
fxsheep: > we can share this code between writel and writeb. 

Yeah, I thought about that too. 

>we should simply use the standard strtok here

Ah, I thought that won't work with more than one space. Turned out I thought too much.
Thanks for the review!

stephan-gh (approved):

TravMurav (approved):

PR #130, opened 2022-01-08
dev_tree: Fix appending cmdline if "bootargs" is already in DT

### LINKED ISSUE #121: [SM-A300FU] LineageOS 17 stuck on splash screen when booted by lk2nd
Here's what I did on my Galaxy A3 2015:
- I installed TWRP on the recovery partition with Heimdall
- I installed LineageOS 17.1 which, by itself, booted just fine.
- With Heimdall, I flashed lk2nd to the boot partition
- postmarketOS now boots fine from an SD Card, obviously LineageOS does not boot at this point
- I flashed the LineageOS boot partition in lk2nd with `fastboot flash boot boot.img`

Now pmOS still boots fine, but when I take out the SD Card, LineageOS first appears to boot, but it gets stuck on its boot logo / animation forever. Flashing its own boot partition in place of lk2nd using Heimdall makes it work again.

I wanted to set up dual boot so that I could boot LineageOS from the internal storage with no SD Card, or insert an SD Card to boot postmarketOS. But right now I have to re-flash the boot partition with either lk2nd or the LOS boot every time I want to switch operating systems.

stephan-gh:
Downstream device trees contain bootargs = "sched_enable_hmp=1".
The original intention was that this would get prepended to the real
command line. However, this currently only works with a modification
in libfdt (see 5ac5542029fa0706829a30247445a998ad6cb93c) and has
therefore been broken since the libfdt update. Booting downstream was
effectively broken because the proper cmdline was missing.

Instead of modifying libfdt, fix this properly by replacing the null
terminator with a space.

Might fix #121

### DIFF
--- a/platform/msm_shared/dev_tree.c
+++ b/platform/msm_shared/dev_tree.c
@@ -1413,6 +1413,13 @@ int update_device_tree(void *fdt, const char *cmdline,
 	offset = ret;
 	if (cmdline)
 	{
+		int len;
+		char *oldargs = fdt_getprop_w(fdt, offset, "bootargs", &len);
+
+		/* Replace old null terminator so the strings are concatenated */
+		if (oldargs && len >= 1 && oldargs[len-1] == '\0')
+			oldargs[len-1] = ' ';
+
 		/* Adding the cmdline to the chosen node */
 		ret = fdt_appendprop_string(fdt, offset, (const char*)"bootargs", (const void*)cmdline);
 		if (ret)


minlexx:
So, first it appends string after the null terminator byte, and then replaces null terminator char by space char back in the string?

stephan-gh:
> So, first it appends string after the null terminator byte, and then replaces null terminator char by space char back in the string?

Yes, but I changed the order now. Now it replaces the null terminator with a space first and then appends the other string (which will have its own null terminator).

TravMurav (approved):

PR #131, opened 2022-01-09
dev_tree: Allow mapping power key

timbz:
Allows entering fastboot mode on the Huawei Watch using the (single) power button.

### DIFF
--- a/dts/lk2nd.h
+++ b/dts/lk2nd.h
@@ -30,5 +30,6 @@
 #define GPIO_ACTIVE_HIGH (1 << 8)
 
 #define LK2ND_KEY_RESIN (2 << 16) 0
+#define LK2ND_KEY_PWR (3 << 16) 0
 
 #endif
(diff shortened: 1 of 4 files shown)


TravMurav (approved):
This looks good to me, I suppose you've tested already that resin isn't tied to the same button? (on lenok they're both tied together so it works "as is" by holding the key long enough)

timbz:
> This looks good to me, I suppose you've tested already that resin isn't tied to the same button? (on lenok they're both tied together so it works "as is" by holding the key long enough)

Ye I did. I tried all sorts of patterns. There was no way to enter fastboot mode for me.

stephan-gh:
Can you split this into two commits, one for `dts: ...` and one for `target: keys_lk2nd: ...` or something like that? The `dev_tree: ...` does not seem right.

timbz:
> Can you split this into two commits, one for `dts: ...` and one for `target: keys_lk2nd: ...` or something like that? The `dev_tree: ...` does not seem right.

done

PR #132, opened 2022-01-09
README: Add better introduction and various clarifications

stephan-gh:
This is supposed to be a better introduction for lk2nd, without relying on too much inside knowledge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,39 @@
 # lk2nd
-lk2nd is a bootloader for Qualcomm MSM devices (at the moment only MSM8916 and MSM8939),
-based on the [CodeAurora Little Kernel fork](https://source.codeaurora.org/quic/la/kernel/lk/).
-It provides an Android Fastboot interface on devices where the stock bootloader
-does not provide Fastboot (e.g. Samsung).
+lk2nd is a custom bootloader for many different devices based on Qualcomm
+system-on-chips (SoCs), including smartphones, tablets, smartwatches and a few
+single board computers (SBCs). It simplifies booting standard operating systems
+by providing:
 
-On MSM8916 it is also used for some quirks for mainline devices, e.g. to set
-a proper WiFi/BT MAC address in the device tree.
+  - **A unified boot and flashing interface** (Android Fastboot), independent of
+    device-specific modifications made to the boot chain (e.g. the proprietary
+    flashing interface on Samsung devices)
+  - **Automatic device and hardware detection** (e.g. display panels, batteries, ...)
+  - **Additional bootloader features** (e.g. file system boot, booting from SD card)
+  - **Additional firmware features** (e.g. booting secondary CPU cores if the firmware
+    lacks [PSCI] support)
+  - **Workarounds for various device-specific issues** (e.g. USB on some Samsung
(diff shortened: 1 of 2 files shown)


minlexx (approved):

README.md
```diff
@@ -1,16 +1,38 @@
...
-does not provide Fastboot (e.g. Samsung).
-
-On MSM8916 it is also used for some quirks for mainline devices, e.g. to set
-a proper WiFi/BT MAC address in the device tree.
-
-lk2nd does not replace the stock bootloader. It is packaged into an Android
-boot image and then loaded by the stock bootloader as a "secondary" bootloader.
-The real Android boot image is placed into the boot partition with 512 KiB offset,
-and then loaded by lk2nd.
+# lk2nd (and lk1st)
```
TravMurav, line 1: Maybe call the project lk2nd and refer to lk1st as it's configuration variant? Thus the first paragraph will refer to lk2nd as a bigger entity that includes lk1st and whatever else?

They are *the same* bootloader, just configured to boot differently.

README.md
```diff
@@ -1,16 +1,38 @@
...
+  - **A unified boot and flashing interface** (Android Fastboot), independent of
+    device-specific modifications made to the boot chain (e.g. the proprietary
+    flashing interface on Samsung devices)
+  - **Automatic device and hardware detection** (e.g. display panels, batteries, ...)
+  - **Additional bootloader features** (e.g. file system boot, booting from SD card)
+  - **Additional firmware features** (e.g. booting secondary CPU cores if the firmware
+    lacks [PSCI] support)
+  - **Workarounds for various device-specific issues** (e.g. USB on some Samsung
+    devices, modem on Motorola smartphones, ...)
+  - **Various debugging commands** that allow better understanding of the hardware
```
TravMurav, line 16:
```suggestion
  - **Various debugging commands** that allow better understanding of the hardware
      for research and development purposes
```

clarification seems useful there as this is a rather niche feature that few people actually will rely on.

README.md
```diff
@@ -129,10 +151,13 @@ $ make TOOLCHAIN_PREFIX=arm-none-eabi- lk2nd-msmXXXX
   - Arch Linux: `dtc`
 
 Replace `TOOLCHAIN_PREFIX` with the path to your tool chain.
-`lk2nd.img` is built and placed into `build-msm8916-secondary/lk2nd.img`.
+`lk2nd.img` is built and placed into `build-lk2nd-msmXXXX/lk2nd.img`.
+
+**Note:** Unlike lk2nd, lk1st is still experimental and therefore not described
```
TravMurav, line 156: tbh I'd prefer to not list an experimental feature in the first description paragraphs, maybe move it's description down to here, so it's only lk2nd until there and then "yeah, you can also replace aboot if you know stuff but this is WIP and experimental"?

stephan-gh: I think lk1st is pretty close and I do say in the introduction that it's for SBCs and expert users only. The only reason I didn't describe it yet is that it needs some careful thought on how to prevent users from copying those commands and destroying their bootloader when they really just want lk2nd.

TravMurav (approved):

PR #133, opened 2022-01-14
Make lk1st really great

stephan-gh:
This set of commits makes lk1st really great (I think). It makes it possible to select a device dtb using `LK1ST_DTB` and adds display support using the https://github.com/msm8916-mainline/linux-mdss-dsi-panel-driver-generator.

Usage for wingtech-wt88047 would be e.g. like:

```
make TOOLCHAIN_PREFIX=arm-none-eabi- lk1st-msm8916 LK1ST_DTB=msm8916-qrd7+12-v1 LK1ST_COMPATIBLE=wingtech,wt88047 LK1ST_PANEL=r61308_720p_video
```

See commit list for details.


minlexx:
Wow, `29 files +1,987 −114` I hope resulting binary still fits in 512k? :)

stephan-gh:
Of course. :) Most of the big diff is auto-generated with the panel driver generator and should(TM) only be used when the panel is selected.

stephan-gh:
@greatofdream The main branch is a completely new version of lk2nd. There the equivalent flag  is called `LK2ND_BUNDLE_DTB=`, see https://github.com/msm8916-mainline/lk2nd/blob/main/Documentation/building.md

PR #135, opened 2022-01-15
dts: huawei-y635-l01: Add panel detection for Huawei Y635-L01

ungeskriptet:


stephan-gh:
Pushed to master, thanks!

Issue #134, opened 2022-01-15
[SM-J500M] Future support for this device?


mintphin:
Hey everyone! I have a Galaxy J5 2015 which its model is SM-J500M. As you may have expected it's not on the list of supported devices. My question is, will we see support for more models of the Galaxy J5 2015 in the foreseeable future? And is there a way in which I can contribute? Thank you so much.

mintphin:
Update: I have tried flashing it from another device and it works! Only issue is that on the fastboot menu it doesn't detect the model correctly, but otherwise it's running just fine!

mintphin:
Another update! The recovery menu doesn't work. Instead of loading the screen does this:

![image](https://user-images.githubusercontent.com/26839383/149611673-015a9307-838d-472d-97f9-f1d8ae83e513.png)
 
Also when plugged in to a computer while in fastboot mode its not being listed when running `fastboot devices`.
So in a nutshell, work needs to be done before it can be considered as supported. And lastly, repeating the question on my first post, is there any way I can help with this info?

mintphin:
Ok, after a bit of help from the mainline channel on postmarketOS I managed to get it working except for the recovery which just boots the default kernel. I will PR my changes anyways!

PR #136, opened 2022-01-15
dts: msm8916: Added support for SM-J500M

mintphin:
This PR adds support for the Samsung Galaxy J5 2015 (SM-J500M). Check #134 for more info.

IMG_20220115_190620

IMG_20220115_191313


TravMurav (approved):
The compatible (`"samsung,j5lte-latam"`) might need to be changed later (all of them needs to be cleaned up and carefully synced with Linux as part of the generic images work), but for now it looks good, 
Photos are also nice, consider adding them to the pmOS wiki page :)
Thanks!

PR #137, opened 2022-01-17
Improve behavior for command mode displays

stephan-gh:
These changes reduce the lag with screen updates on command mode displays by using an extra thread. Also, a `fastboot oem display-auto-refresh` is added to explicitly enable the auto refresh feature in the hardware for command mode displays. Unfortunately, we cannot enable it by default at the moment since it seems unreliable on some panels and/or MSM8974.

### DIFF
--- a/app/aboot/fastboot-extra.c
+++ b/app/aboot/fastboot-extra.c
@@ -2,10 +2,12 @@
 #include <debug.h>
 #include <dev/fbcon.h>
 #include <malloc.h>
+#include <mdp5.h>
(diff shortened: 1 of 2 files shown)


TravMurav (approved):
I see no obvious problems here, however I'd like to eventually see the autorefresh being enabled by default on as many targets as possible so we can have working fb on cmd mode panels in linux, e.g. for the purposes of the fb takeover (that I hope we will have eventually) Maybe in the future this should be a compile time flag? Or a switch in the project makefile?

PR #141, opened 2022-01-27
Implement oem dump-speedbin for msm8974

z3ntu:
Loosely based on lk2nd implementation for msm8916 and on Linux mainline
get_krait_bin_format_b.

Example output:
```
$ fastboot oem dump-speedbin
(bootloader) PVS valid: 1, Redundant fuse select: 0
(bootloader) Speed bin: 1, PVS: 12, PVS version: 1 (speed1-pvs12-v1)
```

@minlexx Would be great if you could also test this on some devices that you have (especially 8974-AC if possible). So far I've only seen speed bin 1, different PVS (2, 8, 9, 10, 12, 13) and different PVS version (0, 1). 8974-AC should be speed bin 3.

### DIFF
--- a/app/aboot/fastboot-msm8974.c
+++ b/app/aboot/fastboot-msm8974.c
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <reg.h>
+#include <stdio.h>
+#include "fastboot.h"
+
+#define EFUSE	0xfc4b80b0
+
+#define readq(a)	(*REG64(a))
+
+static void cmd_oem_dump_speedbin(const char *arg, void *data, unsigned sz)
+{
+	char response[MAX_RSP_SIZE];
+	uint32_t redundant_fuse_select, speed_bin, pvs, pvs_version, pvs_valid;
+	/*
+	 * Format of efuse (8 bytes):
+	 * [53] PVS valid
+	 * [31] PVS (bit 3)
+	 * [29:27] redundant fuse
+	 * [26:24] redundant fuse select
+	 * [8:6] PVS (bit 2 & 1)
+	 * [5:4] PVS version
+	 * [2:0] speed bin
+	 */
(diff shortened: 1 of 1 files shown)


minlexx:
samsung-klte:
```
(bootloader) PVS valid: 1, Redundant fuse select: 0
(bootloader) Speed bin: 3, PVS: 4, PVS version: 1 (speed3-pvs4-v1)
```

minlexx:
lg-hammerhead:
```
(bootloader) PVS valid: 1, Redundant fuse select: 0
(bootloader) Speed bin: 1, PVS: 2, PVS version: 0 (speed1-pvs2-v0)
```

z3ntu:
Perfect, works as expected!

minlexx (approved):

app/aboot/fastboot-msm8974.c
```diff
@@ -0,0 +1,56 @@
+#include <reg.h>
```
stephan-gh, line 1: Would prefer having an explicit SPDX license header here, you can choose between `GPL-2.0-only` (most lk2nd-specific things), `BSD-3-Clause` (qcom's additions to LK) and `MIT` (original LK). We have it all covered :D

I know some source files are currently missing it, need to fix it in the file where it's missing.

Also, if you have copied substantial code as-is from somewhere you may want to add the original copyright texts and use the same license.

z3ntu: Added GPL-2.0-only.

> Also, if you have copied substantial code as-is from somewhere you may want to add the original copyright texts and use the same license.

Definitely not substantial, the closest is the couple of lines doing the bit shifting but that's just the way how to get the required bits from the variable and very trivial.

stephan-gh (approved):

TravMurav (approved):

minlexx (approved):

PR #142, opened 2022-02-03
master+msm8939 2nd cluster poweron

bryanodonoghue:
Doing a bit of hackary on msm8939 its possible to get both clusters and all eight cores booted.

msm8936/msm8939 comes in a big.LITTLE 2x4 A53 cluster configuration.
Each core has its own L1 cache
All four cores in a given cluster, share an L2 cache

BootROM needs to enable the L2 cache for the cluster the boot CPU lives in, however it does not enable for the second cluster.
msm8916 has these same L2 enable registers but since it has only one cluster, there's no need to take these specific steps to switch the L2 on for msm8916.

lk2nd as is, switches on four cores in the primary cluster for me via the spin-table hack.
This series then extends that out to all eight cores.

Reference DTS

    cpus {
                #address-cells = <1>;
                #size-cells = <0>;

                CPU0: cpu@100 {
                        device_type = "cpu";
                        compatible = "arm,cortex-a53", "arm,armv8";
                        enable-method = "spin-table";
                        reg = <0x100>;
                        next-level-cache = <&L2_1>;
                        qcom,acc = <&acc0>;
                        qcom,saw = <&saw0>;
                        cpu-idle-states = <&CPU_SPC>;
                        clocks = <&apcs1_mbox>;
                        #cooling-cells = <2>;
                        L2_1: l2-cache-cluster1 {
                                compatible = "cache";
                                cache-level = <2>;
                                reg = <0x0b011000 0x1000>;
                        };
                };

                CPU1: cpu@101 {
                        device_type = "cpu";
                        compatible = "arm,cortex-a53", "arm,armv8";
                        enable-method = "spin-table";
                        reg = <0x101>;
                        next-level-cache = <&L2_1>;
                        qcom,acc = <&acc1>;
                        qcom,saw = <&saw1>;
                        clocks = <&apcs1_mbox>;
                        #cooling-cells = <2>;
                };

                CPU2: cpu@102 {
                        device_type = "cpu";
                        compatible = "arm,cortex-a53", "arm,armv8";
                        enable-method = "spin-table";
                        reg = <0x102>;
                        next-level-cache = <&L2_1>;
                        qcom,acc = <&acc2>;
                        qcom,saw = <&saw2>;
                        clocks = <&apcs1_mbox>;
                        #cooling-cells = <2>;
                };

                CPU3: cpu@103 {
                        device_type = "cpu";
                        compatible = "arm,cortex-a53", "arm,armv8";
                        enable-method = "spin-table";
                        reg = <0x103>;
                        next-level-cache = <&L2_1>;
                        qcom,acc = <&acc3>;
                        qcom,saw = <&saw3>;
                        clocks = <&apcs1_mbox>;
                        #cooling-cells = <2>;
                };

                CPU4: cpu@0 {
                        device_type = "cpu";
                        compatible = "arm,cortex-a53", "arm,armv8";
                        enable-method = "spin-table";
                        reg = <0x0>;
                        qcom,acc = <&acc4>;
                        qcom,saw = <&saw4>;
                        clocks = <&apcs0>;
                        #cooling-cells = <2>;
                        next-level-cache = <&L2_0>;
                        L2_0: l2-cache-cluster0 {
                                compatible = "cache";
                                cache-level = <2>;
                                reg = <0x0b111000 0x1000>;
                        };
                };

                CPU5: cpu@1 {
                        device_type = "cpu";
                        compatible = "arm,cortex-a53", "arm,armv8";
                        enable-method = "spin-table";
                        reg = <0x1>;
                        next-level-cache = <&L2_0>;
                        qcom,acc = <&acc5>;
                        qcom,saw = <&saw5>;
                        clocks = <&apcs0>;
                        #cooling-cells = <2>;
                };

                CPU6: cpu@2 {
                        device_type = "cpu";
                        compatible = "arm,cortex-a53", "arm,armv8";
                        enable-method = "spin-table";
                        reg = <0x2>;
                        next-level-cache = <&L2_0>;
                        qcom,acc = <&acc6>;
                        qcom,saw = <&saw6>;
                        clocks = <&apcs0>;
                        #cooling-cells = <2>;
                };

                CPU7: cpu@3 {
                        device_type = "cpu";
                        compatible = "arm,cortex-a53", "arm,armv8";
                        enable-method = "spin-table";
                        reg = <0x3>;
                        next-level-cache = <&L2_0>;
                        qcom,acc = <&acc7>;
                        qcom,saw = <&saw7>;
                        clocks = <&apcs0>;
                        #cooling-cells = <2>;
                };
    };

Partial bootlog timestamp 8180 is where the second cluster gets powered on, 8210 shows other cores in that cluster, skipping L2 init.

[8050] Booting CPU100
[8070] CPU (256) L2 cache already powered-up
[8080] Skipping boot of current CPU (256)
[8090] Booting CPU101
[8110] CPU (257) L2 cache already powered-up
[8120] Booting CPU102
[8140] CPU (258) L2 cache already powered-up
[8150] Booting CPU103
[8170] CPU (259) L2 cache already powered-up
**[8180] Booting CPU0
[8210] Booting CPU1
[8230] CPU (1) L2 cache already powered-up**
[8250] Booting CPU2
[8270] CPU (2) L2 cache already powered-up
[8280] Booting CPU3
[8300] CPU (3) L2 cache already powered-up
[8310] Updating device tree: done
[8320] booting linux @ 0x80000000, ramdisk @ 0x82000000 (13356916), tags/device tree @ 0x81e00000
[8330] Jumping to kernel via monitor
[8330] HVC call: 0x200010f failed with :fffffffe
[8330] Falling back to SCM call
[    0.000000] Booting Linux on physical CPU 0x0000000100 [0x410fd034]
[    0.000000] Linux version 5.17.0-rc1-00025-g9cdd53362b37 (deckard@sagittarius-a) (aarch64-linux-gnu-gcc (Linaro GCC 7.4-2019.02) 7.4.1 20181213 [linaro-7.4-2019.02 revision 56ec6f6b2
[    0.000000] Machine model: SuperDolly FakeName msm8939
[    0.000000] earlycon: msm_serial_dm0 at MMIO 0x00000000078b0000 (options '')
[    0.000000] printk: bootconsole [msm_serial_dm0] enabled
[    0.000000] efi: UEFI not found.
[    0.000000] NUMA: No NUMA configuration found
[    0.000000] NUMA: Faking a node at [mem 0x0000000080000000-0x00000000ffffffff]
[    0.000000] NUMA: NODE_DATA [mem 0xffbd3040-0xffbd4fff]
[    0.000000] Zone ranges:
[    0.000000]   DMA      [mem 0x0000000080000000-0x00000000ffffffff]
[    0.000000]   DMA32    empty
[    0.000000]   Normal   empty
[    0.000000] Movable zone start for each node
[    0.000000] Early memory node ranges
[    0.000000]   node   0: [mem 0x0000000080000000-0x0000000085ffffff]
[    0.000000]   node   0: [mem 0x0000000086000000-0x0000000089dfffff]
[    0.000000]   node   0: [mem 0x0000000089e00000-0x000000008e9fffff]
[    0.000000]   node   0: [mem 0x000000008ea00000-0x000000008eafffff]
[    0.000000]   node   0: [mem 0x000000008eb00000-0x00000000ffffffff]
[    0.000000] Initmem setup node 0 [mem 0x0000000080000000-0x00000000ffffffff]
[    0.000000] cma: Reserved 32 MiB at 0x00000000fb800000
[    0.000000] percpu: Embedded 28 pages/cpu s74456 r8192 d32040 u114688
[    0.000000] Detected VIPT I-cache on CPU0
[    0.000000] CPU features: detected: ARM erratum 843419
[    0.000000] CPU features: detected: ARM erratum 845719
[    0.000000] Fallback order for Node 0: 0 
[    0.000000] Built 1 zonelists, mobility grouping on.  Total pages: 516096
[    0.000000] Policy zone: DMA
[    0.000000] Kernel command line: console=ttyMSM1,115200,n8 earlycon=msm_serial_dm,0x78b0000 rootwait rw root=/dev/ram0 no_console_suspend
[    0.000000] Dentry cache hash table entries: 262144 (order: 9, 2097152 bytes, linear)
[    0.000000] Inode-cache hash table entries: 131072 (order: 8, 1048576 bytes, linear)
[    0.000000] mem auto-init: stack:off, heap alloc:off, heap free:off
[    0.000000] Memory: 1921068K/2097152K available (14016K kernel code, 2104K rwdata, 4968K rodata, 3072K init, 412K bss, 143316K reserved, 32768K cma-reserved)
[    0.000000] SLUB: HWalign=64, Order=0-3, MinObjects=0, CPUs=8, Nodes=1
[    0.000000] trace event string verifier disabled
[    0.000000] rcu: Preemptible hierarchical RCU implementation.
[    0.000000] rcu:     RCU restricting CPUs from NR_CPUS=64 to nr_cpu_ids=8.
[    0.000000]  Trampoline variant of Tasks RCU enabled.
[    0.000000]  Tracing variant of Tasks RCU enabled.
[    0.000000] rcu: RCU calculated value of scheduler-enlistment delay is 25 jiffies.
[    0.000000] rcu: Adjusting geometry for rcu_fanout_leaf=16, nr_cpu_ids=8
[    0.000000] NR_IRQS: 64, nr_irqs: 64, preallocated irqs: 0
[    0.000000] Root IRQ handler: gic_handle_irq
[    0.000000] random: get_random_bytes called from start_kernel+0x4f0/0x758 with crng_init=0
[    0.000000] arch_timer: cp15 and mmio timer(s) running at 19.20MHz (virt/virt).
[    0.000000] clocksource: arch_sys_counter: mask: 0xffffffffffffff max_cycles: 0x46d987e47, max_idle_ns: 440795202767 ns
[    0.000002] sched_clock: 56 bits at 19MHz, resolution 52ns, wraps every 4398046511078ns
[    0.011338] Console: colour dummy device 80x25
[    0.018893] Calibrating delay loop (skipped), value calculated using timer frequency.. 38.40 BogoMIPS (lpj=76800)
[    0.023272] pid_max: default: 32768 minimum: 301
[    0.033689] LSM: Security Framework initializing
[    0.038337] SELinux:  Initializing.
[    0.043112] Mount-cache hash table entries: 4096 (order: 3, 32768 bytes, linear)
[    0.046161] Mountpoint-cache hash table entries: 4096 (order: 3, 32768 bytes, linear)
[    0.057531] cblist_init_generic: Setting adjustable number of callback queues.
[    0.061574] cblist_init_generic: Setting shift to 3 and lim to 1.
[    0.068746] cblist_init_generic: Setting shift to 3 and lim to 1.
[    0.075156] rcu: Hierarchical SRCU implementation.
[    0.082670] EFI services will not be available.
**[    0.086645] smp: Bringing up secondary CPUs ...
[    0.091135] Detected VIPT I-cache on CPU1
[    0.091310] CPU1: Booted secondary processor 0x0000000101 [0x410fd034]
[    0.092489] Detected VIPT I-cache on CPU2
[    0.092555] CPU2: Booted secondary processor 0x0000000102 [0x410fd034]
[    0.093519] Detected VIPT I-cache on CPU3
[    0.093570] CPU3: Booted secondary processor 0x0000000103 [0x410fd034]
[    0.095360] Detected VIPT I-cache on CPU4
[    0.097001] CPU4: Booted secondary processor 0x0000000000 [0x410fd034]
[    0.101823] Detected VIPT I-cache on CPU5
[    0.103389] CPU5: Booted secondary processor 0x0000000001 [0x410fd034]
[    0.108182] Detected VIPT I-cache on CPU6
[    0.109750] CPU6: Booted secondary processor 0x0000000002 [0x410fd034]
[    0.114477] Detected VIPT I-cache on CPU7
[    0.116045] CPU7: Booted secondary processor 0x0000000003 [0x410fd034]
[    0.119323] smp: Brought up 1 node, 8 CPUs**
[    0.191854] SMP: Total of 8 processors activated.
[    0.195922] CPU features: detected: 32-bit EL0 Support
[    0.200718] CPU features: detected: CRC32 instructions
[    0.207056] CPU features: emulated: Privileged Access Never (PAN) using TTBR0_EL1 switching
[    0.210887] CPU: All CPU(s) started at EL1
[    0.219582] alternatives: patching kernel code

### DIFF
--- a/lk2nd/smp/cpu-boot.c
+++ b/lk2nd/smp/cpu-boot.c
@@ -23,6 +23,48 @@
 #define APC_PWR_GATE_CTL_GHDS_EN	BIT(0)
 #define APC_PWR_GATE_CTL_GHDS_CNT(cnt)	((cnt) << 24)
 
+/*
+ * Each CPU within a cluster has a clock controller.
+ * There is just once L2 cache controller per cluster. We can take the base
+ * address of a given CPU clock controller and using a combination of mask
+ * and offset find the base address of the L2 cache controller.
+ *
+ * APCS_GLB_BASE takes the base of given CPU clock controller and offsets
+ * back the single APCS_GLB register which is the base for the L2
+ * cache controller in the cluster.
+ *
+ */
+#define APCS_GLB_MASK			0xfff00000
+#define APCS_GLB_OFFSET			0x00011000
+#define APCS_GLB_BASE(base)		(((base) & APCS_GLB_MASK) + \
+					APCS_GLB_OFFSET)
+
+#define APCS_PWR_CTL_OVERRIDE			0xc
+#define APCS_PWR_CTL_OVR_PRESETDBG		BIT(22)
+
+#define APCS_L2_PWR_CTL				0x14
+#define APCS_L2_PWR_CTL_PMIC_APC_ON		BIT(28)
+#define APCS_L2_PWR_CTL_L2_HS_CNT(cnt)		((cnt) << 16)
+#define APCS_L2_PWR_CTL_L2_ARRAY_HS_CLAMP	BIT(15)
+#define APCS_L2_PWR_CTL_SCU_ARRAY_HS_CLAMP	BIT(14)
+#define APCS_L2_PWR_CTL_L2_RET_SLP		BIT(13)
+#define APCS_L2_PWR_CTL_SYS_RESET		BIT(12)
+#define APCS_L2_PWR_CTL_L2_SLEEP_STATE		BIT(11)
+#define APCS_L2_PWR_CTL_L2_HS_RST		BIT(10)
+#define APCS_L2_PWR_CTL_L2_HS_EN		BIT(9)
+#define APCS_L2_PWR_CTL_L2_HS_CLAMP		BIT(8)
+#define APCS_L2_PWR_CTL_L2_RST_DIS		BIT(2)
+#define APCS_L2_PWR_CTL_SCU_ARRAY_HS		BIT(1)
+#define APCS_L2_PWR_CTL_L2_ARRAY_HS		BIT(0)
+
+#define APCS_L2_PWR_STATUS		0x18
+#define APCS_L2_PWR_STATUS_L2_HS_STS	BIT(9)
+
+#define APCS_L2_CORE_CBCR		0x58
+#define APCS_L2_CORE_CBCR_CLK_OFF	BIT(31)
+#define APCS_L2_CORE_CBCR_HW_CTL	BIT(1)
+#define APCS_L2_CORE_CBCR_CLK_ENABLE	BIT(0)
+
 #define QCOM_SCM_BOOT_SET_ADDR		0x01
 #define QCOM_SCM_BOOT_FLAG_COLD_ALL	(0 | BIT(0) | BIT(3) | BIT(5))
 #define QCOM_SCM_BOOT_SET_ADDR_MC	0x11
@@ -56,6 +98,70 @@ int qcom_set_boot_addr(uint32_t addr, bool arm64)
 				QCOM_SCM_BOOT_FLAG_COLD_ALL, addr);
 }
 
+void qcom_power_up_l2_cache(uint32_t mpidr, uint32_t base)
+{
+	uint32_t val;
+
+	/* Skip if cluster L2 is already powered on */
+	val = readl(base + APCS_L2_PWR_STATUS);
+	if (val & APCS_L2_PWR_STATUS_L2_HS_STS) {
+		dprintf(INFO, "CPU (%d) L2 cache already powered-up\n", mpidr);
+		return;
+	}
+
+	/* Close L2/SCU Logic GDHS and power up the cache */
+	val = APCS_L2_PWR_CTL_L2_HS_CNT(16) |
+	      APCS_L2_PWR_CTL_L2_ARRAY_HS_CLAMP |
+	      APCS_L2_PWR_CTL_SCU_ARRAY_HS_CLAMP | APCS_L2_PWR_CTL_SYS_RESET |
+	      APCS_L2_PWR_CTL_L2_HS_RST | APCS_L2_PWR_CTL_L2_HS_EN |
+	      APCS_L2_PWR_CTL_L2_HS_CLAMP;
+
+	writel(val, base + APCS_L2_PWR_CTL);
+
+	/* Assert PRESETDBGn */
+	writel(APCS_PWR_CTL_OVR_PRESETDBG, base + APCS_PWR_CTL_OVERRIDE);
+	dsb();
+	udelay(2);
+
+	/* De-assert L2/SCU memory Clamp */
+	val &= ~(APCS_L2_PWR_CTL_L2_ARRAY_HS_CLAMP |
+		 APCS_L2_PWR_CTL_SCU_ARRAY_HS_CLAMP);
+	writel(val, base + APCS_L2_PWR_CTL);
+
+	/* Wakeup L2/SCU RAMs by deasserting sleep signals */
+	val |= (APCS_L2_PWR_CTL_SCU_ARRAY_HS | APCS_L2_PWR_CTL_L2_ARRAY_HS);
+	writel(val, base + APCS_L2_PWR_CTL);
+	dsb();
+	udelay(2);
+
+	/* Enable clocks via SW_CLK_EN */
+	writel(APCS_L2_CORE_CBCR_CLK_ENABLE, base + APCS_L2_CORE_CBCR);
+
+	/* De-assert L2/SCU logic clamp */
+	val &= ~APCS_L2_PWR_CTL_L2_HS_CLAMP;
+	writel(val, base + APCS_L2_PWR_CTL);
+	dsb();
+	udelay(2);
+
+	/* De-assert PRESSETDBg */
+	writel(0x0, base + APCS_PWR_CTL_OVERRIDE);
+
+	/* De-assert L2/SCU Logic reset */
+	val &= ~(APCS_L2_PWR_CTL_SYS_RESET | APCS_L2_PWR_CTL_L2_HS_RST);
+	writel(val, base + APCS_L2_PWR_CTL);
+	dsb();
+	udelay(54);
+
+	/* Turn on the PMIC_APC */
+	val |= APCS_L2_PWR_CTL_PMIC_APC_ON;
+	writel(val, base + APCS_L2_PWR_CTL);
+
+	/* Set H/W clock control for the cpu CBC block */
+	writel(APCS_L2_CORE_CBCR_HW_CTL | APCS_L2_CORE_CBCR_CLK_ENABLE,
+	       base + APCS_L2_CORE_CBCR);
+	dsb();
+}
+
 void qcom_power_up_arm_cortex(uint32_t mpidr, uint32_t base)
 {
 	uint32_t pwr_ctl;
@@ -64,6 +170,8 @@ void qcom_power_up_arm_cortex(uint32_t mpidr, uint32_t base)
 		dprintf(INFO, "Skipping boot of current CPU (%d)\n", mpidr);
 		return;
 	}
+
+	qcom_power_up_l2_cache(mpidr, APCS_GLB_BASE(base));
 
 	pwr_ctl = CPU_PWR_CTL_CLAMP | CPU_PWR_CTL_CORE_MEM_CLAMP |
 		  CPU_PWR_CTL_CORE_RST | CPU_PWR_CTL_COREPOR_RST;
--- a/platform/msm8916/msm8916-clock.c
+++ b/platform/msm8916/msm8916-clock.c
@@ -684,6 +684,41 @@ static struct clk_lookup msm_clocks_8916[] =
 	CLK_LOOKUP("gcc_blsp1_qup5_i2c_apps_clk", gcc_blsp1_qup5_i2c_apps_clk.c),
 };
 
+#define APCS_ALIAS0_CMD_RCGR_BASE		0xb111050
+#define APCS_ALIAS0_CMD_RCGR_REG		0x0
+#define APCS_ALIAS0_CMD_RCGR_UPDATE		BIT(0)
+#define APCS_ALIAS0_CFG_REG			0x4
+#define APCS_ALIAS0_CFG_RCGR_SRC_SEL_CLK3	BIT(10)
+#define APCS_ALIAS0_CFG_RCGR_SRC_DIV_2		(BIT(1) | BIT(0))
+
+static void msm8939_clock_c0_init(void)
+{
+	uint32_t base;
+	int regval;
+	int count;
+
+	base = APCS_ALIAS0_CMD_RCGR_BASE;
+
+	/* Source GPLL0 and 1/2 the rate of GPLL0 0x0403 */
+	regval = APCS_ALIAS0_CFG_RCGR_SRC_SEL_CLK3 |
+		 APCS_ALIAS0_CFG_RCGR_SRC_DIV_2;
+	writel(regval, base + APCS_ALIAS0_CFG_REG);
+	regval = readl(base + APCS_ALIAS0_CFG_REG);
+
+	/* update bit */
+	regval = readl(base + APCS_ALIAS0_CMD_RCGR_REG);
+	regval |= APCS_ALIAS0_CMD_RCGR_UPDATE;
+	writel(regval, base + APCS_ALIAS0_CMD_RCGR_REG);
+
+	/* Wait for update to take effect */
+	for (count = 500; count > 0; count--) {
+		regval = readl(base + APCS_ALIAS0_CMD_RCGR_REG);
+		if (!regval & APCS_ALIAS0_CMD_RCGR_UPDATE)
+			break;
+		udelay(1);
+	}
+}
+
 void msm8939_clock_override()
 {
 	mdss_mdp_clk_src.freq_tbl = ftbl_mdss_mdp_clk_src;
@@ -691,7 +726,9 @@ void msm8939_clock_override()
 
 void platform_clock_init(void)
 {
-	if (platform_is_msm8939() || platform_is_msm8929())
+	if (platform_is_msm8939() || platform_is_msm8929()) {
 		msm8939_clock_override();
+		msm8939_clock_c0_init();
+	}
 	clk_init(msm_clocks_8916, ARRAY_SIZE(msm_clocks_8916));
 }


lk2nd/smp/cpu-boot.c
```diff
@@ -56,6 +61,57 @@ int qcom_set_boot_addr(uint32_t addr, bool arm64)
...
+	/* De-assert PRESSETDBg */
+	writel(0x0, base + L2_PWR_CTL_OVERRIDE);
+
+	/* De-assert L2/SCU Logic reset */
+	writel(0x100203, base + L2_PWR_CTL);
+	dsb();
+	udelay(54);
+
+	/* Turn on the PMIC_APC */
+	writel(0x10100203, base + L2_PWR_CTL);
```
stephan-gh, line 108: Can you rewrite this with proper register definitions, similar to `qcom_power_up_arm_cortex()` below? They are publicly documented in the APQ8016E TRM and/or the HRD. This will also allow you to drop most of the comments actually, because `|= L2_PWR_CTL_PMIC_APC_ON` should be as clear as `Turn on the PMIC_APC`.

lk2nd/smp/spin-table.c
```diff
@@ -81,17 +81,26 @@ static void smp_spin_table_setup_cpu(struct smp_spin_table *table,
...
 	node = lkfdt_lookup_phandle(fdt, cpu_node, "next-level-cache");
 	if (node < 0) {
 		dprintf(CRITICAL, "Cannot find next-level-cache: %d\n", node);
 		return;
 	}
 
+#if TARGET_MSM8916 || TARGET_MSM8226
+
+	/* Power Domain l2 cache for second cluster */
+	val = fdt_getprop(fdt, node, "reg", &len);
```
stephan-gh, line 95: Is there any existing example for using a memory `reg` in the l2 cache controller node? It does sound reasonable to me, but I am not sure if this will be accepted in mainline. Does this pass `make dtbs_check`?

lk2nd/smp/cpu-boot.c
```diff
@@ -56,6 +61,57 @@ int qcom_set_boot_addr(uint32_t addr, bool arm64)
 				QCOM_SCM_BOOT_FLAG_COLD_ALL, addr);
 }
 
+void qcom_power_on_l2_cache(uint32_t mpidr, uint32_t base)
```
TravMurav, line 64: Maybe
```suggestion
void qcom_power_up_l2_cache(uint32_t mpidr, uint32_t base)
```
to be consistent with other functions in this file?

lk2nd/smp/cpu-boot.c
bryanodonoghue: Sure NP, I think I can find a copy

lk2nd/smp/spin-table.c
bryanodonoghue: Well, that's the one open question on this.

I don't think "reg" should be a problem since it is a standard property but, I won't know until the DTS gets a first pass upstream review.

The intention is to submit the DTS with 

- enable-method = "spin-table"
- a reference to lk2nd in the commit log and DTS

so that the means of getting all fours cores up is explicit.

The other possibility is to encode the two necessary addresses for the L2 cache inside of lk2nd but, I wasn't sure how you'd feel about that

We could special case it with a compat for msm8936 and msm8939 and it would require no Linux kernel changes to support .. ?

lk2nd/smp/cpu-boot.c
bryanodonoghue: Sure np, makes sense to me

lk2nd/smp/spin-table.c
bryanodonoghue: If we are open to encoding the address inside of lk2nd somehow, that's one less bit of churn to be submitted and brought through the Linux upstreaming process, might be a good bit easier to do

lk2nd/smp/cpu-boot.c
stephan-gh: There is no need to search for a copy, you can easily download it from the Qualcomm web page. :)

- TRM: https://developer.qualcomm.com/download/sd410/snapdragon-410e-technical-reference-manual.pdf, page 97
- HRD: https://developer.qualcomm.com/download/sd410/hardware-register-description-qualcomm-snapdragon-410.pdf

lk2nd/smp/spin-table.c
bryanodonoghue: actually let me check what dtbs_check makes of this..

stephan-gh: I have mixed feelings about this so I can probably be convinced for many different kinds of options. :)

The main reason for taking the addresses from the DT was really just simplicity. I already had the code to parse the CPUs in the DT to set the `cpu-release-addr`. Also, the `qcom,acc` and `qcom,saw` properties were already present in `msm8916.dtsi` upstream because I added support for some devices with (signed) 32-bit-only firmware there.

The difference in your case is that there are no existing bindings for describing the L2 cache memory region. Also, it seems unlikely that anyone will ever want to boot ARM32 Linux on MSM8939 (I think all MSM8939 devices were shipped with a 64-bit TZ firmware). In other words, you would only add the address upstream for lk2nd to use, while there would be no Linux code making use of the additional properties.

I think this is a bit hard to justify, especially considering there is already plenty of Qualcomm code with hardcoded addresses in lk2nd. So, I'm inclined to agree with you that it is not worth the hassle upstream and we should just encode the address in lk2nd for MSM8939.

You could just cleverly derive the base address for the L2 cache from the one given to `qcom_power_up_arm_cortex()` :D

```diff
diff --git a/lk2nd/smp/cpu-boot.c b/lk2nd/smp/cpu-boot.c
index 82b7ffc08..5668dab23 100644
--- a/lk2nd/smp/cpu-boot.c
+++ b/lk2nd/smp/cpu-boot.c
@@ -65,6 +65,9 @@ void qcom_power_up_arm_cortex(uint32_t mpidr, uint32_t base)
 		return;
 	}
 
+	/* Make sure the L2 cache is powered up */
+	qcom_power_up_l2_cache((base & 0xfff00000) + 0x11000);
+
 	pwr_ctl = CPU_PWR_CTL_CLAMP | CPU_PWR_CTL_CORE_MEM_CLAMP |
 		  CPU_PWR_CTL_CORE_RST | CPU_PWR_CTL_COREPOR_RST;
 	writel(pwr_ctl, base + CPU_PWR_CTL);

```

wonderfulShrineMaidenOfParadise:
```
[    0.065705] smp: Bringing up secondary CPUs ...
[    0.067899] Detected VIPT I-cache on CPU1
[    0.068090] CPU1: Booted secondary processor 0x0000000101 [0x410fd031]
[    0.069297] Detected VIPT I-cache on CPU2
[    0.069418] CPU2: Booted secondary processor 0x0000000102 [0x410fd031]
[    0.070437] Detected VIPT I-cache on CPU3
[    0.070523] CPU3: Booted secondary processor 0x0000000103 [0x410fd031]
[    0.072238] Detected VIPT I-cache on CPU4
[    0.073776] CPU4: Booted secondary processor 0x0000000000 [0x410fd031]
[    0.078609] Detected VIPT I-cache on CPU5
[    0.080080] CPU5: Booted secondary processor 0x0000000001 [0x410fd031]
[    0.084813] Detected VIPT I-cache on CPU6
[    0.086278] CPU6: Booted secondary processor 0x0000000002 [0x410fd031]
[    0.090942] Detected VIPT I-cache on CPU7
[    0.092414] CPU7: Booted secondary processor 0x0000000003 [0x410fd031]
[    0.095695] smp: Brought up 1 node, 8 CPUs
[    0.168200] SMP: Total of 8 processors activated.
[    0.172276] CPU features: detected: 32-bit EL0 Support
[    0.177076] CPU features: detected: 32-bit EL1 Support
[    0.182079] CPU features: detected: CRC32 instructions
[    0.237318] CPU: All CPU(s) started at EL1
[    0.237875] alternatives: patching kernel code
[    0.250308] devtmpfs: initialized
[    0.272402] KASLR disabled due to lack of seed
[    0.272988] clocksource: jiffies: mask: 0xffffffff max_cycles: 0xffffffff, max_idle_ns: 7645041785100000 ns
[    0.275863] futex hash table entries: 2048 (order: 5, 131072 bytes, linear)
[    0.289464] pinctrl core: initialized pinctrl subsystem
[    0.296331] NET: Registered PF_NETLINK/PF_ROUTE protocol family
[    0.311612] DMA: preallocated 256 KiB GFP_KERNEL pool for atomic allocations
[    0.318079] DMA: preallocated 256 KiB GFP_KERNEL|GFP_DMA pool for atomic allocations
[    0.319462] DMA: preallocated 256 KiB GFP_KERNEL|GFP_DMA32 pool for atomic allocations
[    0.325832] audit: initializing netlink subsys (disabled)
[    0.334913] audit: type=2000 audit(0.272:1): state=initialized audit_enabled=0 res=1
[    0.335554] thermal_sys: Registered thermal governor 'step_wise'
[    0.339995] cpuidle: using governor menu
[    0.352894] NET: Registered PF_QIPCRTR protocol family
[    0.356860] hw-breakpoint: found 6 breakpoint and 4 watchpoint registers.
[    0.364026] ASID allocator initialised with 65536 entries
[    0.428502] HugeTLB registered 1.00 GiB page size, pre-allocated 0 pages
[    0.428628] HugeTLB registered 32.0 MiB page size, pre-allocated 0 pages
[    0.434381] HugeTLB registered 2.00 MiB page size, pre-allocated 0 pages
[    0.441106] HugeTLB registered 64.0 KiB page size, pre-allocated 0 pages
[    0.455830] cryptd: max_cpu_qlen set to 1000
[    0.469027] iommu: Default domain type: Translated 
[    0.469111] iommu: DMA domain TLB invalidation policy: strict mode 
[    0.473448] SCSI subsystem initialized
[    0.479258] usbcore: registered new interface driver usbfs
[    0.482889] usbcore: registered new interface driver hub
[    0.488412] usbcore: registered new device driver usb
[    0.495369] i2c-gpio i2c-fg: using lines 496 (SDA) and 495 (SCL)
[    0.499683] i2c-gpio i2c-nfc: using lines 390 (SDA) and 391 (SCL)
[    0.505801] i2c-gpio i2c-sensor: using lines 474 (SDA) and 475 (SCL)
[    0.511708] i2c-gpio i2c-tkey: using lines 406 (SDA) and 407 (SCL)
[    0.517873] pps_core: LinuxPPS API ver. 1 registered
[    0.523470] pps_core: Software ver. 5.3.6 - Copyright 2005-2007 Rodolfo Giometti <giometti@linux.it>
[    0.528384] PTP clock support registered
[    0.538975] qcom_scm: convention: smc arm 32
[    0.544008] Advanced Linux Sound Architecture Driver Initialized.
[    0.550526] clocksource: Switched to clocksource arch_sys_counter
[    0.552064] VFS: Disk quotas dquot_6.6.0
[    0.557884] VFS: Dquot-cache hash table entries: 512 (order 0, 4096 bytes)
[    0.562897] AppArmor: AppArmor Filesystem Enabled
[    0.672623] NET: Registered PF_INET protocol family
[    0.673467] IP idents hash table entries: 32768 (order: 6, 262144 bytes, linear)
[    0.678626] tcp_listen_portaddr_hash hash table entries: 1024 (order: 2, 16384 bytes, linear)
[    0.684257] TCP established hash table entries: 16384 (order: 5, 131072 bytes, linear)
[    0.692760] TCP bind hash table entries: 16384 (order: 6, 262144 bytes, linear)
[    0.700958] TCP: Hash tables configured (established 16384 bind 16384)
[    0.708107] UDP hash table entries: 1024 (order: 3, 32768 bytes, linear)
[    0.714125] UDP-Lite hash table entries: 1024 (order: 3, 32768 bytes, linear)
[    0.721450] NET: Registered PF_UNIX/PF_LOCAL protocol family
[    0.737701] Unpacking initramfs...
[    0.752412] hw perfevents: enabled with armv8_cortex_a53 PMU driver, 7 counters available
[    0.772856] kvm [1]: HYP mode not available
[    0.910247] Freeing initrd memory: 2544K
[    0.925377] Initialise system trusted keyrings
[    0.931691] workingset: timestamp_bits=46 max_order=19 bucket_order=0
[    3.680225] Key type asymmetric registered
[    3.680804] Asymmetric key parser 'x509' registered
[    3.686900] Block layer SCSI generic (bsg) driver version 0.4 loaded (major 247)
[    3.692556] io scheduler mq-deadline registered
[    3.697111] io scheduler kyber registered
```

Good to see another cluster brought up. But the booting stops here.

bryanodonoghue:
@wonderfulShrineMaidenOfParadise 
Try this base SoC dts
https://git.linaro.org/people/bryan.odonoghue/kernel.git/tree/arch/arm64/boot/dts/qcom/msm8939.dtsi?h=linux-stable-apq8039-dts-5.17-snoc-v2

Not sure what you have for
- Regulators
- Operating points
in that 5.10 tree you are working with

lk2nd/smp/spin-table.c
bryanodonoghue: `qcom_power_up_l2_cache((base & 0xfff00000) + 0x11000);`

That's nice, I'll steal that code for V2 if you don't mind

wonderfulShrineMaidenOfParadise:
I patched [msm8939.dtsi](https://gitlab.com/msm8939-mainline/linux/-/blob/v5.15-8916+8939/arch/arm64/boot/dts/qcom/msm8939.dtsi). Not sure how to convert the dts to be compatible with the tree.

```c
(line 270)

 		L2_0: l2-cache { /* BIG */
 			compatible = "cache";
 			cache-level = <2>;
+			reg = <0x0b111000 0x1000>;
 		};
 
 		L2_1: l2-cache-1 { /* little */
 			compatible = "cache";
 			cache-level = <2>;
+			reg = <0x0b011000 0x1000>;
 		};
```

bryanodonoghue:
@wonderfulShrineMaidenOfParadise 
Yep, that will go away with the changes myself and @stephan-gh discussed, we'll just offset from the CPU base + 0x11000

lk2nd/smp/spin-table.c
stephan-gh: Sure, maybe add a `#define` for `0x11000` like `APCS_GLB_OFFSET`. Just so it doesn't look like complete magic :)

bryanodonoghue:
OK new series just the one commit now

- Dropped the DTS binding
- Implemented the offset to the cache regs we discussed
- Named all of the bits instead of using magic writes - two undocumented bits found
- Verified that the values being written are a match for the original qcom magic write values

Verified cluster bootup as before on msm8939

lk2nd/smp/cpu-boot.c
```diff
@@ -23,6 +23,80 @@
...
+#define APCS_PWR_CTL_OVR_CLK_EN_PISLEEP_CLK	(BIT(0) | BIT(1))
+#define APCS_PWR_CTL_OVR_CLK_EN_REF_CLK		BIT(1)
+#define APCS_PWR_CTL_OVR_CLK_EN_AHB_CLK		BIT(0)
+#define APCS_PWR_CTL_OVR_CLK_EN_QSB_CLK		0
+
+#define APCS_L2_PWR_CTL_ADDR			0x14
+#define APCS_PMIC_APC_O				BIT(28)
+#define APCS_L2_HS_CNT(cnt)			((cnt) << 16)
+#define APCS_L2_SCU_MAGIC_BIT1			BIT(15)
+#define APCS_L2_SCU_MAGIC_BIT0			BIT(14)
```
stephan-gh, line 67: Hmm these bits are documented in the TRM on page 121?

```c
#define APCS_L2_PWR_CTL_L2_ARRAY_HS_CLAMP	BIT(15)
#define APCS_L2_PWR_CTL_SCU_ARRAY_HS_CLAMP	BIT(14)
```

Ah, they are missing in the HRD. That's quite typical for Qualcomm, you always need to check both documents, some things are more complete in TRM and others more complete in the HRD... :)

lk2nd/smp/cpu-boot.c
```diff
@@ -23,6 +23,80 @@
...
+#define APCS_L2_PWR_L2_RET_SLP			BIT(13)
+#define APCS_L2_PWR_CTL_SYS_RESET		BIT(12)
+#define APCS_L2_PWR_CTL_L2_SLEEP_STATE		BIT(11)
+#define APCS_L2_PWR_CTL_L2_HS_RST		BIT(10)
+#define APCS_L2_PWR_CTL_L2_HS_EN		BIT(9)
+#define APCS_L2_PWR_CTL_L2_HS_CLAMP		BIT(8)
+#define APCS_L2_PWR_CTL_L2_ARRAY_HS_3		BIT(3)
+#define APCS_L2_PWR_CTL_L2_ARRAY_HS_2		BIT(2)
+#define APCS_L2_PWR_CTL_L2_ARRAY_HS_1		BIT(1)
+#define APCS_L2_PWR_CTL_L2_ARRAY_HS_0		BIT(0)
```
stephan-gh, line 77: Hmm, great the TRM and HRD seems to have some disagreement about the naming here. But it looks like the TRM is more correct, so this should be:

```c
#define APCS_L2_PWR_CTL_L2_RST_DIS		BIT(2)
#define APCS_L2_PWR_CTL_SCU_ARRAY_HS		BIT(1)
#define APCS_L2_PWR_CTL_L2_ARRAY_HS		BIT(0)
```

and bit 3 likely doesn't exist in the hardware...

If you have contacts at Qualcomm, feel free to thank them for creating such confusing documentation. :) (Although I'm honestly really happy Qualcomm decided to have any public documentation at all for APQ8016E...)

stephan-gh (changes requested):
Thanks for clearing up the code. It looks good to me mostly, just some minor nits. :)

lk2nd/smp/cpu-boot.c
```diff
@@ -23,6 +23,80 @@
 #define APC_PWR_GATE_CTL_GHDS_EN	BIT(0)
 #define APC_PWR_GATE_CTL_GHDS_CNT(cnt)	((cnt) << 24)
 
+/*
+ * Each CPU within a cluster has a qcom,kpss-acc-v2.
+ * There is just once L2 cache controller per cluster.
+ *
+ * APCS_KPSS_GLB_BASE takes the base of given qcom,kpss-acc-v2 and offsets
+ * back the single APCS_KPSS_GLB register which is the base for the L2
+ * cache controller in the cluster.
```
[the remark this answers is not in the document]
bryanodonoghue: Referencing Documentation/devicetree/bindings/arm/msm/qcom,kpss-acc.txt::"qcom,kpss-acc-v2"
Not to worry I'll drop the KPSS here

lk2nd/smp/cpu-boot.c
```diff
@@ -23,6 +23,80 @@
...
+ *       qcom,kpss-acc-v2@0b1a8000
+ *       qcom,kpss-acc-v2@0b1b8000
+ *       => APCS_KPSS_GLB_BASE@0b110000
+ */
+#define APCS_KPSS_GLB_MASK			0xfff00000
+#define APCS_KPSS_GLB_OFFSET			0x00011000
+#define APCS_KPSS_GLB_BASE(base)		(((base) & APCS_KPSS_GLB_MASK) + \
+						  APCS_KPSS_GLB_OFFSET)
+
+#define APCS_PWR_CTL_OVERRIDE_ADDR		0xc
```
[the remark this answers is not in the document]
bryanodonoghue: Ah I saw
#define QCOM_SCM_BOOT_SET_ADDR          0x01
NP I'll drop the _ADDR in the names

lk2nd/smp/cpu-boot.c
```diff
@@ -23,6 +23,80 @@
...
+#define APCS_PWR_CTL_OVR_PRESETDBG		BIT(22)
```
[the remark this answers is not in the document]
bryanodonoghue: OK, I drop unused bits from all defintions

lk2nd/smp/cpu-boot.c
```diff
@@ -56,6 +130,79 @@ int qcom_set_boot_addr(uint32_t addr, bool arm64)
 				QCOM_SCM_BOOT_FLAG_COLD_ALL, addr);
 }
 
+void qcom_power_up_l2_cache(uint32_t mpidr, uint32_t base)
+{
+	uint32_t val;
+
+	/* Skip if cluster L2 is already powered on */
+	val = readl(base + APCS_L2_PWR_STATUS_ADDR);
+	if (val & BIT(9)) {
```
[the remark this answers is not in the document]
bryanodonoghue: doh - I made a mental note to change that before submitting
Have the blinkers on

lk2nd/smp/cpu-boot.c
```diff
@@ -56,6 +130,79 @@ int qcom_set_boot_addr(uint32_t addr, bool arm64)
...
+	 * 100001101011100000000 including undocumented BIT(15) | BIT(14)
+	 * 20 10000	APCS_L2_HS_CNT(CNT) => 0x10
+	 * 15 1		APCS_L2_SCU_MAGIC_BIT1
+	 * 14 10	APCS_L2_SCU_MAGIC_BIT0
+	 * 12 10	APCS_L2_PWR_CTL_SYS_RESET
+	 * 10 1		APCS_L2_PWR_CTL_L2_HS_RST
+	 * 9  1		APCS_L2_PWR_CTL_L2_HS_EN
+	 * 8  1		APCS_L2_PWR_CTL_L2_HS_CLAMP
+	 */
+	val = APCS_L2_HS_CNT(0x10) | APCS_L2_SCU_MAGIC_BIT1 |
```
stephan-gh, line 157: Another very minor nit: I usually count using decimals, not hex. This is a bit like writing `udelay(0x36)` instead of `udelay(54)` otherwise. ;)

```suggestion
	val = APCS_L2_HS_CNT(16) | APCS_L2_SCU_MAGIC_BIT1 |
```

lk2nd/smp/cpu-boot.c
bryanodonoghue: Ah missed that entirely, I'll check the TRM so

lk2nd/smp/cpu-boot.c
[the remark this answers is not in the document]
stephan-gh: Ah, that one is not a register address though, it's the SCM function ID for the "set boot addr(ess)" function call :)

lk2nd/smp/cpu-boot.c
bryanodonoghue: Sure decimal it is

bryanodonoghue: I find it a bit easier to mentally calculate shifts in hex ..

stephan-gh: Right, from that perspective hex would be better.

My thinking comes from an (idealized) standpoint where we would actually understand the hardware well enough to write such a sequence on our own (rather than just decoding the sequence from Qualcomm). In that case it does not matter if it ends up being 0x10 shifted, 0x42 or whatever. We want to set the "count value" to 16 and the compiler can worry about assembling the register value. :)

lk2nd/smp/cpu-boot.c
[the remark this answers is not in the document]
bryanodonoghue: I scrubbed just the APCS_PWR_CTL_OVERRIDE bits, the code doesn't look to long to me with the other registers, which we use more bits from fully defined.

Let me know if you want all unused bits zapped though, I can do a V4.

bryanodonoghue:
OK, I think I've captured most/all of your feedback there Stephan

lk2nd/smp/cpu-boot.c
```diff
@@ -23,6 +23,61 @@
...
+#define APCS_L2_PWR_STATUS_L2_SLV_IDLE		BIT(24)
+#define APCS_L2_PWR_STATUS_L2_HW_EVENT		(BIT(20) | BIT(19) | BIT(18) | \
+						 BIT(17) | BIT(16))
+#define APCS_L2_PWR_STATUS_DBG_RST_STS		BIT(13)
+#define APCS_L2_PWR_STATUS_L2_HS_RST_STS	BIT(12)
+#define APCS_L2_PWR_STATUS_L2_RET_SLP_STS	BIT(10)
+#define APCS_L2_PWR_STATUS_L2_HS_STS		BIT(9)
+#define APCS_L2_PWR_STATUS_L2_CLMP_STS		BIT(8)
+#define APCS_L2_PWR_STATUS_L2_ARRAY_HS_STS	(BIT(3) | BIT(2) | BIT(1) | \
+						 BIT(0))
```
stephan-gh, line 74: I'd drop unused bits for APCS_L2_PWR_STATUS as well, since we're only using APCS_L2_PWR_STATUS_L2_HS_STS

stephan-gh (approved):
Looks good to me, just this minor comment that can also be done when merging if you don't want to push again. :)

lk2nd/smp/cpu-boot.c
bryanodonoghue: NP I can update - we are waiting on @TravMurav to ack anyway

stephan-gh (approved):
This looks good to me and I also verified that the compiler produces the correct register values with this. Thanks! :+1: 

Before merging this I have asked the MSM8939 people at postmarketOS to verify that this works correctly for them. Note that it might take a bit since I assume everyone is a bit busy this weekend because of FOSDEM. There is no need to rush this in, right?

lk2nd/smp/cpu-boot.c
```diff
@@ -56,6 +98,70 @@ int qcom_set_boot_addr(uint32_t addr, bool arm64)
...
+		dprintf(INFO, "CPU (%d) L2 cache already powered-up\n", mpidr);
+		return;
+	}
+
+	/* Close L2/SCU Logic GDHS and power up the cache */
+	val = APCS_L2_PWR_CTL_L2_HS_CNT(16) |
+	      APCS_L2_PWR_CTL_L2_ARRAY_HS_CLAMP |
+	      APCS_L2_PWR_CTL_SCU_ARRAY_HS_CLAMP | APCS_L2_PWR_CTL_SYS_RESET |
+	      APCS_L2_PWR_CTL_L2_HS_RST | APCS_L2_PWR_CTL_L2_HS_EN |
+	      APCS_L2_PWR_CTL_L2_HS_CLAMP;
```
TravMurav, line 117: Smallest nit: IMO it's a bit easier to read bits here if they're all aligned on the new line, so the `APCS_L2_PWR_CTL_` prefix is in the same place:

```suggestion
	/* Close L2/SCU Logic GDHS and power up the cache */
	val = APCS_L2_PWR_CTL_L2_HS_CNT(16) |
	      APCS_L2_PWR_CTL_L2_ARRAY_HS_CLAMP |
	      APCS_L2_PWR_CTL_SCU_ARRAY_HS_CLAMP |
	      APCS_L2_PWR_CTL_SYS_RESET |
	      APCS_L2_PWR_CTL_L2_HS_RST |
	      APCS_L2_PWR_CTL_L2_HS_EN |
	      APCS_L2_PWR_CTL_L2_HS_CLAMP;
```

TravMurav (dismissed):
Looks fine to me as well!

Also want to hear from the msm8939 people as we're changing their bootflow a bit...

bryanodonoghue:
The only impact will be that the second cluster will have to be brought up to a higher performance level, which is a kernel/DTS task.

That additional work isn't present in the DTS I shared above but is in a backlog of patches we have in a Linaro tree to apply to 8939.

stephan-gh:
> The only impact will be that the second cluster will have to be brought up to a higher performance level, which is a kernel/DTS task.

Which "performance level" are you referring to here? CPU frequency? CCI frequency? Some other interconnect?  
If e.g. the CPU frequency is way too low I wonder if we should set it up differently in lk2nd already.

> That additional work isn't present in the DTS I shared above but is in a backlog of patches we have in a Linaro tree to apply to 8939.

OK, I think we should keep this PR open then until the second cluster is fully working. It would be unfortunate if merging this patch actually makes the devices perform worse, not better. :)

bryanodonoghue:
> 

I mean CPR. The second cluster will operate at nominal frequency until CPR is applied.
But it is still useful even at the nominal frequency @ ~ 800MHz

It simply won't ramp up to the ~1.5GHz until we do CPR in msm8939 upstream

So what I'm proposing here is

- Merging this, so that cluster booting is a non-issue on msm8939
- Bringing an initial msm8939 to upstream
- Adding in CPR on top of that

CPR is ready to go but, I think doing both 8939 basic upstream and CPR at the same time is *alot*

bryanodonoghue:
> > 
> 
> I mean CPR. The second cluster will operate at nominal frequency until CPR is applied. But it is still useful even at the nominal frequency @ ~ 800MHz

Is that really a blocker for this PR ?

stephan-gh:
No! Let's not block this PR with CPR please. :) At least on 8916 there are still several open blockers to make CPR work cleanly upstream, so 800 MHz sounds fine for now.

However, it looks like the second cluster is actually starting at 19.2 MHz (XO) with these changes? This is what a `clk_summary` revealed (from @vknecht):

```
 xo-board                             8        8        0    19200000          0     0  50000         Y
    a53mux@b111000                    1        1        0    19200000          0     0  50000         Y
```

Should we configure `APCS_CFG_RCGR` appropriately in the l2 power up sequence, so it actually takes the clock from gpll0 and ends up at 800 MHz instead of 19.2 MHz? I'm not sure if it's a good idea to let the CPU boot up at 19.2 MHz...

I'm expecting something like https://github.com/shawnguo2/linux/commit/4cc21165e482e1f61e8d0ab92d4105725d19a1a1, although some parts of that commit are confusing me:

1. Why does it configure the a53-pll if the CPU clock is taken from gpll0?
2. It configures 400 MHz (gpll0 with divider 2) for some reason rather than 800 MHz.

Also, what about the CCI clock? Does it need to be configured? Somehow it shows up with rate 0 in @vknecht's `clk_summary`:

```
 a53pll@b1d0000                       1        1        0           0          0     0  50000         ?
    a53mux@b1d1000                    1        1        0           0          0     0  50000         Y
```

I'm still a bit confused about how the second cluster should be brought up properly. :)

bryanodonoghue:
Agreed CPR is out for lk2nd

I believe the default clocks - without CPR are
     C0:  800 MHz
     C1:  1113 MHz
     CCI: 297 MHz

But that said it doesn't make much sense to me that a54mux is @ 19.2 CXO 

I can't give quick answers to any of these questions but, agreed lets look at the cpu_ops gpll configuration since

- That afaik is not upstream
- Might be difficult to get upstream

so importing the logic to lk2nd is the right thing to do

stephan-gh:
> But that said it doesn't make much sense to me that a53mux is @ 19.2 CXO

I think this is simply the reset value for `APCS_CFG_RCGR`. This patch is related: https://gitlab.com/msm8939-mainline/linux/-/commit/4295b4303e22e85e9f5bd0966232bd49cbd07be2

bryanodonoghue:
Hi @stephan-gh @TravMurav 

So I looked into the default clock for the little cluster and indeed it was not starting.
I imported the downstream implementation and per our previous reviews included only the used bits in registers
I have the msm8916 documentation but for C0_PLL_CONFIG_CTL_REG there are some bits I haven't identified.

Booting to Linux we find the two cluster clocks look like this

before:
`a53mux@b111000              1        1        0           0          0     0  50000         Y
a53mux@b011000              1        1        0   800000000          0     0  50000         Y`

after:
`a53mux@b111000              1        1        0   400000000          0     0  50000         Y
a53mux@b011000              1        1        0   800000000          0     0  50000         Y`

Looking at the performance levels in the msm8936/msm8939 clock plan we see
Level0 == 400MHz
Level1 == 800MHz

Bootup is nice and fast and I've verified on two separate msm8939 devices.

platform/msm8916/msm8916-clock.c
```diff
@@ -684,14 +684,150 @@ static struct clk_lookup msm_clocks_8916[] =
...
+	uint32_t base = APCS_C0_PLL_BASE;
+	int regval;
+
+	/* Disable Mode */
+	writel(0, base + C0_PLL_MODE_REG);
+
+	/* Configure L/M/N values - these are accumulator fields */
+	writel(52, base + C0_PLL_L_VAL_REG);
+	writel(0,  base + C0_PLL_M_VAL_REG);
+	writel(1,  base + C0_PLL_N_VAL_REG);
```
TravMurav, line 736: I'd prefer that configurations like this to have a comment describing values in a more "human friendly" way:

(as per apq8016 TRM, page 2572)
```
/* 19.2 * (52 + 0/1) = 998.4 MHz */
```

platform/msm8916/msm8916-clock.c
```diff
@@ -684,14 +684,150 @@ static struct clk_lookup msm_clocks_8916[] =
...
+	uint32_t base;
+	int regval;
+	int count;
+
+	msm8939_clock_c0_enable_sr2_pll();
+
+	base = APCS_ALIAS0_CMD_RCGR_BASE;
+
+	/* Source GPLL0 and 1/2 the rate of GPLL0 0x0403 */
+	regval = APCS_ALIAS0_CFG_RCGR_SRC_SEL_CLK3 |
```
TravMurav, line 796: Hm, here the clock is set to be sourced from GPLL0 and it gets to `800 * 1/2 = 400 MHz`, I'm unsure if the sr2 pll is needed in this case?

TravMurav (commented):
In fact we discussed switching the cluster to pll0 in lk2nd.
For a bit of context: we had a little discussion on which rate to set and I was concerned that setting the cluster to 800 on boot is a bit too much in case no one picks up the rate control so I prefer the 400 MHz that you set there. 

I'm still unsure why you enable the sr2 pll here, Is it needed because the kernel would like to switch to it as a higher performance level after boot and can't start it on it's own?

platform/msm8916/msm8916-clock.c
stephan-gh: Exactly. It might make sense to configure `C0_PLL_USER_CTL_REG` and `C0_PLL_CONFIG_CTL_REG` in lk2nd (just to avoid having to do this upstream). But I don't see why we should set L/M/N or enable the PLL.

I suspect this is just a strange "artifact" from downstream that is actually not necessary.

platform/msm8916/msm8916-clock.c
```diff
@@ -684,14 +684,150 @@ static struct clk_lookup msm_clocks_8916[] =
...
+		if (!regval & APCS_ALIAS0_CMD_RCGR_UPDATE)
+			break;
+		udelay(1);
+	}
+
+	/* Enable the branch */
+	regval =  readl(base + APCS_ALIAS0_CORE_CBCR_REG);
+	regval |= APCS_ALIAS0_CORE_CBCR_SW_CTL;
+	writel(regval, base + APCS_ALIAS0_CORE_CBCR_REG);
+	regval = readl(base + APCS_ALIAS0_CORE_CBCR_REG);
```
stephan-gh, line 818: Exactly the same register is already written during the L2 power up sequence:

```c
	/* Enable clocks via SW_CLK_EN */
	writel(APCS_L2_CORE_CBCR_CLK_ENABLE, base + APCS_L2_CORE_CBCR);
```

I don't think this should be here. :)

stephan-gh: I think it's fine to have the configuration here though, set the frequency early here but only enable the clock branch during the l2 startup sequence.

platform/msm8916/msm8916-clock.c
bryanodonoghue: My answer is "that's what downstream does" but 

doing this

`+       //msm8939_clock_c0_enable_sr2_pll();`

still results in this

`a53mux@b111000              1        1        0   400000000          0     0  50000         Y`
`a53mux@b011000              1        1        0   800000000          0     0  50000         Y`

so I guess you are both right that this is dead code for starting the cluster ...

I assume it does something but, I'm happy enough to drop it since its not apparently required for this.

platform/msm8916/msm8916-clock.c
bryanodonoghue: I've dropped the SR2 config here, we can pick it up in Linux if required.

stephan-gh:
@bryanodonoghue Thanks for the quick cleanup! Do I understand it correctly that you are going to do the

```c
	/* Configure USER_CTL and CONFIG_CTL value */
	writel_relaxed(0x0100000f, base + C0_PLL_USER_CTL);
	writel_relaxed(0x4c015765, base + C0_PLL_CONFIG_CTL);
```

in Linux? I think there is similar existing code at least for the `C0_PLL_USER_CTL` (see `clk_pll_configure_sr_hpm_lp()` that you use e.g. in gcc-msm8939). I'm not sure about the `C0_PLL_CONFIG_CTL`, but it's probably best to set it like downstream for best stability.

I think logically these two writes would fit better in Linux, but if you don't want to go through the discussion/work about how to solve this exactly upstream, having these two writes in lk2nd would be also fine for me. (Just don't _enable_ the PLL because that sounds unnecessary.) :)

stephan-gh (approved):
OK, I guess we can still add more writes later in case you need them when enabling cpufreq/pll support for the c0 cluster upstream. :)
Changes look good to me now, I'll ask the 8939 people at postmarketOS again to test this.

TravMurav (approved):

wonderfulShrineMaidenOfParadise:
Confirmed booting on [msm8939-mainline/linux/v5.15-8916+8939-v2](https://gitlab.com/msm8939-mainline/linux/-/commits/v5.15-8916+8939-v2/).
>System Benchmarks Index Score (Partial Only)                         1116.8

It was 1022.4 with 4 cores.

bryanodonoghue:
@wonderfulShrineMaidenOfParadise 
Yep we'd expect something like that I'm also sure once we get cpufreq, opps, cci and bimc up, we will get much better performance. 

@TravMurav @stephan-gh one of you will have to hit the merge button, I don't have that option, so it will need to be one of you

PR #143, opened 2022-02-08
dts: Rename SM-G530Y from gprimeltezt to fortunaltezt

wonderfulShrineMaidenOfParadise:
v2: Rename to fortunaltezt ~~and add panel selection~~

### DIFF
--- a/dts/msm8916/msm8916-samsung-r11.dts
+++ b/dts/msm8916/msm8916-samsung-r11.dts
@@ -12,12 +12,12 @@
 
 	/*
 	 * Before building for G530Y, please comment out all dtbs except
-	 *  "$(LOCAL_DIR)/msm8916-samsung-r11.dtb" at './rules.mk'.
-	 * gprimeltezt doesn't work with multi-dtbs; only goes into "Download mode".
+	 * "$(LOCAL_DIR)/msm8916-samsung-r11.dtb" at './rules.mk'.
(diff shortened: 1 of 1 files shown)


dts/msm8916/msm8916-samsung-r11.dts
```diff
@@ -20,6 +20,14 @@
 		compatible = "samsung,gprimeltezt", "qcom,msm8916", "lk2nd,device";
 		lk2nd,match-bootloader = "G530Y*";
 
+		panel {
```
TravMurav, line 23: Do you really need it for a single entry here?

wonderfulShrineMaidenOfParadise: Some of the Grand Prime e.g. SM-G530FZ/H/W variants have 2 panel entries but this one has only one.

wonderfulShrineMaidenOfParadise:
`gprimeltezt` stands for SM-G530MU in fact. Changing it to `fortunaltezt` to avoid confusions.

stephan-gh (approved):

stephan-gh:
The commit message is very weird, we don't do v2 etc in lk2nd. Please just make the commit message like "dts: Rename SM-G530Y to fortunaltezt` or something like that.

TravMurav (approved):
while this rename is just a droplet into the mess that is "properly give compatibles to all devices in lk2nd for the future dtb selection", it's probably fine to move things a bit :)

PR #145, opened 2022-02-12
dts: huawei-kiwi: Use panel selection

vknecht:


dts/msm8916/msm8939-huawei-kiwi.dts
```diff
@@ -16,4 +16,24 @@
 
 	model = "Huawei Honor 5X";
 	compatible = "huawei,kiwi", "qcom,msm8939", "lk2nd,device";
+
+	panel {
+		compatible = "huawei,kiwi-panel";
+
+		qcom,mdss_dsi_auo_otm1901a_5p5_1080pxa_video {
+			compatible = "huawei,auo-otm1901a-5p5xa";
```
stephan-gh, line 24: Does the `5p5xa` really have any meaningful value here? I think it's unique without that suffix for now, although perhaps we should consider also pre or suffixing these with `kiwi` (not sure yet).

vknecht: afaik `5p5xa` comes from the generator, so no strong feeling about it.
Maybe `5p5` is related to an IC, I guess like in `drivers/gpu/drm/panel/panel-asus-z00t-tm5p5-n35596.c`.
Whatever you prefer :-)

stephan-gh: Yep, I think I'd prefer it without the 5p5xa. The compatibles from the generator should always have a FIXME comment at the end, encouraging you to change them if they are stupid.  :)

stephan-gh (approved):

PR #147, opened 2022-02-15
target: msm8226: Enable sdcard fs support

TravMurav:
Rework the msm8226 mmc init to also bring up the sdcard (in the same way as it is done on msm8916)

### DIFF
--- a/target/msm8226/init.c
+++ b/target/msm8226/init.c
@@ -59,7 +59,7 @@
 #endif
 
 extern  bool target_use_signed_kernel(void);
-static void set_sdc_power_ctrl(void);
+static void set_sdc_power_ctrl(uint8_t slot);
 
 #define PMIC_ARB_CHANNEL_NUM               0
 #define PMIC_ARB_OWNER_ID                  0
@@ -103,6 +103,8 @@ static uint32_t mmc_sdc_pwrctl_irq[] =
 	{ SDCC1_PWRCTL_IRQ, SDCC2_PWRCTL_IRQ, SDCC3_PWRCTL_IRQ };
 
 struct mmc_device *dev;
+struct mmc_device *emmc_dev;
+struct mmc_device *sdcard_dev;
 
 void target_load_ssd_keystore(void)
 {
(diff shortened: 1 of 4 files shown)


target/msm8226/init.c
```diff
@@ -226,35 +228,56 @@ void target_crypto_init_params()
...
 	config.max_clk_rate = MMC_CLK_200MHZ;
-
-	/* Trying Slot 1*/
-	config.slot = 1;
-	config.sdhc_base = mmc_sdhci_base[config.slot - 1];
-	config.pwrctl_base = mmc_pwrctl_base[config.slot - 1];
-	config.pwr_irq     = mmc_sdc_pwrctl_irq[config.slot - 1];
+	config.sdhc_base    = mmc_sdhci_base[config.slot - 1];
+	config.pwrctl_base  = mmc_pwrctl_base[config.slot - 1];
+	config.pwr_irq      = mmc_sdc_pwrctl_irq[config.slot - 1];
```
stephan-gh, line 241: Would prefer to avoid the "code churn" here (just keep it badly aligned since this is qcom code), a minimal diff will look easier to understand

target/msm8226/init.c
```diff
@@ -226,35 +228,56 @@ void target_crypto_init_params()
 
 void target_sdc_init()
 {
-	struct mmc_config_data config = {0};
+	struct mmc_config_data config;
+	struct mmc_config_data sd_config;
```
stephan-gh, line 232: Is there a reason why we cannot re-use the config here like the old code did?

target/msm8226/init.c
```diff
@@ -579,22 +608,33 @@ int set_download_mode(enum dload_mode mode)
...
-static void set_sdc_power_ctrl()
+static void set_sdc_power_ctrl(uint8_t slot)
 {
+	uint32_t reg = 0;
+
+	if (slot == 1) {
+		reg = SDC1_HDRV_PULL_CTL;
+	} else if (slot == 2) {
+		reg = SDC2_HDRV_PULL_CTL;
+	} else {
```
stephan-gh, line 619: this looks more like a switch case

target/msm8226/init.c
```diff
@@ -228,33 +230,53 @@ void target_sdc_init()
...
-	 * Set drive strength & pull ctrl for emmc
-	 */
-	set_sdc_power_ctrl();
-
+#ifndef FORCE_SDCARD
+	/* Try slot 1*/
 	config.bus_width = DATA_BUS_WIDTH_8BIT;
-	config.max_clk_rate = MMC_CLK_200MHZ;
-
-	/* Trying Slot 1*/
```
stephan-gh, line 239: Ooof, sad diff (try slot 1 vs trying slot 1) but at this point I'm not worrying too much anymore about the old branch :)

stephan-gh (approved):

PR #149, opened 2022-02-21
Random i2c_qup fixes and more clocks

stephan-gh:
See commits. Would appreciate a quick check of the addresses if time permits, easy to get confused there...


TravMurav (approved):
offsets seem fine for me compared to the hrd

stephan-gh:
@TravMurav Sorry, it turned out that my previous changes weren't enough, it's also necessary to add more cases in the msm8916 platform gpio code to set the pin function to I2C. I pushed an additional commit to calculate those pin numbers dynamically. Can you take another look? Thanks!

TravMurav (approved):

PR #155, opened 2022-03-04
Add support for MSM8909

TipzRickyCheung:
Boots fine on the devices (Qihoo W701 and Mobvoi TicWatch Pro) added, no display yet.

The W701 has a spi screen which I tried to implement, it no longer tells me that the display type is unknown.
The TicWatch Pro has a usual dsi display, still no display yet.
I have another msm8909 device with no display that I will try to implement as well.

We still need to think about how to handle buttons though, the TicWatch at least has two buttons but the W701 only has one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ https://source.codeaurora.org/quic/la/kernel/lk/) (tag `LA.BR.1.2.9.1-02310-8x16
 which is a heavily modified version of the [Little Kernel Embedded Operating System].
 
 ## Supported SoCs
+- `lk2nd-msm8909`: APQ8009, MSM8909
 - `lk2nd-msm8916`: APQ8016, MSM8216, MSM8916, MSM8929, MSM8939
 - `lk2nd-msm8974`: MSM8974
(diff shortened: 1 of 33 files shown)


stephan-gh:
I'm slowly working on a new branch that is based on a newer LK version from Qualcomm with support for more SoCs like MSM8909. It will be easier to make it work there without backporting as you did here.

If you want to play with it, it's at https://github.com/msm8916-mainline/lk2nd/tree/experimental-rebase but it's just a proof of concept dump so far, I did not actually port most lk2nd functionality from the main branch yet.

TipzRickyCheung:
> I'm slowly working on a new branch that is based on a newer LK version from Qualcomm with support for more SoCs like MSM8909. It will be easier to make it work there without backporting as you did here.
> 
> If you want to play with it, it's at https://github.com/msm8916-mainline/lk2nd/tree/experimental-rebase but it's just a proof of concept dump so far, I did not actually port most lk2nd functionality from the main branch yet.

Cool, I see that you have support for 8952, I could add all of my 8909 and my 8937 in there.

PR #156, opened 2022-03-14
Support GPT on sdcards

TravMurav:
useful for devices like db410c where sd can be a main GPT boot target

### DIFF
--- a/lib/partition/partition.c
+++ b/lib/partition/partition.c
@@ -251,7 +251,9 @@ int partition_publish(const char *device, off_t offset)
 
 				/* Some linux distros make use of subpartitions.
 				 * Scan some devices recursively to publish them. */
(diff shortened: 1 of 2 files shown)


lib/partition/partition.c
```diff
@@ -250,7 +251,9 @@ int partition_publish(const char *device, off_t offset)
 
 				/* Some linux distros make use of subpartitions.
 				 * Scan some devices recursively to publish them. */
-				if (strcmp(name, "system") == 0 || strcmp(name, "userdata") == 0)
+				if (strcmp(name, "system") == 0
+				    || strcmp(name, "userdata") == 0
+				    || strcmp(name, "rootfs") == 0)
```
stephan-gh, line 256: might be worth combining this with `bootable_parts` but who knows what the future will bring!

stephan-gh (approved):

PR #157, opened 2022-03-22
Inital msm8x26 lumia support

FieryFlames:
This PR includes Lk1st for msm8x26 along with a msm8926-v1-mtp dts with support for tesla, moneypenny, dempsey and superman. Makepeace is not included due to not having a tester who has one with msm8926.

The command mode autorefresh stuff is already enabled on tesla, whatever lk2nd does relating to this really messes tesla up. (I commented it out locally)

The membase must be changed for it to boot on our lumias for whatever reason, I used 0x07300000

### DIFF
--- a/dts/msm8226/msm8926-v1-mtp.dts
+++ b/dts/msm8226/msm8926-v1-mtp.dts
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	qcom,msm-id = <200 0>;
+	qcom,board-id = <8 0>;
+
+	/* All the Nokia devices are expected to use lk1st, which passes the 
+	compatible to lk2nd, so no need for extra device matching. 
+	These nodes are listed in no particular order. */
+	nokia-tesla {
+		model = "Nokia Lumia 830";
+		compatible = "nokia,tesla", "qcom,msm8926", "lk2nd,device";
+
+		lk2nd,smd-rpm-hack-opening;
+		// 107 is actually KEY_CAMERA_SHOT but lk2nd doesn't support that
+		lk2nd,keys =
+			<KEY_VOLUMEUP   106 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>,
+			<KEY_HOME   107 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>;
+	};
+
+	nokia-dempsey {
+		model = "Microsoft Lumia 640";
+		compatible = "nokia,dempsey", "qcom,msm8926", "lk2nd,device";
+
+		lk2nd,smd-rpm-hack-opening;
+		lk2nd,keys =
+			<KEY_VOLUMEUP   106 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>;
+	};
+
+	nokia-moneypenny {
+		model = "Nokia Lumia 630";
+		compatible = "nokia,moneypenny", "qcom,msm8926", "lk2nd,device";
+
+		lk2nd,smd-rpm-hack-opening;
+		lk2nd,keys =
+			<KEY_VOLUMEUP   106 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>;
+	};
+
+	nokia-superman {
+		model = "Nokia Lumia 730";
+		compatible = "nokia,superman", "qcom,msm8926", "lk2nd,device";
+
+		lk2nd,smd-rpm-hack-opening;
+		lk2nd,keys =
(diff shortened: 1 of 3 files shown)


dts/msm8226/msm8926-v1-mtp.dts
```diff
@@ -0,0 +1,49 @@
...
+
+	nokia-superman {
+		model = "Nokia Lumia 730"; // Also released under the Microsoft brand but let's keep it simple
+		compatible = "nokia,superman", "qcom,msm8926", "lk2nd,device";
+
+		lk2nd,smd-rpm-hack-opening;
+		lk2nd,keys =
+			<KEY_VOLUMEUP   106 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>;
+	};
+};
```
TravMurav, line 49: I think we want a newline at the end :)

dts/msm8226/msm8926-v1-mtp.dts
```diff
@@ -0,0 +1,50 @@
...
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	compatible = "qcom,msm8926-mtp", "qcom,msm8926";
+	qcom,msm-id = <200 0>;
+	qcom,board-id = <8 0>;
+
+	// All the Nokia devices are expected to use lk1st, which passes the compatible to lk2nd, so no need for extra device matching.
```
TravMurav, line 13: (re-adding it here since GH works now)
I'd prefer the comments/code to be folded at 80 where possible. Also could edit the readme :)

dts/msm8226/msm8926-v1-mtp.dts
```diff
@@ -0,0 +1,52 @@
...
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	compatible = "qcom,msm8926-mtp", "qcom,msm8926";
+	qcom,msm-id = <200 0>;
+	qcom,board-id = <8 0>;
+
+	/* All the Nokia devices are expected to use lk1st, which passes the
```
TravMurav, line 13: I guess a small nit would be to say that those nodes aren't really sorted but don't think we really kept them sorted in other files as well...

dts/msm8226/msm8926-v1-mtp.dts
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	compatible = "qcom,msm8926-mtp", "qcom,msm8926";
```
TravMurav, line 9: Not sure if we really care about the to level compatible as I think the dtbtool in this lk tree is not smart enough to properly generate id's for this one (it can for some files so we use compatible instead of id-s in those).

TravMurav (approved):
It seems fine to me other than a couple of nits, I guess I'd still like to see a readme edit as well but unsure if @stephan-gh wants to have lk1st targets there yet

dts/msm8226/msm8926-v1-mtp.dts
FieryFlames: I based this off msm8916-mtp dtb which has the top level compatible

TravMurav: msm8916-mtp has no top-level board/msm-id's though as they're autogenerated from that compatible :)

FieryFlames: Ohhh now I see..

TravMurav (approved):

dts/msm8226/msm8926-v1-mtp.dts
```diff
@@ -0,0 +1,61 @@
...
+	qcom,board-id = <8 0>;
+
+	/* All the Nokia devices are expected to use lk1st, which passes the 
+	compatible to lk2nd, so no need for extra device matching. 
+	These nodes are listed in no particular order. */
+	nokia-tesla {
+		model = "Nokia Lumia 830";
+		compatible = "nokia,tesla", "qcom,msm8926", "lk2nd,device";
+
+		lk2nd,smd-rpm-hack-opening;
```
stephan-gh, line 19: I think this is not needed anymore on 5.18? Or cherry-pick https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a8f8cc6b39b7ee0dbaccbebd1268c9d3458ebf13. I suppose it's fine for now but I will definitely drop this option in the new experimental-rebase branch.

stephan-gh (commented):
> I guess I'd still like to see a readme edit as well but unsure if @stephan-gh wants to have lk1st targets there yet

I think we definitely need some build instructions in the README for this, especially since it won't be part of the releases for now (I guess).

While this is compiled as lk1st it's strictly speaking more a lk2nd configuration. My main concern with documenting lk1st without care is that people just type in the commands, make some mistakes and maybe end up with an hard to recover device.

PR #162, opened 2022-03-30
Add MDP3 continuous splash support

Dominduchami:
Add modified code from efidroid (https://github.com/efidroid/bootloader_lk/blob/1aee6eef5326cdc9b921b6634e0caa4accc8c3f4/lib/2ndstage_display/target_display.c#L44) to support continuous splash for mdp3. Also create a variable so that mdp3 cont splash can be included instead of the mdp5 one (not sure if that's a good approach, will change it if there's something better)


lk2nd/target_display_cont_splash_mdp3.c
```diff
@@ -0,0 +1,45 @@
...
+
+static void mdss_mdp_cmd_kickoff(void) {
+	writel(0x00000001, MDP_DMA_P_START);
+	mdelay(10);
+}
+
+int mdp3_read_config(struct fbcon_config *fb)
+{
+	fb->base = (void *) readl(MDP_DMA_P_BUF_ADDR);
+	fb->width = readl(MDP_DMA_P_BUF_Y_STRIDE)/3;
```
konradybcio, line 31: isn't this supposed to be /4 for a8r8g8b8?

konradybcio: also shouldn't the format above be variable? or does it just work :tm:?

Dominduchami: > isn't this supposed to be /4 for a8r8g8b8?

/3 works fine, but I'm testing it on a Lumia 550, and Lumias used some different format afaik, so I'm not sure what to do here, might be necessary to change it as it was on msm8226 lumia lk2nd.

> also shouldn't the format above be variable? or does it just work tm?

It just works xd, it's the same in target_display_cont_splash_mdp5

lk2nd/target_display_cont_splash_mdp3.c
```diff
@@ -0,0 +1,45 @@
...
+
+
+static struct fbcon_config fb = {
+	.bpp    = 24,
+	.format = FB_FORMAT_RGB888,
+};
+
+static void mdss_mdp_cmd_kickoff(void) {
+	writel(0x00000001, MDP_DMA_P_START);
+	mdelay(10);
```
stephan-gh, line 25: Please rewrite this to use a thread like in mdp5, this improves the screen update delay.

lk2nd/target_display_cont_splash_mdp3.c
```diff
@@ -0,0 +1,45 @@
...
+}
+
+int mdp3_read_config(struct fbcon_config *fb)
```
stephan-gh, line 28: this function returns a value, you should do some tests (e.g. if `fb->base != NULL)`) like in mdp5.

lk2nd/target_display_cont_splash_mdp3.c
Dominduchami: Thank you for the review. I will try to rewrite it in a couple of days (as soon as I'll have access again to a linux machine)

stephan-gh:
Added together with mdp4 in https://github.com/msm8916-mainline/lk2nd/commit/7f273211ff0caf7439933c922a1e0e5dbd55a901, theoretically :)

PR #164, opened 2022-04-09
Add lk1st for huawei-g7

stephan-gh:


lk2nd/panel/oem_panel.c
```diff
@@ -43,6 +43,41 @@ uint32_t oem_panel_max_auto_detect_panels()
...
+		gpio_set_dir(enn_gpio.pin_id, 0); /* ENN */
+	}
+	return 0;
+}
+
+int target_ldo_ctrl(uint8_t enable, struct msm_panel_info *pinfo)
+{
+	int ret;
+
+	if (panel_select(LK1ST_PANEL) == panel_tianma_nt35521_5p5_720p_video_select)
```
TravMurav, line 72: Would it make any sense to prepare a switch...case here if we ever need to add more panel quirks? Don't think it's really important for now, but something to think about in the new branch

TravMurav (approved):

lk2nd/panel/oem_panel.c
stephan-gh: Can't switch case function pointers :(

PR #168, opened 2022-04-23
dts: msm8226: Add Huawei G6-L11

mirsal:
This patch adds minimal support for the MSM8926-based Huawei G6 4G (G6-L11).
The MSM8926 SoC being very close to MSM8226, I believe it does not need its own make target at this point.

### DIFF
--- a/dts/msm8226/rules.mk
+++ b/dts/msm8226/rules.mk
@@ -4,4 +4,5 @@ DTBS += \
 	$(LOCAL_DIR)/apq8026-asus-sparrow.dtb \
 	$(LOCAL_DIR)/apq8026-huawei-sturgeon.dtb \
 	$(LOCAL_DIR)/apq8026-lg-lenok.dtb \
-	$(LOCAL_DIR)/msm8226-samsung-ms013g.dtb
+	$(LOCAL_DIR)/msm8226-samsung-ms013g.dtb \
+	$(LOCAL_DIR)/msm8926-huawei-g6-l11-vb.dtb
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 
 ### lk2nd-msm8226
 - ASUS ZenWatch 2 - sparrow
-- Huawei Ascend G6 4G - G6-L11
+- Huawei Ascend G6 4G - G6-L11 (quirky - see comment in `dts/msm8226/msm8926-huawei-g6-l11-vb.dts`)
 - Huawei Watch - sturgeon
 - LG G Watch R - lenok
 - Samsung Galaxy Grand 2 - SM-G7102
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 
 ### lk2nd-msm8226
 - ASUS ZenWatch 2 - sparrow
+- Huawei Ascend G6 4G - G6-L11
 - Huawei Watch - sturgeon
 - LG G Watch R - lenok
 - Samsung Galaxy Grand 2 - SM-G7102
--- a/dts/msm8226/msm8926-huawei-g6-l11-vb.dts
+++ b/dts/msm8226/msm8926-huawei-g6-l11-vb.dts
@@ -2,6 +2,13 @@
 
 /dts-v1/;
 
+/*
+ * To build for huawei-g6-l11, comment out all dtb targets except
+ * $(LOCAL_DIR)/msm8926-huawei-g6-l11-vb.dtb in rules.mk in this directory.
+ * the stock g6-l11 bootloader does will not match this dtb unless it is
+ * in first position in the appended dtb.
+ */
+
 #include <skeleton.dtsi>
 
 / {
--- a/dts/msm8226/msm8926-huawei-g6-l11-vb.dts
+++ b/dts/msm8226/msm8926-huawei-g6-l11-vb.dts
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+
+/ {
+	model = "Huawei Ascend G6-L11 vb";
+	compatible = "huawei,g6-l11-vb", "qcom,msm8926", "lk2nd,device";
+
+	qcom,msm-id = <0xc8 0x1f51 0x00>;
+};


dts/msm8226/msm8926-huawei-g6-l11.dts
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+
+/ {
+	model = "Qualcomm MSM 8926 CDP G6-L11";
```
stephan-gh, line 8: This should be the real device name (e.g. `Huawei Ascend ...`).

dts/msm8226/msm8926-huawei-g6-l11.dts
```diff
@@ -0,0 +1,13 @@
...
+	compatible = "huawei,g6-l11", "qcom,msm8926-cdp", "qcom,msm8926", "lk2nd,device";
```
stephan-gh, line 9: Can drop the "qcom,msm8926-cdp", it's not needed

dts/msm8226/msm8926-huawei-g6-l11.dts
```diff
@@ -0,0 +1,13 @@
...
+
+	qcom,msm-id = <200 0x10000>;
+	qcom,board-id = <0 0>;
```
stephan-gh, line 12: Hm, this doesn't seem to match https://github.com/LineageOS/android_kernel_huawei_msm8226/blob/cm-12.1/arch/arm/boot/dts/msm8926-g6_l11-vb.dts#L25 or am I looking at the wrong device?

amonpaike: hi i have this device, helping mirsal testing... 
Indicate me how can i help..
The phone name i Huawei Ascend G6 l11 LTE
 
[440] Device model: Qualcomm MSM 8926 CDP G6-L11
[440] Found an appended flattened device tree (Qualcomm MSM 8926 CDP G6-L11 - 200 0 0 0x10000) 
[450] Device tree exact match the board: <200 0 65536 0x0> == <200 0 0 0x10001>
[460] LK would match our own DTB. Cool!
from the lk.log of lk2nd fixed by mirsal 

the The difference between MSM8926 and MSM8926 is the support of the LTE in the MSM8926

dts/msm8226/msm8926-huawei-g6-l11.dts
mirsal: Indeed, acknowledged.

dts/msm8226/msm8926-huawei-g6-l11.dts
mirsal: Acknowledged.

dts/msm8226/msm8926-huawei-g6-l11.dts
mirsal: @stephan-gh thanks a lot for your review. it it might be a different device, or maybe a different revision. the device this was tested on has the following in the dtb bundled with stock android images:

```
	model = "Qualcomm MSM 8926 CDP g6-l11-va";
	compatible = "qcom,msm8926-cdp", "qcom,msm8926", "qcom,cdp";
	qcom,msm-id = <0xc8 0x1f41 0x00>;
```
which is also different

lineageos has:

```
model = "Qualcomm MSM 8926 CDP g6-l11-va";
compatible = "qcom,msm8926-cdp", "qcom,msm8926", "qcom,cdp";
qcom,board-id = <8001 0>;
```
in https://github.com/LineageOS/android_kernel_huawei_msm8226/blob/cm-12.1/arch/arm/boot/dts/msm8926-g6_l11-va.dts

the stock bootloader matches the first dtb fragment appended to the flashed lk2nd image no matter what, so I used lk2nd logs along with trial and error in order to find matching values.

Any pointer to a better way would be greatly appreciated!

mirsal: If we use the exact values reported by lk2nd (`qcom,msm-id = <200 0>;` and `qcom,board-id = <0 0x10001>;`) then the result is the following:

```
[380] ### Ohai, this is lk2nd (or lk1st?) ###
[380] Board: platform: 200, foundry: 0x1, platform_version: 0x10001, platform_hw: 0x0, platform_subtype: 0x0, target: 0x0, baseband: 0x0, platform_hlos_subtype: 0x0
[400] pmic_info[0]: type: 0x10004, version: 0x20002, target: 0x20204
[400] Command line from primary bootloader: lk2nd androidboot.emmc=true androidboot.serialno=a9a2c30 androidboot.huawei_swtype=normal androidboot.huawei_bootmode=boot androidboot.localproppath=vodafone/default usb.pid.index=25 usb.serial=2008edb117cd androidboot.baseband=msm mdss_mdp.panel=1:dsi:0:qcom,mdss_dsi_tianma_otm9605a_4p5_qhd_video
[430] Device model: Huawei Ascend G6-L11
[440] Found an appended flattened device tree (Huawei Ascend G6-L11 - 200 0 65537 0x0)
[440] Device tree exact match the board: <200 0 0 0x10001> == <200 0 0 0x10001>
[450] Weird DTB selected by bootloader. Need to override. Sad.
```

stephan-gh:
```
qcom,msm-id = <200 0>, ...
qcom,board-id = <8001 0>;
```

is mostly equivalent to

```
qcom,msm-id = <0xc8 0x1f41 0x00> = <200 8001 0>;
```

So the two examples you show are the same. And one of them (preferably the first) is what I'd expect to have in lk2nd. :)

mirsal: Thanks, yes, that was my first guess but then I got confused by lk2nd logs. Fixing.

stephan-gh (approved):

dts/msm8226/rules.mk
```diff
@@ -1,6 +1,7 @@
 LOCAL_DIR := $(GET_LOCAL_DIR)
 
 DTBS += \
+	$(LOCAL_DIR)/msm8926-huawei-g6-l11-vb.dtb \
```
TravMurav, line 4: I think it's nicer to have this list sorted alphabetically

mirsal: @TravMurav Thanks for your review, I'm submitting another patch, keeping the grouping by SoC like for other targets.

mirsal: Unfortunately, the stock bootloader matches the asus zenwatch dtb instead when the g6-l11-vb one is after. On a previous version of this series, I added the g6-l11-va dts before the g6-l11-vb one, and the correct one was matched on a g6-l11-vb device, so the stock bootloader does not always match the first dtb, although order is important somehow.

stephan-gh: @mirsal So does it not work anymore at all with the latest changes you pushed here (since it isn't the first DTB anymore)?

mirsal: @stephan-gh Hi, indeed, the last changes do break it. You can probably close this PR, I'll open a new one if I get another opportunity at working on this device.

The strange thing is that the stock bootloader does not necessarily match the first dtb entry (as illustrated by the patch version with both va and vb, in which the va dtb appears first, but the vb dtb was correctly matched on a vb device) but in this case the bootloader still matches the asus-sparrow dtb as the g6-l11 dtb node appears after.

stephan-gh: You can add the typical `quirky - see comment in ...` comment in the README for now that usually suggests to build the lk2nd.img only with some selected DTBs instead of including all of them.

mirsal: done, I added the quirky notice and comment in a separate commit

stephan-gh (approved):

TravMurav (approved):

TravMurav:
Squashed the commits and merged, thanks!

PR #170, opened 2022-05-02
Experimental Boot Menu support

B83C:
Here's a few lines of codes that allow user to choose to boot from either EMMC or SD card. 
I have also modified some functions in dev/fbcon/fbcon.c to allow it draw in a lazy manner ( so it doesn't have to redraw on every change )
To enable it, add 'BOOT_MENU=1' to DEFINES in makefile


TravMurav:
Hi, the idea looks pretty cool but I think this should better be implemented as part of the big lk2nd rebase and refactoring that is going on right now. One of my plans for that is completely reworking the boot process which will likely result in rewriting the fs-boot as well, at which point it would be preferable to have a "fully-featured" boot menu...

TravMurav:
Hi, sorry for keeping this stale.
Lately we were slowly working on a big rewrite of lk2nd and tackling boot selection is also on the agenda (though it seems like lk2nd boot process and scope might change significantly in the future...)

Thanks for making this but I will close this PR for now.

PR #179, opened 2022-06-02
dts: msm8974: Add Sony Xperia Z3

z3ntu:
The bootloader seems to use the dts of the boot partition to initialize
the panel, so a bunch of panel dts is included here.

---
IMG_20220602_190657

Unfortunately booting something from lk2nd doesn't seem to work, it just boots into "S1 Service" which seems to be some kind of crashdump mode (among others). Also e.g. `fastboot oem lk_log && fastboot get_staged /dev/stdout` just hangs (at the get_staged command)

Others work just fine though:
```
$ fastboot oem dump-speedbin
(bootloader) PVS valid: 1, Redundant fuse select: 0
(bootloader) Speed bin: 3, PVS: 5, PVS version: 1 (speed3-pvs5-v1)
```

Maybe some buffer is smaller on this device compared to other msm8974? (side note but I also saw super weird behavior in the scratch buffer on msm8226 htc-memul so seems some OEMs love modifying core stuff)

### DIFF
--- a/dts/msm8974/msm8974pro-ac-pm8941-mtp.dts
+++ b/dts/msm8974/msm8974pro-ac-pm8941-mtp.dts
@@ -37,3 +37,12 @@
 			<KEY_VOLUMEDOWN PM_GPIO(2) PM_GPIO_PULL_UP_1_5>;
 	};
 };
+
+/*
+ * The stock bootloader of sony-leo uses parts of the dts to initialize the
+ * screen. So to get continous splash initialized by the first bootloader we
+ * need to include these dts parts here.
+ *
+ * FIXME: This won't work if more Sony devices are added here...
+ */
+#include "sony/msm8974pro-ac-shinano_leo-display.dtsi"
--- a/dts/msm8974/msm8974pro-ac-pm8941-mtp.dts
+++ b/dts/msm8974/msm8974pro-ac-pm8941-mtp.dts
@@ -22,4 +22,18 @@
 			<KEY_VOLUMEUP   PM_GPIO(5) PM_GPIO_PULL_UP_1_5>,
 			<KEY_VOLUMEDOWN PM_GPIO(2) PM_GPIO_PULL_UP_1_5>;
 	};
+
+	sony-leo {
+		model = "Sony Xperia Z3";
+		compatible = "sony,xperia-leo", "qcom,msm8974", "lk2nd,device";
+		/*
+		 * Probably matches other Sony devices as well, but good enough
+		 * for now.
+		 */
+		lk2nd,match-cmdline = "* androidboot.bootloader=s1 *";
+
+		lk2nd,keys =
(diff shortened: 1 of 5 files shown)


dts/msm8974/msm8974pro-sony-xperia-shinano-leo.dts
```diff
@@ -0,0 +1,1356 @@
...
+	model = "Sony Xperia Z3";
+	compatible = "sony,xperia-leo", "qcom,msm8974", "lk2nd,device";
+	qcom,msm-id = <194 8 0x10000>;
+	qcom,board-id = <8 0>;
+
+	lk2nd,keys =
+		<KEY_VOLUMEUP   PM_GPIO(5) PM_GPIO_PULL_UP_1_5>,
+		<KEY_VOLUMEDOWN PM_GPIO(2) PM_GPIO_PULL_UP_1_5>;
+
+	soc {
```
TravMurav, line 18: I think we usually move things like this into a dedicated .dtsi
like this: https://github.com/msm8916-mainline/lk2nd/blob/master/dts/msm8916/motorola-carrier-channel-ids.dtsi

dts/msm8974/msm8974pro-sony-xperia-shinano-leo.dts
```diff
@@ -0,0 +1,1356 @@
...
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	model = "Sony Xperia Z3";
+	compatible = "sony,xperia-leo", "qcom,msm8974", "lk2nd,device";
+	qcom,msm-id = <194 8 0x10000>;
+	qcom,board-id = <8 0>;
```
TravMurav, line 12: this smells like `qcom,msm8974_pro-mtp` to me but not sure

z3ntu:
@TravMurav Any idea why booting anything doesn't work on this device? Especially on this I'd like to use lk2nd because of these panel oddities but currently it's not usable for that :(

z3ntu:
From my side this works fine now, the original problems I had only because I had the bootloader from experimental Android 7 release for this device installed, which did all kinds of weird things apparently.

Anyways, confirmed booting mainline Linux and downstream TWRP works fine with this.

dts/msm8974/msm8974pro-sony-xperia-shinano-leo-mdss.dtsi
```diff
@@ -0,0 +1,1340 @@
+/ {
```
TravMurav, line 1: Since you import this, can you also inherit the copyright boilerplate like in https://github.com/msm8916-mainline/lk2nd/blob/master/dts/msm8916/motorola-carrier-channel-ids.dtsi

z3ntu: done

TravMurav (approved):

PR #183, opened 2022-07-03
dts: msm8226: new device (LG Fx0, lg-madai)

peremen:
This is an initial attempt to port lk2nd to Fx0. Although it can enter the fastboot mode, booting any vendor kernel image leads to "TZ Crash" screen and directs to some internal debugging mode. If I boot the same kernel image through the device's native bootloader or fastboot then it boots fine. Currently I am experimenting with mainline kernel.

Also lk2nd doesn't look to correctly detect internal eMMC: when it is booted Variant says "Unknown eMMC".

Edit: it was my fault, at least flashing lk2nd within lk2nd works.

### DIFF
--- a/dts/msm8226/msm8926-lg-madai.dts
+++ b/dts/msm8226/msm8926-lg-madai.dts
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <200 0>;
+	qcom,board-id = <150 0>;
+
+	model = "LG Fx0";
+	compatible = "lg,madai", "qcom,msm8926", "lk2nd,device";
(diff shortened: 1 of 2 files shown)


dts/msm8226/msm8926-lg-madai.dts
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <200 0>, <224 0>, <200 0x10001>, <224 0x10001>;
```
stephan-gh, line 10: You should only have one entry here, probably `<200 0>` for MSM8926.

peremen: The vendor kernel's msm-id specifies all four, and the vendor fastboot log says `200/150/0x10001/0` for msm-id and board-id combination, but lk2nd loads using msm-id `<200, 0>`. So I am changing it.

PR #194, opened 2022-08-17
Add serial-number node to DTB when booting mainline (experimental-tmp)

FieryFlames:


stephan-gh:
Why? Is this a standardized property?

FieryFlames:
Yes, it's in device tree spec 0.4

stephan-gh:
This question still stands though: Why? :)

FieryFlames:
Just more device personalization, it ideally would be shown in settings but I haven't checked if thats implemented in phosh or plasma yet. Not really necessary tbh.. I saw it being done in m1n1 for Apple machines and wanted to try it here, but I guess it makes more sense there.

stephan-gh:
AFAICT this is only read by Linux on ARM32, and only to show it in `/proc/cpuinfo` (wtf). It doesn't really seem to be used for anything :confused:

FieryFlames:
Is that just in the kernel? Maybe userspace does more with it?

stephan-gh:
I wouldn't expect userspace to read `/sys/firmware/device-tree/base/...` personally :thinking:

samueldr:
Given it's specified as a required property in the spec, I think this alone should trump any other "but why?" questions.

stephan-gh:
> Given it's specified as a required property in the spec, I think this alone should trump any other "but why?" questions.

@samueldr No, it's marked as O (= Optional). It's not even OR (= Optional but Recommended) :)

samueldr:
Oh, oops, my bad. I missed the legend and the column, and assumed the wording in 3.1 and the note meant all properties were required.

samueldr:
Are there other ways the OS can ask/request for the serial number of the device? If there are, I guess it doesn't matter then.

stephan-gh:
> Are there other ways the OS can ask/request for the serial number of the device? If there are, I guess it doesn't matter as much then.

The serial number is usually just taken from the eMMC/UFS/flash and is probably available there through sysfs or similar.

I guess I'm not opposed to add the `serial-number`, but the changes itself need some more thought:

- `lk2nd_dev.serialno` exists only for lk2nd configurations but not on lk1st. This won't compile correctly there (this is not tested properly by CI atm).
- The whole thing probably should not be part of the device module but separate (similar to `lk2nd/mac` since LK usually has its own serial number.

FieryFlames:
Will sort out those issues. I wasn't sure whether or not it should go in a separate module or in device, because panel is in device and mac isn't

stephan-gh:
> because panel is in device and mac isn't

Because panel actually requires information from the device DTS, while MAC does not :)

stephan-gh:
Integrated into experimental-tmp-gpio, thanks!

PR #198, opened 2022-09-25
Add fastboot oem commands to access SPMI/PMIC

fxsheep:
Usage:

```
fastboot oem spmi read <address>
fastboot oem spmi write <address> <value>
```

### DIFF
--- a/app/aboot/fastboot-extra.c
+++ b/app/aboot/fastboot-extra.c
@@ -8,6 +8,7 @@
 #include <platform.h>
 #include <platform/iomap.h>
 #include <platform/timer.h>
+#include <pm8x41_hw.h>
 #include <pm8x41_regulator.h>
 #include <reg.h>
 #include <smd.h>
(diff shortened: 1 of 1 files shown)


fxsheep:
Do note that this two commands (among many other lk2nd debugging-orientated features) give the possibility to severely and permanently damage the hardware (e.g. one wrong voltage and half of motherboard is fried). IMHO lk2nd lacks a warning/precaution of such a risk.

TravMurav:
Hi, We are considering to obsolete the current lk2nd branch in favor of the different one (currently in `experimental-tmp`) in the near future, so I'd appreciate if you could target that with your PR. This should be pretty easy, please see e.g. https://github.com/msm8916-mainline/lk2nd/blob/experimental-tmp/lk2nd/fastboot/misc.c

The file will need to be conditionally compiled for those targets that have the pmic.

fxsheep:
Thanks for clarification and advice. I'm closing this temporarily.

stephan-gh:
I'd still like to keep this open as a reminder. Basically the idea is to have a `lk2nd/fastboot/debug` module in the new branch which would implement functionality such as the `readl`/`writel` or something like this.

I suppose we could disable it by default if it's considered too dangerous (although generally I'd hope users are reasonable enough to avoid using commands they do not understand).

stephan-gh:
I added `fastboot oem debug readpmic` and `fastboot oem debug writepmic` in the experimental-tmp3 branch of the new lk2nd rebase. Thanks for the idea!

PR #205, opened 2022-10-24
dts: msm8974: samsung-klte: panel selection

minlexx:
Add support for overwriting panel compatible for Samsung Galaxy S5 for 2nd variation of the panel.

Default is considered to be s6e3fa2; add support for detecting ea8064g.

This is needed to submit panel drivers to mainline, as upstream requires them to be separate and not one combined driver.

### DIFF
--- a/dts/msm8974/msm8974-samsung-r14.dts
+++ b/dts/msm8974/msm8974-samsung-r14.dts
@@ -12,5 +12,25 @@
 		model = "Samsung Galaxy S5 (SM-G900F)";
 		compatible = "samsung,klte", "qcom,msm8974", "lk2nd,device";
 		lk2nd,match-bootloader = "G900F*";
+
+		/*
+		* samsung-klte has at least 2 known panel variations:
+		* - S6E3FA2 (mdss_mdp.panel=1:dsi:0:mdss_dsi_samsung_1080p_cmd_fa2)
+		* - EA8064G (mdss_mdp.panel=1:dsi:0:mdss_dsi_samsung_1080p_cmd_mag)
+		* Upstream mainline kernel already has "samsung,s6e3fa2" compatible set
+		* for panel in dts, and we'll not change that. We'll only do
+		* replacement for ea8064g one (replace samsung,s6e3fa2 to
+		* samsung,ea8064g).
+		*/
+		panel {
+			/* Default is Samsung S6E3FA2 */
+			compatible = "samsung,s6e3fa2";
+			replace-compatible;
+
+			/* MagnaChip EA8064G */
+			mdss_dsi_samsung_1080p_cmd_mag {
+				compatible = "samsung,ea8064g";
+			};
+		};
 	};
 };
--- a/lk2nd/lk2nd-device.c
+++ b/lk2nd/lk2nd-device.c
@@ -257,6 +257,7 @@ static void lk2nd_parse_panels(const void *fdt, int offset)
 	struct lk2nd_panel *panel = &lk2nd_dev.panel;
 	const char *old, *new, *ts;
 	int old_len, new_len, ts_len;
+	bool replace_mode = false;
 
 	offset = fdt_subnode_offset(fdt, offset, "panel");
 	if (offset < 0)
@@ -265,6 +266,8 @@ static void lk2nd_parse_panels(const void *fdt, int offset)
 	old = lkfdt_getprop_str(fdt, offset, "compatible", &old_len);
 	if (!old || old_len < 1)
 		return;
+
+	replace_mode = !!fdt_getprop(fdt, offset, "replace-compatible", NULL);
 
 	offset = fdt_subnode_offset(fdt, offset, panel->name);
 	if (offset < 0) {
@@ -285,6 +288,9 @@ static void lk2nd_parse_panels(const void *fdt, int offset)
 
 	strlcpy((char*) panel->compatible, new, new_len);
 	strlcpy((char*) panel->old_compatible, old, old_len);
+
+	if (replace_mode)
+		panel->compatible_size = new_len;
 
 	ts = lkfdt_getprop_str(fdt, offset, "touchscreen-compatible", &ts_len);
 	if (!ts || ts_len < 1)


dts/msm8974/msm8974-samsung-r14.dts
```diff
@@ -13,4 +13,23 @@
...
 	};
+
+	/*
+	 samsung-klte has at least 2 known panel variations:
+	   * S6E3FA2 (mdss_mdp.panel=1:dsi:0:mdss_dsi_samsung_1080p_cmd_fa2)
+	   * EA8064G (mdss_mdp.panel=1:dsi:0:mdss_dsi_samsung_1080p_cmd_mag)
+	 Upstream mainline kernel already has "samsung,s6e3fa2" compatible set
+	   for panel in dts, and we'll not change that. We'll only do
+	   replacement for ea8064g one (replace samsung,s6e3fa2 to
+	   samsung,ea8064g).
```
stephan-gh, line 24: Please use the

```
/*
 * Hello World!
 */
```

comment format

dts/msm8974/msm8974-samsung-r14.dts
```diff
@@ -13,4 +13,23 @@
...
+	*/
+	panel {
+		/* Default is Samsung S6E3FA2 */
+		compatible = "samsung,s6e3fa2";
+
+		/* MagnaChip EA8064G */
+		mdss_dsi_samsung_1080p_cmd_mag {
+			compatible = "samsung,ea8064g";
```
stephan-gh, line 32: Note that due to the way the panel detection is implemented at the moment this will effectively result in `compatible = "samsung,ea8064g", "samsung,s6e3fa2";`. When the kernel is missing support for `ea8064g`, it would still use the `s6e3fa2` driver even if it is known that it is the wrong one.

This works well for the devices with "generic" compatible in lk2nd, because there it results in something like `compatible = "samsung,s6d78a0-gh9607501a", "samsung,gprime-panel"`. Personally I prefer the approach with the generic compatible because it ensures that users never accidentally run the wrong panel driver, which could, theoretically, cause permanent hardware damage (although I have no heard of such a case so far).

But I'll leave this up to you I think :)

minlexx: First, the kernel will have support for both variations anyway.

Second, panel driver checks panel's MTP ID and refuses to power on unknown panel. Permanent hw damage is not possible.

Third, what? So it is not replacing, it is only prepending? This doesn't seem right, we need a mechanism to fully replace the compatible string. I do not think "generic compatible" is upstreamable right now. Maybe I can add handling for new property like replace-compatible = "xxx" ...?

Though, `compatible = "samsung,ea8064g", "samsung,s6e3fa2";` should still work for my use case. But I couldn't confirm that compatible string is changed at all during my brief test yesterday.

dts/msm8974/msm8974-samsung-r14.dts
minlexx: :ok:

dts/msm8974/msm8974-samsung-r14.dts
minlexx: I'm not sure I've got all bloody C pointer arithmetics right, I meant something like this: https://github.com/msm8916-mainline/lk2nd/pull/205/commits/e67e5232757743a2584e3b3c1beeabc58881a608

stephan-gh: > First, the kernel will have support for both variations anyway.

Not necessarily I think, what if someone boots older MSM8974 releases with these lk2nd changes? It will set the `"samsung,ea8064g"` compatible but the old kernel does not have a driver for it.

> Second, panel driver checks panel's MTP ID and refuses to power on unknown panel. Permanent hw damage is not possible.

That is a good point. Note that most other panel drivers do not have such checks though.

> Third, what? So it is not replacing, it is only prepending? This doesn't seem right, we need a mechanism to fully replace the compatible string.

I'm not opposed to add support to replace the compatible string completely, with something like your proposed patch. I'm not sure why the existing code does not seem right to you though: it works exactly as intended. It just does not fit exactly to your specific requirements. :)

> I do not think "generic compatible" is upstreamable right now. 

Why? My thought behind it is the following:

It is a well-established convention that device-specific details are left up to the bootloader to fill in. Take `/memory` for instance: It is usually set to dummy values (0) by default and the bootloader is expected to fill in the amount of RAM at runtime. We could hardcode the most common amount of RAM but we usually don't, perhaps because this increases the chance that users run an incorrect device tree and get confused.

The generic panel compatible is not just a dummy value though. It is kept by lk2nd (by prepending the more specific compatible) because I argue that it has semantic value. Consider the following example:

The device tree source has:

```dts
compatible = "samsung,gprime-panel";
```

This describes exactly what we know for sure at compile time: The device has one of the possible panels of the `samsung,gprime` device. We don't know which panel exactly.

Having this generic compatible might seem useless because no driver in Linux will bind to it. But this is just one possible implementation: Theoretically, you could write a generic driver for `"samsung,gprime-panel"` by implementing the panel detection directly in Linux or another operating system.

When using lk2nd, this compatible is changed at runtime to:

```dts
compatible = "samsung,s6d78a0-gh9607501a", "samsung,gprime-panel";
```

Semantically this means: The panel is compatible with `samsung,s6d78a0-gh9607501a`, and more in general also `samsung,gprime-panel`.

Now the operating system (e.g. Linux) is relieved of doing the panel detection itself. Linux just binds a driver directly to `samsung,s6d78a0-gh9607501a`. If the specific panel is not yet supported by any driver in Linux then the panel is automatically disabled at runtime.

However, an operating system with the theoretic generic driver from above could still work correctly. It would ignore the `"samsung,s6d78a0-gh9607501a"` compatible and just binds the driver to `"samsung,gprime-panel"`.

This is quite similar to this example from the device tree specification:

> `compatible = "fsl,mpc8641", "ns16550";`
> In this example, an operating system would first try to locate a device driver that supported fsl,mpc8641. If a driver was not found, it would then try to locate a driver that supported the more general ns16550 device type.

minlexx:
I've found the reason the replacement (also as prepending) compatible did not work - need to have `panel {}` node inside the `klte {}` node, not under root `/` node. Now it works as it should, tested

dts/msm8974/msm8974-samsung-r14.dts
minlexx: >  what if someone boots older MSM8974 releases with these lk2nd changes? It will set the "samsung,ea8064g" compatible but the old kernel does not have a driver for it.

It does have a driver for it. S6E3FA2 panel driver in older releases supports both `samsung,s6e3fa2` [1] and `samsung,ea8064g` compatibles. It ignores the difference and reads an actual MTP ID using single-byte read commands [2]. Then it checks is MTP ID is among supported ones, and only then operates accordingly.

In new releases drivers will support each their own compatible, but will still check MTP ID and refuse to work on unknown hardware.

I thnk, safety should be pretty okay here, hardware damage should be prevented in all cases.

[1] https://github.com/msm8974-mainline/linux/blob/qcom-msm8974-5.19.y/drivers/gpu/drm/panel/panel-samsung-s6e3fa2-full.c#L1158
[2] https://github.com/msm8974-mainline/linux/blob/qcom-msm8974-5.19.y/drivers/gpu/drm/panel/panel-samsung-s6e3fa2-full.c#L763

minlexx: > Note that most other panel drivers do not have such checks though.

I'm not suggesting you have to switch all devices to new mechanism. Can klte at least or now have a specific compatible, especially given the fact that it is already upstream in dts? _(And it's corresponding panel driver is not upstream, because I need to rewrite it before I send it again, and for this I need panel selection for klte to work in lk2nd)_



>>   I do not think "generic compatible" is upstreamable right now.

> Why? My thought behind it is the following:

What you're saying makes some sense to me, but I'll actually believe it's upstreamable the moment I see something like `compatible = "samsung,a5lte-panel"` in torvalds/linux repo. Until then it's only a theory :)

> It is a well-established convention that device-specific details are left up to the bootloader to fill in.

Right, but in this case we know exactly which specific panel is inside the device, so we can fill compatible property with exact specific value, so why do we even need generic compatible then? :)

Also, you've probably seen this recently: https://www.spinics.net/lists/linux-input/msg82097.html
```
>    compatible:
>      enum:
> +      - imagis,ist3038
>        - imagis,ist3038c
> +      - imagis,ist30xxb

No wildcards. Real model numbers.
```
Clearly, specific compatibles are preferred over generic or "wildcard" ones.

---

With all that said, I think, prepending specific compatible should also work for my use case, just because that's how compatibles work, right? So in case `compatible = "samsung,ea8064g", "samsung,s6e3fa2";` ea8064g driver will be tried first because it comes first. It's not safer to "fallback" to fa2 driver, it's safe and backwards compatible in both cases. In this case of prepending just a few extra of bytes of memory will be wasted in fdt blob :)

stephan-gh:
If klte also works fine with the existing mechanism and just wastes a few bytes in the DTB there I think would prefer using it rather than introducing the new `replace-compatible`. This keeps the code simpler and avoids having to port it over to the new lk2nd branch (which already has something equivalent to the current code).

minlexx:
So, new lk2nd can replace the compatible or it does prepending too?

stephan-gh:
It works the same as the current lk2nd, it prepends the compatible. But you said above that this would actually work for you as well :)

minlexx:
> This keeps the code simpler and avoids having to port it over to the new lk2nd branch

This is software, it should be the smart one, sometimes complicated, in general. But in this case we're talking about ~20 line change, is it complicated??

Also "the new lk2nd branch" which branch we're talking about, I can port it there too

stephan-gh:
> This is software, it should be the smart one, sometimes complicated, in general (unlike hardware, which can be dumb and require workarounds in sw). But in this specific case we're talking about ~20 line change, is it complicated??

You said yourself above that prepending would also work just fine in your case, so I just picked up your own suggestion (end of this comment https://github.com/msm8916-mainline/lk2nd/pull/205#discussion_r1027371168). :)

Is there any advantage of adding a special case for klte if both approaches have the same functional result?

minlexx:
I'm not denying I said that, though I already regret that a bit, because that's giving you an easy way out of this dispute, so you can save your time on reading 20 lines of code.

That's why I don't do upstreaming, every MR that is not trivial is an impossible fight. I'll give up I guess, and next time I'll avoid sending anything

stephan-gh:
> I'm not denying I said that, though I already regret that a bit, because that's giving you an easy way out of this dispute, so you can save your time on reading 20 lines of code.

I apologize if I have offended you in some way. This was always intended to be a technical discussion and not a dispute. :)

I would like to make one thing absolutely clear though: **I have never tried to get around reading your code changes.** On the contrary - I have already looked at them weeks ago when you initially posted it. I looked again 5 days ago and in more detail earlier today before writing my comment. I have checked the panel code in the new lk2nd branch to see how it could be integrated there. I already thought about suggesting a potential simplification of your code, by keeping the allocation code as-is and just changing `panel->compatible_size` after the two compatibles were already copied. I considered slight changes to the representation in the DT, perhaps something like

```
		panel {
			/* Default is Samsung S6E3FA2 */
			compatible = "samsung,s6e3fa2";
			replace-compatible;

			/* MagnaChip EA8064G */
			mdss_dsi_samsung_1080p_cmd_mag {
				compatible = "samsung,ea8064g";
			};
		};
```

because it wouldn't make sense to have some subnodes using `replace-compatible` and some subnodes using `compatible` (it is kind of a decision that should be made in the outer panel node).

I have not written anything about this because of this part in your comment:

> With all that said, I think, prepending specific compatible should also work for my use case, just because that's how compatibles work, right? So in case `compatible = "samsung,ea8064g", "samsung,s6e3fa2";` ea8064g driver will be tried first because it comes first. It's not safer to "fallback" to fa2 driver, it's safe and backwards compatible in both cases. In this case of prepending just a few extra of bytes of memory will be wasted in fdt blob :)

This sounded to me like you are fine with both approaches (use prepended compatible or add the special replace-compatible code). I am also fine with both approaches so I just suggested going with the easier one.

If you would prefer to use the `replace-compatible` for some reason you just need to say that clearly. The reason could be something very simple like saying that `compatible = "samsung,ea8064g", "samsung,s6e3fa2"` is misleading since the EA8064G panel is not "compatible" with S6E3FA2 (it's a completely different panel).

This wasn't clear to me so far. Also, I believe you have completely misunderstood my reasoning for suggesting to "keep the code simpler". I hope this comment clears up our misunderstanding. :)

TravMurav (approved):

lk2nd/lk2nd-device.c
```diff
@@ -273,15 +274,33 @@ static void lk2nd_parse_panels(const void *fdt, int offset)
...
 
 	new = lkfdt_getprop_str(fdt, offset, "compatible", &new_len);
-	if (!new || new_len < 1)
-		return;
+	if (!new || new_len < 1) {
+		new = lkfdt_getprop_str(fdt, offset, "replace-compatible", &new_len);
+		if (!new || new_len < 1)
+			return;
+		replace_mode = true;
+		dprintf(INFO, "Panel compatible will be *replaced*\n");
```
TravMurav, line 282: not sure if this message adds much value

lk2nd/lk2nd-device.c
```diff
@@ -273,15 +274,33 @@ static void lk2nd_parse_panels(const void *fdt, int offset)
...
+	} else {
+		panel->compatible_size = ++new_len;
+	}
 
 	panel->compatible = malloc(panel->compatible_size);
 	ASSERT(panel->compatible);
-	panel->old_compatible = panel->compatible + new_len;
+
+	if (!replace_mode) {
+		/* Old compatible points into the middle of allocated block? */
```
TravMurav, line 296: Yes, due to it being `<new compatible>\0<old compatible>\0` so the comment is out of place

lk2nd/lk2nd-device.c
```diff
@@ -273,15 +274,33 @@ static void lk2nd_parse_panels(const void *fdt, int offset)
...
+		panel->old_compatible = panel->compatible + new_len;
+	} else {
+		/* Allocate separate memory for old compatible */
```
TravMurav, line 299: Do we need to if we are replacing it? (I suppose this is what was suggested)

lk2nd/lk2nd-device.c
minlexx: Yes, it was a debug print and plan was to remove it anyway

minlexx: fixed

lk2nd/lk2nd-device.c
minlexx: this code is not present now

lk2nd/lk2nd-device.c
minlexx: this code is not present now

minlexx:
> I considered slight changes to the representation in the DT ... because it wouldn't make sense to have some subnodes using `replace-compatible` and some subnodes using `compatible` (it is kind of a decision that should be made in the outer panel node).

I've implemented yuour suggestion, `replace_mode` is now being checked by `fdt_getprop` in outer `panel` node.

> I already thought about suggesting a potential simplification of your code, by keeping the allocation code as-is and just changing `panel->compatible_size` after the two compatibles were already copied.

Now old allocation code is left as is, two compatibles are copied into the buffer, and after tjhe copy is complete, `panel->compatible_size` is truncated to be only `new_len`, so that later when `lk2nd_update_panel_compatible()` is executed, only first part with new compatible is written, effectively doing replacement instead of prepending.

I've tested that is works, first by spreading multiple debug prints (now removed in final version), then by booting
image

> If you would prefer to use the `replace-compatible` for some reason you just need to say that clearly. The reason could be something very simple like saying that `compatible = "samsung,ea8064g", "samsung,s6e3fa2"` is misleading since the EA8064G panel is not "compatible" with S6E3FA2 (it's a completely different panel).

Right, I should have said that clearly, I prefer to replace the compatible in this case, because of what you've said.

Sorry for the delay

lk2nd/lk2nd-device.c
```diff
@@ -286,6 +289,9 @@ static void lk2nd_parse_panels(const void *fdt, int offset)
 	strlcpy((char*) panel->compatible, new, new_len);
 	strlcpy((char*) panel->old_compatible, old, old_len);
 
+	if (replace_mode)
+		panel->compatible_size = new_len;
```
minlexx, line 293: we can make this even simpler, get rid of dedicated `replace_mode` variable and just write here
```c
	if (fdt_getprop(fdt, offset, "replace-compatible", NULL))
		panel->compatible_size = new_len;
```
this will make it only 2 line change to lk2nd-device.c

TravMurav: This would look nicer indeed

stephan-gh:
Looks good to me aside from the possible simplification you already noticed yourself. Thanks for making the changes! :+1:

lk2nd/lk2nd-device.c
minlexx: This didn't work nicely because `offset` variable points into wrong node inside fdt at this point in code
```
offset1:	panel {
			/* Default is Samsung S6E3FA2 */
			compatible = "samsung,s6e3fa2";
			replace-compatible;

			/* MagnaChip EA8064G */
offset2:		mdss_dsi_samsung_1080p_cmd_mag {
				compatible = "samsung,ea8064g";
			};
		};
```
offset1 is the value after `offset = fdt_subnode_offset(fdt, offset, "panel");`
offset2 is after `offset = fdt_subnode_offset(fdt, offset, panel->name);`
and `replace-compatible` is in the parent `panel` node now.

minlexx: I had to revert this attempt, I don't see how I can do this without introducing extra variables, so I guess having `bool replace_mode` is not so bad then

stephan-gh (approved):

TravMurav:
Thanks!

Also staged the same change for the -next: https://github.com/msm8916-mainline/lk2nd/commit/0847293ad736172a7dfef4fc356e1659c9721256

PR #210, opened 2022-11-08
Docs fixes

pevik:
Typos + whitespace.


TravMurav (approved):
Thanks for going through, I should probably learn to run aspell on my texts :D

PR #212, opened 2022-11-09
dts: msm8939: Use match-panel for Aquaris M5 (Longcheer L9100)

a-andre:


dts/msm8916/msm8939-qrd-skuk.dts
```diff
@@ -9,9 +9,17 @@
...
-		compatible = "bq,piccolo", "qcom,msm8939", "lk2nd,device";
-		lk2nd,match-cmdline = "* mdss_mdp.panel=1:dsi:0:qcom,mdss_dsi_truly_nt35695_1080p_cmd:1:none";
+		model = "BQ Aquaris M5 (Longcheer L9100)";
+		compatible = "longcheer,l9100", "qcom,msm8939", "lk2nd,device";
+		lk2nd,match-panel;
+
+		panel {
+			compatible = "longcheer,l9100-panel";
+
+			qcom,mdss_dsi_truly_nt35695_1080p_cmd {
```
stephan-gh, line 19: Please also add entries for the other possible panels, see https://github.com/msm8916-mainline/android_kernel_qcom_msm8916/blob/901d0b2ca05c0b0af49786fb548eb0fc289867ff/arch/arm/boot/dts/qcom/piccolo.dtsi#L13-L16

They should be added here but there is no need to add them to linux-panel-drivers yet since we have no way of knowing if the panel drivers will actually work correctly.

dts/msm8916/msm8939-qrd-skuk.dts
```diff
@@ -9,9 +9,26 @@
...
-		lk2nd,match-cmdline = "* mdss_mdp.panel=1:dsi:0:qcom,mdss_dsi_truly_nt35695_1080p_cmd:1:none";
+		model = "BQ Aquaris M5 (Longcheer L9100)";
+		compatible = "longcheer,l9100", "qcom,msm8939", "lk2nd,device";
+		lk2nd,match-panel;
+
+		panel {
+			compatible = "longcheer,l9100-panel";
+
+			qcom,mdss_dsi_truly_nt35695_1080p_video {
+				compatible = "longcheer,truly-nt35695";
```
a-andre, line 20: With piccolo moving to the `_cmd` variant, what naming do you suggest for the compatible string to distinguish between `truly_nt35695_1080p_video` and `truly_nt35695_1080p_cmd`?

stephan-gh: I think the _video one should also work with the modified _cmd driver, so just keep assigning the same compatible to both.

stephan-gh:
Pushed, thanks!

PR #214, opened 2022-11-11
[experimental-tmp] Add Nokia feature phones and partial goods support

affenull2345:
Let me know if I should create separate pull requests for the different changes. ~~The msm8996 partial goods code is untested, so I'm not sure if it can be kept as-is. Also, I can't think of a proper match rule for the 8110 since the bootloader doesn't pass anything special except for a rather generic-looking panel ID.~~

### DIFF
--- a/lk2nd/partial-goods/partial-goods.c
+++ b/lk2nd/partial-goods/partial-goods.c
@@ -1,0 +1,41 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/* Copyright (c) 2022, Otto Pflüger <affenull2345@gmail.com> */
+
+#include <platform/partial_goods.h>
+#include <boot.h>
+#include <lk2nd/util/lkfdt.h>
+
+#if TARGET_MSM8909
+static const char *cpu_disable_paths[] = {
+	"/cpus/cpu@2",
+	"/soc@0/power-manager@b0a8000",
+	"/soc@0/power-manager@b0a9000",
+	"/cpus/cpu@3",
+	"/soc@0/power-manager@b0b8000",
+	"/soc@0/power-manager@b0b9000",
+	NULL,
+};
+#endif
+
+static int lk2nd_partial_goods_dt_update(void *dtb, const char *cmdline,
+					 enum boot_type boot_type)
+{
+#if TARGET_MSM8909
+	int flag = platform_partial_goods_val();
+
+	switch (flag) {
+		case 0:
+		case 1:
+			/* Dual-core MSM8905 variant */
+			lkfdt_setprop_by_node_paths(dtb,
+				"status", "fail", sizeof("fail"),
+				cpu_disable_paths);
+			break;
+		default:
+			/* Full-featured SoC, no missing cores */
+			break;
+	}
+#endif
+	return 0;
+}
+DEV_TREE_UPDATE(lk2nd_partial_goods_dt_update);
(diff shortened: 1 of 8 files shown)


lk2nd/partial-goods/partial-goods.c
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/* Copyright (c) 2022, Otto Pflüger <affenull2345@gmail.com> */
+
+#include <platform/partial_goods.h>
+#include <boot.h>
+#include <libfdt.h>
+#include <bits.h>
+#include <reg.h>
+
+static void disable_node(void *dtb, const char *path)
```
TravMurav, line 10: maybe this better to be in util/lkfdt?

stephan-gh:
> Also, I can't think of a proper match rule for the 8110 since the bootloader doesn't pass anything special except for a rather generic-looking panel ID.

The panel(s) are still better than nothing I think. Actually it's probably even good enough as it's quite rare that the same panels are used in multiple devices.

lk2nd/partial-goods/partial-goods.c
affenull2345: I've refactored it a bit and moved the helper function, which is now more generic, to util/lkfdt.

lk2nd/partial-goods/partial-goods.c
```diff
@@ -0,0 +1,41 @@
...
+#include <lk2nd/util/lkfdt.h>
+
+#if TARGET_MSM8909
+static const char *cpu_disable_paths[] = {
+	"/cpus/cpu@2",
+	"/soc@0/power-manager@b0a8000",
+	"/soc@0/power-manager@b0a9000",
+	"/cpus/cpu@3",
+	"/soc@0/power-manager@b0b8000",
+	"/soc@0/power-manager@b0b9000",
```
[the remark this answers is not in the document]
stephan-gh: Also, if we special case the disabling of CPUs it should be quite easy to check the `enum boot_type` for BOOT_DOWNSTREAM and replace the `device_type` like in the original implementation instead. That way it would also work for downstream if there is some msm-4.9 device with MSM8905 out there. This is optional though.

Would you like to try implementing this? I can also try to make these changes if you prefer to focus on other things.

lk2nd/project/base.mk
```diff
@@ -54,4 +54,11 @@ DEFINES := $(filter-out USER_FORCE_RESET_SUPPORT=1, $(DEFINES))
 # Allow flashing independent of battery voltage
 DEFINES := $(filter-out CHECK_BAT_VOLTAGE=1, $(DEFINES))
 
+# lk2nd provides its own mainline-friendly partial-goods implementation
+ifeq ($(ENABLE_PARTIAL_GOODS_SUPPORT), 1)
```
stephan-gh, line 58: Can you make this conditional to `msm8909` for now? We don't handle msm8996 in this implementation yet and I'm unsure on which devices partial-goods would be relevant for msm8996.

affenull2345: This condition disables the downstream-specific partial goods code, which is needed on all platforms in order to avoid problems with that code. The lk2nd partial goods replacement already has an `#if` condition for msm8909 and does nothing on other platforms, so I'm not sure if adding another condition here would improve anything.

stephan-gh: I specifically want it to fail on MSM8996 because if someone reports that we know that these devices exist (rather than failing to do the necessary patching there). I suspect most MSM8996 devices don't have any "partial goods" restriction and in this case the original code does actually nothing and doesn't fail like on MSM8905.

affenull2345:
I'll open a separate pull request for the new partial goods code based on the `experimental-tmp3` branch.

PR #220, opened 2022-12-14
[DNM] [experimental-tmp] Add LG Nexus 5X support

Dominduchami:
Add support for LG Nexus 5X. Also enable autorefresh so that the framebuffer works on linux.

I would greatly appriciate any feedback regarding the framebuffer enabler, the way this is done here feels really hacky.

### DIFF
--- a/lk2nd/device/dts/msm8994/msm8992-lg-bullhead.dts
+++ b/lk2nd/device/dts/msm8994/msm8992-lg-bullhead.dts
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	model = "LG Nexus 5X";
+	qcom,msm-id = <251 0>, <252 0>;
+	qcom,pmic-id = <0x10009 0x1000A 0x0 0x0>;
+	qcom,board-id = <0xb64 0>;
+};
+
+&lk2nd {
+	model = "LG Nexus 5X";
+	compatible = "lg,bullhead";
+};
--- a/lk2nd/display/cont-splash/refresh.c
+++ b/lk2nd/display/cont-splash/refresh.c
@@ -40,6 +40,9 @@ static void mdp_cmd_signal_refresh(void)
 
 static void mdp_cmd_refresh_start(struct fbcon_config *fb)
 {
+#if ENABLE_AUTOREFRESH
+	writel((BIT(31) | 1), MDP_PP_0_BASE + MDSS_MDP_REG_PP_AUTOREFRESH_CONFIG);
+#else
 	thread_t *thr;
 
 	event_init(&refresh_event, false, EVENT_FLAG_AUTOUNSIGNAL);
@@ -53,6 +56,7 @@ static void mdp_cmd_refresh_start(struct fbcon_config *fb)
 
 	thread_resume(thr);
 	fb->update_start = mdp_cmd_signal_refresh;
+#endif
 }
 
 bool mdp_start_refresh(struct fbcon_config *fb)
--- a/lk2nd/device/dts/msm8994/rules.mk
+++ b/lk2nd/device/dts/msm8994/rules.mk
@@ -1,0 +1,5 @@
+# SPDX-License-Identifier: BSD-3-Clause
+LOCAL_DIR := $(GET_LOCAL_DIR)
+
+ADTBS += \
+	$(LOCAL_DIR)/msm8992-lg-bullhead.dtb \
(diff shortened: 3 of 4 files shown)


lk2nd/display/cont-splash/refresh.c
```diff
@@ -40,6 +40,9 @@ static void mdp_cmd_signal_refresh(void)
 
 static void mdp_cmd_refresh_start(struct fbcon_config *fb)
 {
+#if ENABLE_AUTOREFRESH
```
TravMurav, line 43: The reason hw autorefresh wasn't used is because it didn't work on some platforms (i.e. on 8974) so it was decided to use a more "reliable" (even if ugly) solution. See #137 that adds this thread as well as a manual autorefresh kick via fastboot.

lk2nd/device/dts/msm8994/msm8992-lg-bullhead.dts
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	model = "LG Nexus 5X";
+	qcom,msm-id = <251 0>, <252 0>;
```
TravMurav, line 8:
```suggestion
	qcom,msm-id = <QCOM_ID_MSM8992 0>;
```
Let's use the `qcom,ids.h` like all other files around, I also doubt there is a phone with APQ8092 :)

lk2nd/display/cont-splash/refresh.c
stephan-gh: I want to avoid making such changes unconditionally. lk2nd is meant to boot a variety of different boot images (downstream, mainline, lk2nd itself, other operating systems) and enabling autorefresh may break booting them (downstream would be the most likely candidate for that).

However, we could potentially enable autorefresh immediately before booting a Linux that needs this. It could be either detected using a special parameter on the kernel command line (e.g. `lk2nd.display-autorefresh`). Or we could introduce special magic: If `/chosen/simple-framebuffer` exists then enable autorefresh.

Josh4lifee: Appreciate any help at all thx

Dominduchami: Thanks for the feedback!
> I want to avoid making such changes unconditionally. lk2nd is meant to boot a variety of different boot images (downstream, mainline, lk2nd itself, other operating systems) and enabling autorefresh may break booting them (downstream would be the most likely candidate for that).
> 
Ah, I understand, makes sense :-) Downstream probably wouldn't like that.

> However, we could potentially enable autorefresh immediately before booting a Linux that needs this. It could be either detected using a special parameter on the kernel command line (e.g. `lk2nd.display-autorefresh`). Or we could introduce special magic: If `/chosen/simple-framebuffer` exists then enable autorefresh.

To me the second option seems better, in that case adding framebuffer on such devices upstream would be the same as normally.

lk2nd/device/dts/msm8994/msm8992-lg-bullhead.dts
Dominduchami: True :-) Sorry, I should've checked other examples before adding that, will fix it once I get home

lk2nd/display/cont-splash/refresh.c
pevik: But `/chosen/simple-framebuffer` would require mainline accepting this, but kernel command line can be used on vanilla mainline (unmodified kernel sources in case mainline does not want to accept `/chosen/simple-framebuffer`.

TravMurav: FWIW I was looking at provisioning the full framebuffer data from the lk2nd, which could include enabling autorefresh as one of the steps. https://github.com/msm8916-mainline/lk2nd/blob/rebase/fb/lk2nd/display/simplefb.c

Interestingly upstream allows extending the simplefb bindings *just for the bootloader data* so theoretically we could even have custom `lk2nd,simple-framebuffer` compatible that would allow `lk2nd,auto-refresh;` property in mainline

pevik: Interesting, that would be then good try to send a patch :). I suppose we'd have to add our `lk2nd` and it's property into `allOf:` section in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/devicetree/bindings/display/simple-framebuffer.yaml, right? 
BTW what from that file states simplefb bindings can be extended for bootloader data? That example with `pipeline` for `allwinner` and `amlogic`?

TravMurav: Indeed, specifically the last part of the long description says *firmware* may need extra data.

However I think we shouldn't hurry extending it in case lk2nd can handle quirks internally, maybe it's nicer to have the required flags in lk2nd's own dts.

PR #221, opened 2022-12-17
dts: msm8974: Add OnePlus One

z3ntu:
This device needs to use lk2nd-appended-dtb.img, the other one fails with 'dtb not found'.

---

Not sure if we should call this "msm8974" or "msm8974pro". Linux now has both file names and compatibles adjusted for Snapdragon 801 (vs 800) devices to use msm8974pro. Shall we follow with this here?

### DIFF
--- a/app/aboot/aboot.c
+++ b/app/aboot/aboot.c
@@ -639,7 +639,9 @@ unsigned char *update_cmdline(const char *cmdline)
 	bool lk2nd = lk2nd_cmdline_scan(cmdline, "lk2nd");
 
 	/* Only add to cmdline if downstream or lk2nd */
-	if (!lk2nd_cmdline_scan(cmdline, "androidboot.hardware=qcom") && !lk2nd)
+	if (!lk2nd_cmdline_scan(cmdline, "androidboot.hardware=qcom") &&
+	    !lk2nd_cmdline_scan(cmdline, "androidboot.hardware=bacon") &&
+	    !lk2nd)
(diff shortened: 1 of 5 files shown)


dts/msm8974/msm8974-oneplus-bacon.dts
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	qcom,msm-id = <194 0x10000>;
+	qcom,board-id = <0x8 0x0>;
```
stephan-gh, line 10: This looks like `msm8974pro-ac-pm8941-mtp.dts` (8 0 is always standard -mtp from qcom). You'll need the same setup as for FP2 here, with a match property for oneplus-bacon so other devices can be added to the same dts.

z3ntu: fixed

z3ntu:
As mentioned on Matrix, downstream `msm8974pro-ac-pm8941-mtp` also has `<210 0x10000>, <213 0x10000>, <216 0x10000>` for `qcom,msm-id` but with this even appended dtb image fails with `dtb not found` so leaving it out

stephan-gh (approved):

TravMurav (approved):

PR #222, opened 2022-12-17
[experimental-tmp] Add Nokia Lumia msm8x26 range of devices

FieryFlames:
New version of #157 

This PR adds a proper device tree with all Nokia msm8x26 devices. Some of these have variants that come with msm8926 instead of msm8226 (and I think the 830 only comes with msm8926) but I don't think it's worth adding a msm8926-v1-mtp.dts that would be identical.

These devices boot lk2nd differently to other devices, as they use a UEFI bootloader which we break out of using a shim. They need the device tree bundling in with `LK2ND_BUNDLE_DTB=msm8226-v1-mtp.dtb` and the compatible setting with `LK2ND_COMPATIBLE=<device compatible>`. They also need a different MEMBASE, `0x01D00000` or `0x07300000` is what have worked for me.

Last PR was talking about putting some build instructions in README and having a comment about these devices in the dts. Not super sure what to put in the dts, what I put previously apparently had weird format?


lk2nd/device/dts/msm8226/msm8226-v1-mtp.dts
```diff
@@ -0,0 +1,36 @@
...
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8226 0>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0>;
+};
+
+&lk2nd {
+	nokia-dempsey {
+		model = "Microsoft Lumia 640";
+		compatible = "nokia,dempsey";
```
wonderfulShrineMaidenOfParadise, line 14: There is no matching property and it won't distinguish from these devices.

wonderfulShrineMaidenOfParadise:
When `LK2ND_COMPATIBLE=<device compatible>` is required, it might need some documentations or comments in the files.

PR #225, opened 2022-12-28
Can someone share the lk2nd image file for msm8953

generouspanda:
I've been having trouble trying to compile it for my moto g7 power.


devops117:
https://github.com/msm8953-mainline/lk2nd#building
`$ make TOOLCHAIN_PREFIX=arm-none-eabi- msm8953-secondary`

?

wonderfulShrineMaidenOfParadise:
Well, wrong repo. You are looking for https://github.com/msm8953-mainline/lk2nd

generouspanda:
I am not very savvy with this and am a newbie and there aren't very many instructions. I tried running it but got all kinds of errors and I'm not sure what I'm doing wrong. It would be easier for me if someone was able to share it please. I'm just doing this so I can install postmarketOS on my moto g7 power because I prefer linux based OSes.

PR #237, opened 2023-02-20
dts: msm8916: Add support for 4G UFI stick (UFI003_MB_V02)

MarvelousBlack:
UFI003_MB_V02 is a tiny 4G-WiFi-USB-Router based on msm8916.

Specification:

- System-On-Chip: msm8916
- Flash size: 4 GiB
- RAM: 512 MiB
- 1x male USB 2.0 port
- An EDL button on board

Reference:
https://forum.openwrt.org/t/uf896-qualcomm-msm8916-lte-router-384mib-ram-2-4gib-flash-android-openwrt/131712 https://hackaday.com/2022/08/03/hackable-20-modem-combines-lte-and-pi-zero-w2-power/

### DIFF
--- a/dts/msm8916/msm8916-unknow-ufi003mb.dts
+++ b/dts/msm8916/msm8916-unknow-ufi003mb.dts
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+ 	model= "Unbranded 4G UFI stick (UFI003_MB_V02)";
+	compatible = "unknow,ufi003-mb", "qcom,msm8916", "lk2nd,device";
+ 	qcom,board-id = <0x08 0x100>;
+ 	qcom,msm-id = <0xce 0x00 0xf8 0x00 0xf9 0x00 0xfa 0x00>;
+	//Register EDL button as HOME button in order to use the only button on the board into fastboot
+ 	lk2nd,keys = <KEY_HOME 37 (GPIO_PULL_DOWN | GPIO_ACTIVE_HIGH)>;
+};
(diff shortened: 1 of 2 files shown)


dts/msm8916/msm8916-unknow-ufi003mb.dts
```diff
@@ -0,0 +1,15 @@
...
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+ 	model= "Unbranded 4G UFI stick (UFI003_MB_V02)";
+	compatible = "unknow,ufi003-mb", "qcom,msm8916", "lk2nd,device";
+ 	qcom,board-id = <0x08 0x100>;
```
TravMurav, line 11: This is "msm8916-512mb-mtp" and it looks like there is an "ufi-001b/ufi-001c" board already that seems exactly compatible, given you declare the same button.

Are you sure you *have to* add a new board and can't use lk2nd as-is?

MarvelousBlack:
Oh, the msm8916-512mb-mtp can exactly compatible for UFI003_MB_V02. lk2nd can work fine without this DT. I check the gpio and other things, UFI003_MB_V02 seen like a variant of ufi-001c, fully compat with ufi-001c's DT. It can closed, thanks for you reply.

185264646:
> Oh, the msm8916-512mb-mtp can exactly compatible for UFI003_MB_V02. lk2nd can work fine without this DT. I check the gpio and other things, UFI003_MB_V02 seen like a variant of ufi-001c, fully compat with ufi-001c's DT. It can closed, thanks for you reply.

My mistake, I forgot to mention that in comment. I thought it has been widely-known that ufi001c(b), ufi003, uf103s are just very compatible.

dts/msm8916/msm8916-unknow-ufi003mb.dts
185264646: > This is "msm8916-512mb-mtp" and it looks like there is an "ufi-001b/ufi-001c" board already that seems exactly compatible, given you declare the same button.
> 
> Are you sure you _have to_ add a new board and can't use lk2nd as-is?
> 
> (...) Thinking of this more, it might still be worth defining this in the same file for lk1st, if the board needs a different DT in Linux than that one.

No, we don't need to. As I said below. The 4 models are very similar to each other. The board design is almost identical. They can share one dts. I got two of them and have seen many photos of the other two. Actually, the only thing to notice is the modem firmware, which can be resolved by msm-firmware-loader or droid-juicer, that has nothing to do with linux kernel or lk2nd/lk1st.

PR #240, opened 2023-02-28
[Patch] Add support for another LTE dongle uz801-v3.0

185264646:
Also added lk2nd,match-cmdline for UFI-001C.

### DIFF
--- a/dts/msm8916/msm8916-512mb-mtp.dts
+++ b/dts/msm8916/msm8916-512mb-mtp.dts
@@ -17,5 +17,6 @@
 		/* Register EDL button as HOME button */
 		/* This is the only button on this device */
 		lk2nd,keys = <KEY_HOME 37 (GPIO_PULL_DOWN | GPIO_ACTIVE_HIGH)>;
+		lk2nd,match-cmdline = "* mdss_mdp.panel=1:spi:0:qcom,mdss_spi_st7735s_128128_cmd";
 	};
 };
--- a/dts/msm8916/msm8916-512mb-mtp.dts
+++ b/dts/msm8916/msm8916-512mb-mtp.dts
@@ -20,6 +20,12 @@
 		lk2nd,match-cmdline = "* mdss_mdp.panel=1:spi:0:qcom,mdss_spi_st7735s_128128_cmd";
 	};
 
+	/*
+	 * Note:
+	 * Due to some unknown reasons, the stock aboot firmware is incompatible
+	 * with qhypstub and tz firmware from db410c, use lk1st if possible.
+	 */
+
 	zhihe-uz801-v3 {
 		model = "uz801 v3.0 4G Modem Stick";
 		compatible = "zhihe,uz801-v3", "qcom,msm8916", "lk2nd,device";
--- a/dts/msm8916/msm8916-512mb-mtp.dts
+++ b/dts/msm8916/msm8916-512mb-mtp.dts
@@ -19,4 +19,14 @@
 		lk2nd,keys = <KEY_HOME 37 (GPIO_PULL_DOWN | GPIO_ACTIVE_HIGH)>;
 		lk2nd,match-cmdline = "* mdss_mdp.panel=1:spi:0:qcom,mdss_spi_st7735s_128128_cmd";
 	};
+
+	zhihe-uz801-v3 {
+		model = "uz801 v3.0 4G Modem Stick";
+		compatible = "zhihe,uz801-v3", "qcom,msm8916", "lk2nd,device";
+
+		/* Register the reset button as HOME button */
+		/* This is the only button on this device */
+		lk2nd,keys = <KEY_HOME 23 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>;
+		lk2nd,match-cmdline = "* mdss_mdp.panel=1:dsi:0:qcom,mdss_dsi_jdi_1080p_video";
+	};
 };
--- a/dts/msm8916/msm8916-mtp.dts
+++ b/dts/msm8916/msm8916-mtp.dts
@@ -7,6 +7,10 @@
 / {
 	compatible = "qcom,msm8916-mtp", "qcom,msm8916";
 
+	// The bootloader of UZ801 requires a dts with this board-id to boot
+	qcom,msm-id = <206 0>;
+	qcom,board-id = <8 0>;
+
 	// There are devices that identify themselves as qcom,msm8916-mtp,
 	// even when they are really not (at least physically).
 	// Therefore we need other things we can match on (e.g. the cmdline).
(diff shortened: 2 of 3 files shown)


dts/msm8916/msm8916-512mb-mtp.dts
```diff
@@ -17,5 +17,16 @@
...
+	};
+
+	zhihe-uz801-v3 {
+		model = "uz801 v3.0 4G Modem Stick";
+		compatible = "zhihe,uz801-v3", "qcom,msm8916", "lk2nd,device";
+
+		/* Register the reset button as HOME button */
+		/* This is the only button on this device */
+		lk2nd,keys = <KEY_HOME 23 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>;
+		lk2nd,match-cmdline = "* mdss_mdp.panel=1:dsi:0:qcom,mdss_dsi_jdi_1080p_video";
```
wonderfulShrineMaidenOfParadise, line 30: A panel?

dts/msm8916/msm8916-mtp.dts
```diff
@@ -3,9 +3,12 @@
 /dts-v1/;
 
 #include <skeleton.dtsi>
+#include <lk2nd.h>
 
 / {
 	compatible = "qcom,msm8916-mtp", "qcom,msm8916";
+	qcom,msm-id = <206 0>;
+	qcom,board-id = <8 0>;
```
wonderfulShrineMaidenOfParadise, line 11: It doesn't look right to change this file, if it's not an 1gb modem (`qcom,board-id = <8 0>;`).
Check your downstream device tree to confirm.

dts/msm8916/msm8916-512mb-mtp.dts
185264646: > A panel?

No, no panel. The bootloader simply passes this cmdline to kernel.

dts/msm8916/msm8916-512mb-mtp.dts
```diff
@@ -17,5 +17,22 @@
...
 		/* This is the only button on this device */
 		lk2nd,keys = <KEY_HOME 37 (GPIO_PULL_DOWN | GPIO_ACTIVE_HIGH)>;
+		lk2nd,match-cmdline = "* mdss_mdp.panel=1:spi:0:qcom,mdss_spi_st7735s_128128_cmd";
+	};
+
+	/*
+	 * Note:
+	 * Due to some unknown reasons, the stock aboot firmware is incompatible
+	 * with qhypstub and tz firmware from db410c, use lk1st if possible.
+	 */
```
stephan-gh, line 27: Is it really worth supporting lk2nd on those at all? They'd probably be much better suited with lk1st.

185264646: > Is it really worth supporting lk2nd on those at all? They'd probably be much better suited with lk1st.

I don't really want to support lk2nd. lk1st is confirmed to work quite well on almost every msm8916-based dongles. On the contrary, lk2nd is often broken due to quirky stock bootloader provided by vendors. Even sbl1 can be ported from db410c to avoid unnecessary problems. However, since the project is named after "lk2nd", is it acceptable to drop the support of lk2nd?

stephan-gh: It would be nice if you could make sure that booting lk2nd from lk1st (rather than the stock bootloader) works. This should still allow dropping most of the necessary quirks.

185264646: > It would be nice if you could make sure that booting lk2nd from lk1st (rather than the stock bootloader) works. This should still allow dropping most of the necessary quirks.

So qhypstub + tz from db410c + lk1st + lk2nd? Probably works. I remembered testing it before.

> It would be nice if you could make sure that booting lk2nd from lk1st (rather than the stock bootloader) works. This should still allow dropping most of the necessary quirks.

> It would be nice if you could make sure that booting lk2nd from lk1st (rather than the stock bootloader) works. This should still allow dropping most of the necessary quirks.

So qhypstub + tz from db410c + lk1st + lk2nd? Probably works. I remembered testing it before.

stephan-gh: Yes, exactly.

dts/msm8916/msm8916-mtp.dts
```diff
@@ -7,6 +7,10 @@
 / {
 	compatible = "qcom,msm8916-mtp", "qcom,msm8916";
 
+	// The bootloader of UZ801 requires a dts with this borad-id to boot
```
TravMurav, line 10: Do we want to keep this given the stock aboot seems to be dubious anyway? Maybe just make lk1st mandatory for the board and drop this?

dts/msm8916/msm8916-512mb-mtp.dts
185264646: > Hello! I can confirm that lk2nd can be booted from lk1st on my uz801 with db410c firmware.
> 
> ```
> $ fastboot getvar product
> product: LK2ND_MSM8916
> ```

I've confirmed again mine is not able to work with qhypstub+tz firmware from db410c.
To ensure we are talking about the same tz firmware, please look up the size of your tz firmware, it should be around ~400 KB, rather than the tz firmware from openstick, which is about 600KB.

185264646: > Hello! I can confirm that lk2nd can be booted from lk1st on my uz801 with db410c firmware.
> 
> ```
> $ fastboot getvar product
> product: LK2ND_MSM8916
> ```

Please try again with this link: https://releases.linaro.org/96boards/dragonboard410c/linaro/rescue/17.09/dragonboard410c_bootloader_emmc_android-88.zip
The tz firmware from here is smaller than 512kb, can be flashed directly to tz partition, and should be the suggested tz firmware rather than the larger one you may encounter elsewhere.

iovxw: > Please try again with this link: https://releases.linaro.org/96boards/dragonboard410c/linaro/rescue/17.09/dragonboard410c_bootloader_emmc_android-88.zip The tz firmware extracted from it is smaller than 512kb, can be flashed directly to tz partition, and should be the suggested tz firmware rather than the larger one you may encounter elsewhere.

Ah, my mistake, I'm replying to stephan-gh, the lk1st here is not the stock aboot, it is compiled from your branch

And this is how I flashed the firmware (without the modifications to partition.txt): https://github.com/iovxw/device-thwc-ufi001c/tree/master/partitions

dts/msm8916/msm8916-mtp.dts
iovxw: Please keep it, it's still useful for backing up stock partitions without entering QDL (by temporarily booting to lk2nd)

TravMurav:
Sorry, lost this for a while, thanks for the pr!

PR #245, opened 2023-04-06
target: msm8916: meminfo: reduce max download size to 192 MiB

### LINKED ISSUE #239: [Bug] Crash on low-memory devices.
For LTE dongles with 512MB memory, `fastboot flash system`(or userdata) would crash the device, an additional argument like `-S 200M` is required to flash a large partition.

185264646:
256 MiB is too big for some low-memory devices and cause crashes.
Tested on a LTE dongle.

Resolve #239

### DIFF
--- a/target/msm8916/meminfo.c
+++ b/target/msm8916/meminfo.c
@@ -81,5 +81,6 @@ void *target_get_scratch_address(void)
 
 unsigned target_get_max_flash_size(void)
 {
-	return (256 * 1024 * 1024);
+	// 256 MiB breaks some low-memory devices, 192 MiB should be sufficient.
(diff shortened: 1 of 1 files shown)


minlexx:
Could this parameter be optional, or configurable per board? As some variable in some makefile I assume..

185264646:
> Could this parameter be optional, or configurable per board? As some variable in some makefile I assume..

lk2nd is designed to be generic across different devices I think. This patch is simple and does not harm flashing speed too much anyway.

TravMurav:
This will need some more thought for -next, maybe we could do this dynamically by checking the board ram size.

Should be fine for now to fix the "oom" on 512MiB devices, Thanks!

PR #247, opened 2023-04-20
dts: msm8916: Add Galaxy A5U CAN (SM-A500W)

RDS5:
Add dts support for galaxy a5u canbmc (a500w)
its just a5u with different firmware, picks strange dts tho.
[downstream dts](https://github.com/Galaxy-MSM8916/android_kernel_samsung_msm8916/blob/lineage-17.1/arch/arm/boot/dts/samsung/msm8916/msm8916-sec-a5u-canbmc-r01.dts)

### DIFF
--- a/dts/msm8916/msm8916-samsung-r11.dts
+++ b/dts/msm8916/msm8916-samsung-r11.dts
@@ -16,6 +16,13 @@
 		lk2nd,match-bootloader = "A500YZ*";
 		lk2nd,smd-rpm-hack-opening;
 	};
+	
+	a5u-canbmc {
+		// In downstream it is called Samsung A5U CAN PROJECT Rev01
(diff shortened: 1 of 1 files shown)


RDS5:
> Thanks! Can you make sure you can boot downstream kernel via lk2nd? As you say it picks a dubious id not in downstream, maybe you might need to add an override for lk2nd to correct it when it boots... See i.e. https://github.com/msm8916-mainline/lk2nd/blob/master/dts/msm8916/msm8916-samsung-r04.dts#L40-L41

that was the case, i have tested [these new changes](https://github.com/msm8916-mainline/lk2nd/pull/247/commits/d246be03e449b944b1a0668d5ea830c8568e4a1d) and they boot downstream fine (tested on latest canbmc firmware).

PR #248, opened 2023-04-22
dts: apq8016: Fix downstream booting (gtelwifiue)

RDS5:
lk2nd used to boot the wrong dtb and downstream wouldnt work
tested with lineageos19.1 on canadian gtelwifiue (although afaik theres no differences between the firmwares bootloaders)


TravMurav:
Thanks!

PR #250, opened 2023-04-30
README: add details about lg-lenok

z3ntu:


TravMurav:
Maybe we could just move it to the top? Could you check if it doesn't break other smartwatches iirc you had a bunch?)

lenok bootloader is pretty funny but I'd hope this is a lenok specific mess...

i.e. the fact that it actually parses the compatible array....

https://github.com/msm8916-mainline/lk2nd/blob/rebase/mp-menu/lk2nd/device/dts/msm8226/lenok.dts#L6-L15

z3ntu:
Yes last I checked the other watches booted fine with the same image, lenok being on top.

Does the PR look okay now?

TravMurav:
Thanks! Hope we will not have way too many of those messy bootloaders....

PR #256, opened 2023-05-13
dts: msm8974: Fix matching OnePlus One

z3ntu:
The previous match-cmdline wasn't correct since the bootloader actually doesn't set androidboot.hardware=bacon in the cmdline. Let's match with the panel instead.

### DIFF
--- a/dts/msm8974/msm8974pro-ac-pm8941-mtp.dts
+++ b/dts/msm8974/msm8974pro-ac-pm8941-mtp.dts
@@ -16,7 +16,7 @@
 	oneplus-bacon {
 		model = "OnePlus One";
 		compatible = "oneplus,bacon", "qcom,msm8974", "lk2nd,device";
-		lk2nd,match-cmdline = "* androidboot.hardware=bacon *";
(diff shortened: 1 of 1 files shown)


dts/msm8974/msm8974pro-ac-pm8941-mtp.dts
```diff
@@ -16,7 +16,7 @@
 	oneplus-bacon {
 		model = "OnePlus One";
 		compatible = "oneplus,bacon", "qcom,msm8974", "lk2nd,device";
-		lk2nd,match-cmdline = "* androidboot.hardware=bacon *";
+		lk2nd,match-cmdline = "* mdss_mdp.panel=1:dsi:0:qcom,mdss_dsi_jdi_1080p_cmd *";
```
TravMurav, line 19: Is there only one panel known to be used in the field? If no, maybe like this?

https://github.com/msm8916-mainline/lk2nd/blob/master/dts/msm8916/msm8916-qrd-skut1.dts#L15

z3ntu: To my knowledge, yes. If we find more in the future I'll expand this for sure (and then we also need to use the panel selection of lk2nd + kernel support for the new panel)

TravMurav:
Thanks!

PR #257, opened 2023-05-31
dts: Add Samssung A5 (Duos) (SM-A5000)

exkc:
aa

it is likely an china variant or a hk variant of a5 because there are seem to be only hk and cn version of stock firmw
      are for it. 
[build.prop.txt](https://github.com/msm8916-mainline/lk2nd/files/11614522/build.prop.txt)
There are 2 code name beibg found in the /system/build.prop (a5ltechn and a5ltezc) of the stock rom 

a5ltechn should be the code name of it bc the hostname of the stock rom is a5ltechn


TravMurav:
Hm, I remember someone already having the a5000, not sure why it wasn't there already...

Thanks for adding it!

PR #271, opened 2023-07-02
dts: msm8926: add support for Coolpad 8730L

185264646:
8730l

Tried with downstream kernel, vol-down gets into fastboot mode.

~TODO: add gpio definition for HOME button.~

### DIFF
--- a/dts/msm8226/msm8926-v2-720p-mtp.dts
+++ b/dts/msm8226/msm8926-v2-720p-mtp.dts
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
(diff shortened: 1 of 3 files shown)


dts/msm8226/msm8926-v2-720p-mtp.dts
```diff
@@ -0,0 +1,18 @@
...
+/ {
+	compatible = "qcom,msm8926-mtp", "qcom,msm8926";
+
+	qcom,msm-id = <200 0>;
+	qcom,board-id = <8 0>;
+
+	coolpad-8730l {
+		model = "Coolpad 8730L";
+		compatible = "coolpad,8730l", "qcom,msm8926", "lk2nd,device";
+	};
```
TravMurav, line 17: Hm, maybe we should add some match rule here, in case some other 8926-mtp device appears later? i.e. match-cmdline or match-panel + a display panel list node

TravMurav:
Thanks!

PR #272, opened 2023-07-04
dts: msm8916: add support for Asus ZE500KL

doomvrr:
adapted from Asus z00l, and changed the 2nd panel to one that my unit has, as first panel from z00l is found on z00e aswell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 - ARK Benefit A3 - peach
 - Asus Zenfone 2 Laser (720p) - Z00L
 - Asus Zenfone 2 Laser (1080p) - Z00T
+- Asus Zenfone 2 Laser ZE500KL - Z00E
 - Asus Zenfone Max ZC550KL (2016) - Z010D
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -51,6 +51,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 - ARK Benefit A3 - peach
 - Asus Zenfone 2 Laser (720p) - Z00L
 - Asus Zenfone 2 Laser (1080p) - Z00T
+- Asus Zenfone 2 Laser ZE500KL - Z00E
```
TravMurav, line 54: is this device any different from the other 720p one? i.e. does it have some distinctive marketing, or maybe a different variant of the same thing?
If it's "same marketing, different region" than i.e. z00l, maybe this?
```suggestion
- Asus Zenfone 2 Laser (720p) - Z00L, Z00E
- Asus Zenfone 2 Laser (1080p) - Z00T
```

doomvrr: size is different, z00l is 5.5 inch, whereas z00e is 5 inch screen size.

TravMurav:
Thanks!

PR #280, opened 2023-08-01
Huawei GX8 (RIO-L01)

MarkMuth:
Adds support for the Huawei G8/GX8. Tested with (RIO-L01).
Removed some board-id entries from `kiwi` that are now conflicting with `rio` .
For reference I mentioned the files in Huawei's downstream kernel sources containing those IDs.

### DIFF
--- a/dts/msm8916/msm8939-huawei-kiwi.dts
+++ b/dts/msm8916/msm8939-huawei-kiwi.dts
@@ -7,12 +7,27 @@
 / {
 	qcom,msm-id = <239 0>;
 	qcom,board-id =
-		<8001 4>, <8007 4>, <8017 4>, <8019 4>,
-		<8023 4>, <8033 4>, <8039 4>, <8049 4>,
-		<8051 4>, <8054 4>, <8065 4>, <8067 4>,
-		<8070 4>, <8081 4>, <8083 4>, <8086 4>,
-		<8097 4>, <8099 4>, <8103 4>, <8115 4>,
-		<8116 4>, <8119 4>, <8131 4>;
+		<8001 4>, // huawei_kiw_al10_va.dts
+		<8003 4>, // huawei_kiw_al10_vb.dts
+		<8007 4>, // huawei_kiw_cl00_vc.dts
+		<8017 4>, // huawei_kiw_ul00_va.dts
+		<8019 4>, // huawei_kiw_ul00_vb.dts
+		<8023 4>, // huawei_kiw_tl00h_vc.dts
+		<8033 4>, // huawei_kiw_cl00_va.dts
+		<8035 4>, // huawei_kiw_cl00_vb.dts
+		<8039 4>, // huawei_kiw_l21_vc.dts
+		<8049 4>, // huawei_kiw_tl00h_va.dts
+		<8051 4>, // huawei_kiw_tl00h_vb.dts
+		<8065 4>, // huawei_kiw_l21_va.dts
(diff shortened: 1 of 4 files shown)


dts/msm8916/msm8939-huawei-rio.dts
```diff
@@ -0,0 +1,22 @@
...
+		<8054 0>, <8054 4>, // huawei_rio_cl00_va.dts
+		<8055 0>, <8055 4>, // huawei_rio_l02_va.dts
+		<8070 0>, <8070 4>, // huawei_rio_al00_va.dts
+		<8071 0>, <8071 4>, // huawei_rio_al00_vb.dts
+		<8086 0>, <8086 4>, // huawei_rio_l01_va.dts
+		<8087 0>, <8087 4>, // huawei_rio_l01_vb.dts
+		<8102 0>, <8102 4>, // huawei_rio_tl00_va.dts
+		<8134 0>, <8134 4>, // huawei_rio_cl00_vb.dts
+		<8135 0>, <8135 4>, // huawei_rio_cl00_vc.dts
+		<8118 0>, <8118 4>; // huawei_rio_l03_va.dts
```
TravMurav, line 19: I believe in #49 we decided to omit the <xxxx 0> devices. Can you check if you need 0 for your device?

dts/msm8916/msm8939-huawei-kiwi.dts
```diff
@@ -7,12 +7,27 @@
...
+		<8065 4>, // huawei_kiw_l21_va.dts
+		<8067 4>, // huawei_kiw_l21_vb.dts
+		<8081 4>, // huawei_kiw_l22_va.dts
+		<8083 4>, // huawei_kiw_l22_vb.dts
+		<8097 4>, // huawei_kiw_l23_va.dts
+		<8099 4>, // huawei_kiw_l23_vb.dts
+		<8103 4>, // huawei_kiw_l22_vc.dts
+		<8115 4>, // huawei_kiw_al10_vc.dts
+		<8119 4>, // huawei_kiw_l23_vc.dts
+		<8131 4>; // huawei_kiw_ul00_vc.dts
```
TravMurav, line 30: Can you confirm that none of the removed IDs correspond to both devices? i.e. if there is multiple downstream dts for different devices with the same ID? I'm not sure how the original list was collected, but if they overlap, we would probably need to merge these files and use different selection mechanism to make sure all devices are still detected properly.

dts/msm8916/msm8939-huawei-rio.dts
```diff
@@ -0,0 +1,22 @@
...
+	model = "Huawei RIO";
```
TravMurav, line 20:
```suggestion

	model = "Huawei RIO";
```
Rio seems to be the codename? Could you please put marketing name(s) here, like with kiwi.

Also would be nice to add it to the readme, like other device introductions do

dts/msm8916/msm8939-huawei-kiwi.dts
MarkMuth: I have only kernel source for RIO-L01 that was once available in Huawei's customer support site. In there, every ID is used only once. Maybe someone with access to kiwi hardware and source can check as well?

TravMurav: I suppose if both rio and kiwi in the same tree and they don't collide, it's a reasonable assumption to make there is no overlap, thanks for making sure!

TravMurav:
Thanks!

PR #283, opened 2023-08-06
dts: msm8916: samsung-j5(x): Refactor

wonderfulShrineMaidenOfParadise:


TravMurav:
Thanks for this cleanup!

Note that long term we are planning to switch to a different lk2nd branch with a different DT "format" so please keep that in mind starting any significant changes or refactors since all DTs would have to be converted later (but only when we freeze the new bindings so we don't have to re-do everything multiple times :)

PR #286, opened 2023-08-17
dts: msm8916: qrd/7+12-v1: Also match qrd/7-v1.1

wonderfulShrineMaidenOfParadise:
The stock Kitkat bootloader selects qrd/7-v1.1.

Cc: @TravMurav


TravMurav:
Thanks!

PR #288, opened 2023-08-23
README.md: add preview

wonderfulShrineMaidenOfParadise:
(c) 2022 Nikita Travkin CC BY-SA 4.0

Cc: @TravMurav


stephan-gh:
That picture is huge. :D Can get it down to 852K with zopflipng but as a photo it would probably be better saved as a JPEG.

wonderfulShrineMaidenOfParadise:
With JPG it should be much smaller now.

README.md
```diff
@@ -1,4 +1,7 @@
 # lk2nd
+
+![](FOSDEM22_msm8916_lk2nd.png)
```
TravMurav, line 3:
```suggestion
![](FOSDEM22_msm8916_lk2nd.jpg)
```

:)

but tbh I'd consider if we should hide that file somewhere into `.github/` or some dedicated `.assets/` along with the license file, so it doesn't catch the eye too much. Also can probably drop "msm8916" since there are a couple of 8226s on the photo too.

FOSDEM22_msm8916_lk2nd.jpg
TravMurav, line 1: I'm concerned that this photo takes a bit too much space as-is:

изображение

Any chance you can cut it in half or more, to make a "thin" banner instead?
Here is the original jpeg so you don't have to transcode the photo too many times: 
[IMG_20220115_152237_wS6wtXeb.jpg](https://github.com/msm8916-mainline/lk2nd/assets/22081178/c3a3cf00-327f-485c-bbcb-bab968d45175)

You can probably resize it to about 900px horizontal resolution too, it would still fit 1:1 on a 4k screen then

TravMurav:
Thanks!

PR #293, opened 2023-09-04
[experimental-tmp2] dts: msm8226: Add HTC One Mini 2

z3ntu:
We need to use a different scratch region since the one typically used is reserved for something else and just returns zeroes when read.
```
memul:/ # cat /sys/kernel/debug/memblock/memory
   0: 0x00000000..0x05afffff
   1: 0x05d00000..0x074fffff
   2: 0x11600000..0x3f9fffff
memul:/ # cat /sys/kernel/debug/memblock/reserved
   0: 0x00000000..0x00007fff
   1: 0x00100000..0x01522f1b
   2: 0x01e00000..0x01e2e401
   3: 0x02008000..0x02160d47
   4: 0x05b00000..0x05bcffff
   5: 0x2b400000..0x2fffffff
   6: 0x302cb000..0x302fefff
   7: 0x302ffda0..0x302fffff
```


lk2nd/device/dts/msm8226/rules.mk
```diff
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: BSD-3-Clause
 LOCAL_DIR := $(GET_LOCAL_DIR)
 
-DTBS += \
```
stephan-gh, line 4: I guess removing this wasn't intended?

z3ntu: I don't recall exactly why I removed it, but can I have both DTBS and this QCDTBS in the makefile?

stephan-gh: Yes.

z3ntu: Appears to work, thanks! :)

z3ntu:
@stephan-gh Do you think I can recommend people on the pmOS wiki page for this device to pull this branch and build it to use on that phone?

Or shall I spend some effort porting this device to the main branch so it's more easily part of the next release there? Not sure what we can do about this different scratch region in any case...

stephan-gh:
I don't know how to handle the difference in the scratch region properly yet. Perhaps it's easiest if we only merge the DT as first step and recommend people to override the region when building like

```
$ make TOOLCHAIN_PREFIX=... SCRATCH_ADDR=0x20000000 SCRATCH_SIZE=0x10000000 lk2nd-msm8916
```

I think this should work on both branches?

z3ntu:
Thanks for the help! At least in main branch this appears to work well! See #300 (incl. the comment in the dts file). Still keeping this open for experimental-tmp2 branch :) Do you plan on merging PRs with dts into this branch or not? If yes I can also make this branch ready with the SCRATCH_ADDR changes like in master.

TravMurav:
The branches were switched and memul dts is included already with the note. Thanks!

PR #297, opened 2023-09-10
[experimental-tmp3] Add partial goods support

affenull2345:
Add code to disable unavailable CPU cores on MSM8905 and disable Qualcomm's original code that does this because it is broken on mainline.

Suggested changes from #214 applied.

### DIFF
--- a/lk2nd/partial-goods/partial-goods.c
+++ b/lk2nd/partial-goods/partial-goods.c
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/* Copyright (c) 2022-2023, Otto Pflüger <affenull2345@gmail.com> */
+
+#include <platform/partial_goods.h>
+#include <boot.h>
+#include <libfdt.h>
+#include <lk2nd/util/lkfdt.h>
+
+#if !TARGET_MSM8909
+#error partial-goods not supported for this platform
+#endif
+
+static void cpu_disable(void *dtb, const char *path, bool is_downstream)
+{
+	int tmp, offset = fdt_path_offset(dtb, path);
+
+	if (offset < 0)
+		return;
+
+	if (is_downstream) {
+		fdt_setprop_string(dtb, offset, "device_type", "nak");
+	} else {
+		fdt_setprop_string(dtb, offset, "status", "fail");
+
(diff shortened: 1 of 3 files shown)


lk2nd/include/lk2nd/util/lkfdt.h
```diff
@@ -59,4 +59,37 @@ int lkfdt_get_reg(const void *fdt, int parent, int node, uint32_t *addr, uint32_
...
+ *
+ * In a static array of paths, the LKFDT_PATH macro can be used to construct
+ * these lists:
+ *
+ * LKFDT_PATH("/some/node") points to the node at "/some/node".
+ *
+ * LKFDT_PATH("/some/node", "reference") points to the node referenced by
+ * the "reference" property of "/some/node".
+ *
+ * LKFDT_PATH("/some/node", "reference", "other-handle") points to
```
stephan-gh, line 74: Can you think of a use case for having these multiple "recursions" of lookups?

Maybe a slightly more efficient representation would be something like

`(<path>, <additional-phandles to disable>)`

i.e

```c
static lkfdt_path cpu_disable_paths[] = {
	SOMETHING("/cpus/cpu@2", "qcom,acc", "qcom,saw"),
	SOMETHING("/cpus/cpu@3", "qcom,acc", "qcom,saw"),
	NULL,
};

static lkfdt_path cpu_disable_downstream_paths[] = {
	SOMETHING("/cpus/cpu@2"),
	SOMETHING("/cpus/cpu@3"),
	NULL,
};
```

From a performance perspective it would be slightly preferable to do `fdt_path_offset("/cpus/cpu@2")` only once (and then do `lkfdt_lookup_phandle()` on it twice) as traversing the DT tree is a potential expensive operation. We don't need to overoptimize prematurely but finding the right abstractions is still important.

Perhaps as an additional "data point" I think we may have a need to disable based on compatibles later, e.g. consider the 8996 modem/gpu/venus/adsp disabling https://github.com/msm8916-mainline/lk2nd/blob/4990a47005b3b209d7e478b65abfc05aa6cc2b7a/platform/msm8996/include/platform/partial_goods.h#L43-L46

Might be worth thinking about how a compatible lookup (rather than a path lookup) could be represented. But this is not a requirement right now so feel free to postpone that part for the future.

stephan-gh (commented):
Thanks for the quick changes!

PR #300, opened 2023-09-15
dts: msm8226: Add HTC One Mini 2 (memul)

z3ntu:
In the current code the display is not drawn correctly, this is fixed through PR #299.


TravMurav:
Thanks!

PR #302, opened 2023-09-19
dts: Add support for Samsung Tab A 8.0 2014

Susurrus:
Very similar to the existing SM-T530 support, but went with a separate dts file for clarity.

Couldn't find documentation on how the `-r*` suffixes are chosen for `dts` files, so I just incremented it for Samsung. Let me know if it should be something else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ page on the EFIDroid wiki for an exact mapping of LK targets to SoCs.
 - Motorola Moto G 2013 - falcon
 - Samsung Galaxy Grand 2 - SM-G7102
 - Samsung Galaxy Tab 4 10.1 (2014) - SM-T530, SM-T535
+- Samsung Galaxy Tab 4 8.0 (2014) - SM-T330, SM-T330NU
 
 ## Installation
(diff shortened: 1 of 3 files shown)


dts/msm8226/apq8026-samsung-r04.dts
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <0xC708FF01 4 0x20000>;
```
TravMurav, line 10: did you arbitrarily pick 4 here or the device doesn't boot with any other value?

the -rXX in the name reflects the second cell in this property (4). Usually you would pick the highest number available in the downstream sources or the dtbs dumped from the device itself.

dts/msm8226/apq8026-samsung-r04.dts
```diff
@@ -0,0 +1,22 @@
...
+
+	matissewifi {
```
TravMurav, line 12:
```suggestion
	milletwifi {
```

TravMurav:
Thanks!

PR #311, opened 2023-10-05
x86: include/ops.h: Fix a typo in include guard

pevik:
Similar to fix c35069a8e ("arm: include/ops.h: Fix a typo").

Although lk2nd does not support QEMU, thus this header is unsupported, it's better to fix it.


minlexx:
Wow

PR #314, opened 2023-11-20
Support for msm8974-based HTC One (M8) 

amessier:
Add initial support for the HTC One (M8), codename `m8`, based on MSM8974.

Three fixes are required before the `m8` can be supported:
- Implement `gpio_status()` for GPIOs on the SOC, as these are used for the volume up/down buttons.
- Fix eMMC initialization for non-AC variants, from looking at the HTC kernel vendor code, only the Pro-AC variant supports HS400.
- Fix detection of the Pro-AC variant, needed for the eMMC fix.

Then the device tree is added, using the `htc,project-id` values of the vendor kernel (which are used by the HTC bootloader) as the `qcom,msm-id` values, which makes it possible to use the common build for the `m8`. Panel name match is used to detect if it lk2nd booted on a `m8`.

I was able to boot a [near-mainline kernel](https://github.com/msm8974-mainline/linux) using PostmarketOS. I cannot confirm it still works now, as I don't have UART access and SSH is broken for `m8` in PostmarketOS right now.

Known issues:
- Flashing `system` partition using `lk2nd` makes the phone shutdown during the flashing phase. Workaround is to use HTC bootloader to flash `system` partition, and `lk2nd` to flash the `boot` partition.
- I'm not sure about the eMMC fix (due to issue described above), and I can't test on other MSM8974 phones, so hopefully it still works for the other phones.


dts/msm8974/msm8974pro-htc-m8.dts
```diff
@@ -0,0 +1,45 @@
...
+		lk2nd,match-panel;
+
+		lk2nd,keys =
+			<KEY_VOLUMEDOWN	27 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>,
+			<KEY_VOLUMEUP	28 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>;
+
+		panel {
+			compatible = "htc,m8-panel";
+
+			qcom,mdss_dsi_glu_sharp_1080p_cmd {};
```
TravMurav, line 35: IMO it would be nice to assign compatibles to those already, even if we don't know if any other are actually used.

platform/msm8974/gpio.c
```diff
@@ -52,6 +52,11 @@ void gpio_set(uint32_t gpio, uint32_t dir)
 	return;
 }
 
+uint32_t gpio_status(uint32_t gpio)
```
TravMurav, line 55: Hm interesting, For the future lk2nd rebase we have implemented a different way to use gpios to avoid things like this.

Any chance you can try to add your device on top of the `experimental-tmp4` branch? The dts model is a bit different, but that branch (and also `rebase/dts`) has some examples.

target/msm8974/init.c
```diff
@@ -237,16 +237,18 @@ static void target_mmc_sdhci_init()
...
+	 * only for emmc (slot 1)
 	 */
-	if (platform_is_8974ac())
+	if (platform_is_8974ac()) {
 		config.max_clk_rate = MMC_CLK_192MHZ;
-	else
+		config.hs400_support = 1;
+	} else {
 		config.max_clk_rate = MMC_CLK_200MHZ;
+		config.hs400_support = 0;
```
TravMurav, line 247: nit: The config is set to 0 initially so this line is unnecessary.

platform/msm8974/platform.c
```diff
@@ -129,8 +129,8 @@ int platform_is_8974Pro()
 	return ret;
 }
 
-/* Check for 8974PRO AC chip */
-int platform_is_8974ac()
+/* Check for 8974 PRO chip */
+int platform_is_8974Pro()
```
TravMurav, line 133: I'm not sure if this change is unsafe as-is or if the original impl was bugged, but this would change how usb_ss phy is inited/reset on pro-not-ac I think..

https://github.com/msm8916-mainline/lk2nd/blob/master/platform/msm_shared/usb30_udc.c#L116-L155

But I'm somewhat inclined to think it was their mistake so this is a reasonable change

TravMurav (commented):
> Flashing system partition using lk2nd makes the phone shutdown during the flashing

Does the device crash when uploading to ram or when writing to the emmc? Perhaps we need to shrink the scratch memory region....

platform/msm8974/gpio.c
amessier: OK I will try to use that branch as a base, will let you know when I get it to work.

minlexx:
plz remind me (or someone else) to test this for regressions on 8974-hammerhead or klte before merging...

amessier:
> > Flashing system partition using lk2nd makes the phone shutdown during the flashing
> 
> Does the device crash when uploading to ram or when writing to the emmc? Perhaps we need to shrink the scratch memory region....

It crashes (either reboot or shutdown) when writing to the eMMC, i.e. the "Writing" part of a `fastboot flash` command.

I did some more investigation, and it crashes when writing a [sparse image](https://github.com/msm8916-mainline/lk2nd/blob/521e011a9ac4ccb48189bb0ac58eeb1d4de50cbb/app/aboot/aboot.c#L3204), but not when writing [regular data](https://github.com/msm8916-mainline/lk2nd/blob/521e011a9ac4ccb48189bb0ac58eeb1d4de50cbb/app/aboot/aboot.c#L3208). I did a test, writing 256MiB of random data to `system` partition, confirmed it actually wrote the correct data by using the `fastboot oem read-partition` command to verify 2 sectors, which they were. A minimal PostmarketOS image (about 60MiB) always crashes.

This explains why it does not crash when writing a `boot.img`, since it would be considered regular data. Anyway, I'll try investigating more if I can.

platform/msm8974/gpio.c
amessier: I tried the `experimental-tmp4` branch, and the new gpio driver works great, no change required.

I did have to rollback to the older (current) version of `mkbootimg`. The new version in that branch did not generate a working image.

In any case, I don't know what is the plan for that new codebase, I can wait for it to be merged and in use, and resend a new PR. I will fix the other comments in that new PR.

TravMurav: We want to eventually switch to the new branch as a primary one, carrying over all the devices we support here.

It's not an issue for dts-only changes but I want to make sure all new code changes are addressed on the new branch, so this PR is "non-trivial" in that regard...

I'll still be happy to accept it here so you have a working bootloader in the meantime, though the changes and crashes are "interesting" here... I suppose all the same changes are necessary there and the crash is still present on the new branch?

amessier:
Sorry for the delay, I found the issue when flashing an image last month. Thanks to [this post](https://lore.kernel.org/lkml/6450f07d-3e96-4b45-b8fa-8e1b61a2459a@linaro.org/T/), I tried to disable the watchdog in lk2nd too, and it works now! I have been using in the past month, flashing multiple images, and I never saw the issue again.

I've also addressed the issue raised in the review. Please let me know if there are other issues!

Finally, I've also added two commits that fixed some issues I was not happy about: the device tree rename, to match what will be used in the kernel, and a small change in the device name in the device tree.

Also let me know what is preferred, fix-up commits, or should I squash the fix-up commits and force push the new changes?

amessier:
> plz remind me (or someone else) to test this for regressions on 8974-hammerhead or klte before merging...

@minlexx if @TravMurav gives the OK then this would be ready to test on the other 8974 devices, if you are available to! Thanks!

dts/msm8974/msm8974pro-htc-m8.dts
amessier: I added some compatibles, but I don't know if they are correct and if they will be the same that will be used in the Linux kernel. Let me know if I should change them!

TravMurav: FWIW It's an arbitrary choice and later when/if we need to use panel selection, we can use those compatibles when creating/generating panel drivers. I.e. for msm8916 we generate drivers for known tested panels with a special tool, and assign compatibles in it's config like here:

https://github.com/msm8916-mainline/linux-panel-drivers/blob/921a9179af5624a6980d2f9572b8ebe4ce3e8fc7/config/wingtech-wt88047.sh#L1-L12

TravMurav (approved):
This looks good to me.

> Also let me know what is preferred, fix-up commits, or should I squash the fix-up commits and force push the new changes?

I'd appreciate if you can squash the fixups within this PR into relevant commits, so I'd like to see last 4 commits squashed in

> Watchdog

Gah, something we don't always have in the future branch, I will need to add it later...

minlexx:
> > plz remind me (or someone else) to test this for regressions on 8974-hammerhead or klte before merging...
> 
> @minlexx if TravMurav gives the OK then this would be ready to test on the other 8974 devices, if you are available to! Thanks!

I'll try this weekend :pensive:

minlexx:
Seems to be okay on samsung-klte: https://p.calebs.dev/20b1b5

minlexx:
Seems to be fine on lg-hammerhead https://p.calebs.dev/bc29e2

USB works on both, booting to linux is ok..

amessier:
> I'd appreciate if you can squash the fixups within this PR into relevant commits, so I'd like to see last 4 commits squashed in

Done!

> USB works on both, booting to linux is ok..

Thanks for testing!

TravMurav:
Thanks!

PR #322, opened 2024-01-23
dts: msm8916: samsung-j5x3gjv: rename from j5xn3g

wonderfulShrineMaidenOfParadise:
J5 2016 3G (SM-J510H) doesn't support NFC, thus j5x+3g.


TravMurav:
Thanks!

PR #325, opened 2024-02-18
dts: Fix boot on BQ M5

a-andre:
To boot lk2nd on BQ M5 the corresponding dtb file needs to be appended.

I don't know if with this change lk2nd can still be booted on Lenovo PHAB Plus, which is also part of `msm8939-qrd-skuk.dts`.


TravMurav:
Thanks, applied to rebase/dts and updated next

PR #328, opened 2024-02-23
[rebase-dts] lk2nd: device: dts: add support for huawei-angler

Dominduchami:


Dominduchami:
I have an additional question by the way: Is there a branch with msm8994 support and cmd-autorefresh added that I could do a PR too? 
The goal for me personally is to add support for angler so that one could run the autorefresh command and then boot mainline with fb enabled (this device as well as the Nexus 5X has autorefresh disabled by stock BL). I was planning to send a patch upstream adding a framebuffer node, but first I'd like to have an upstream lk2nd branch/build I could point to in that patch.

Unless there's another soultion I don't know about (it's been some time since I messed around with lk2nd), in which case I'm all ears :-)

lk2nd/device/dts/msm8994/msm8994-huawei-angler.dts
```diff
@@ -0,0 +1,16 @@
...
+/ {
+	model = "Huawei Nexus 6P";
+	qcom,msm-id = <QCOM_ID_MSM8994 0x20000>;
+	qcom,board-id = <8026 0>;
+	qcom,pmic-id = <0x10009 0x1000A 0x0 0x0>;
+};
+
+&lk2nd {
+	model = "Huawei Nexus 6P";
+	compatible = "huawei,angler";
```
TravMurav, line 15: I suppose this would be correct for the one upstream?

```suggestion
	compatible = "huawei,angler";
	
	lk2nd,dtb-files = "msm8994-huawei-angler-rev-101";
```

lk2nd/device/dts/msm8994/msm8994-huawei-angler.dts
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	model = "Huawei Nexus 6P";
```
TravMurav, line 7: I think we decided to not specify model in / unless bootloader explicitly wants it

TravMurav:
> Is there a branch with msm8994 support and cmd-autorefresh added that I could do a PR too?

I have rebase/fb branch that automatically provisions simplefb node when you add `lk2nd.pass-simplefb`, I merge it into rebase/next as well. I was planning to implement autorefresh there too, guarded by cmdline like `lk2nd.pass-simplefb=autorefresh` but I didn't look into that yet.

lk2nd/device/dts/msm8994/msm8994-huawei-angler.dts
Dominduchami: Oh, okay, I''ll test with it removed, if it boots I'll add a follow-up commit

lk2nd/device/dts/msm8994/msm8994-huawei-angler.dts
Dominduchami: Yeah, should be correct to my knowledge

pevik: FYI [`msm8994-huawei-angler-rev-101.dts` from mainline](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/boot/dts/qcom/msm8994-huawei-angler-rev-101.dts) contains:

```
compatible = "huawei,angler", "qcom,msm8994";
```

TravMurav: Yes, we decided to drop the soc compatible for lk2nd since it would be pretty much the same on all nodes anyway and for us doesn't really contain much information.

So for lk2nd I ask for a slightly different convention:

- If possible, most specific compatible (i.e. for a variant of the device, even if hw is the same) which doesn't have to exist upstream.
- A chain of compatibles for this device used upstream (so i.e. either just one generic one or also more specific one if upstream has a dedicated dts for it)
- No non-device-specific compatibles (so no compatible for a SoM a device is based on (if applicable) and no soc compatible)

FWIW for now this property is mostly informational anyway but if we ever i.e. want to match a dtb based on it, "match any "qcom,msm8994" if you have nothing else" is not a great idea.. :D

PR #340, opened 2024-03-23
[experimental-tmp4] dts: msm8994: add LG G4 (h815)

FuchtelJockel:
booting kernel, flashing and buttons are working 

PXL_20240323_192844589

### DIFF
--- a/lk2nd/device/dts/msm8994/msm8992-lg-h815.dts
+++ b/lk2nd/device/dts/msm8994/msm8992-lg-h815.dts
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8992 0>;
+	qcom,board-id = <0xb64 0>;
+	qcom,pmic-id = <0x10009 0x1000a 0x00 0x00>;
+};
+
+&lk2nd {
+	model = "LG G4 (International)";
+	compatible = "lg,h815";
+	lk2nd,dtb-files = "msm8992-lg-h815";
+};
--- a/lk2nd/device/dts/msm8994/rules.mk
+++ b/lk2nd/device/dts/msm8994/rules.mk
@@ -1,0 +1,6 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+LOCAL_DIR := $(GET_LOCAL_DIR)
+
+ADTBS += \
+	$(LOCAL_DIR)/msm8992-lg-h815.dtb \


lk2nd/device/dts/msm8994/msm8992-lg-h815.dts
```diff
@@ -0,0 +1,15 @@
...
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8992 0>, <252 0>;
+	qcom,board-id = <0xb64 0>;
+	qcom,pmic-id = <0x10009 0x1000a 0x00 0x00>;
+};
+
+&lk2nd {
+	model = "LG G4 (International)";
+	compatible = "lg,h815";
```
TravMurav, line 14: If there is some (close-to-)mainline dts for this device, could you add an `lk2nd,dtb-hints = ` property like here:

https://github.com/msm8916-mainline/lk2nd/blob/f47144913fe17851e0e628729ddf6a061cff46e0/lk2nd/device/dts/msm8916/apq8016-sbc.dts#L15

or, if there is none at the moment, "guess" it and add a fixme like:

https://github.com/msm8916-mainline/lk2nd/blob/f47144913fe17851e0e628729ddf6a061cff46e0/lk2nd/device/dts/msm8916/msm8916-asus-z00e.dts#L15

lk2nd/device/dts/msm8994/rules.mk
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
TravMurav, line 1:
```suggestion
# SPDX-License-Identifier: BSD-3-Clause
```

We license make rules as bsd-3-clause

lk2nd/device/dts/msm8994/msm8992-lg-h815.dts
FuchtelJockel: I have a dt for this device that isn't upstreamed currently but I will someday. Should I make it a comment or let it as is?

pevik: @FuchtelJockel +1 for this PR.
BTW what blocks you to send DTS patch to mainline kernel? Because now it will be on second downstream project. Feel free to Cc me when you send the patch.

TravMurav: > I have a dt for this device that isn't upstreamed currently but I will someday. Should I make it a comment or let it as is?

If you have some own close-to-mainline tree and you don't plan on changing the name of the file while upstreaming it, add as is.

TravMurav:
I've amended the commit to fix the makefile comment, thanks!

PR #342, opened 2024-03-26
[rebase/dts] dts: msm8916-samsung: drop panel selection for fortunaltezt

wonderfulShrineMaidenOfParadise:
fortunaltezt uses hx8389c without PWM, which is not used on most of the variants.


TravMurav:
(as discussed in chat) Let's keep the panel selection as-is for now to make sure we don't regress old kernels, it shouldn't cause any harm anyway...

PR #344, opened 2024-03-29
msm8226: Initial support for Galaxy S3 Neo (s3ve3g)

luka177:


dts/msm8226/msm8226-samsung-ms01-r4.dts
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id =  <0x9e00ff01 0x04 0x20000 0x9e08ff01 0x04 0x20000 0xdd08ff01 0x04 0x20000>;
```
TravMurav, line 10:
```suggestion
	qcom,msm-id =  <0x9e00ff01 0x04 0x20000>, <0x9e08ff01 0x04 0x20000>, <0xdd08ff01 0x04 0x20000>;
```
and I assume you only need the second value for msm8226 so probably:
```suggestion
	qcom,msm-id =  <0x9e08ff01 0x04 0x20000>;
```

dts/msm8226/msm8226-samsung-ms01.dts
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id =  <0x91 0x08 0x10001 0x9e 0x08 0x10001 0x9f 0x08 0x10001 0xc6 0x08 0x10001 0xcd 0x08 0x10001>;
```
TravMurav, line 10:
```suggestion
	qcom,msm-id =  <0x91 0x08 0x10001>, <0x9e 0x08 0x10001>, <0x9f 0x08 0x10001>, <0xc6 0x08 0x10001>, <0xcd 0x08 0x10001>;
```

but this looks like MTP which is not what samsung usually uses. Are you sure you need this file at all?

dts/msm8226/msm8226-samsung-ms01-r4.dts
```diff
@@ -0,0 +1,19 @@
...
+	
+	ms01 {
+		model = "Samsung Galaxy S3 Neo";
+		compatible = "samsung,ms01", "qcom,msm8226", "lk2nd,device";
```
TravMurav, line 14: Is there something in the device cmdline that could be used as a match rule here?

dts/msm8226/rules.mk
```diff
@@ -9,6 +9,8 @@ DTBS += \
 	$(LOCAL_DIR)/apq8026-samsung-r09.dtb \
 	$(LOCAL_DIR)/msm8226-motorola-falcon.dtb \
 	$(LOCAL_DIR)/msm8226-samsung-ms013g.dtb \
+	$(LOCAL_DIR)/msm8226-samsung-ms01.dtb \
+       $(LOCAL_DIR)/msm8226-samsung-ms01-r4.dtb \
```
TravMurav, line 13: Please use tabs

dts/msm8226/msm8226-samsung-ms01.dts
luka177: I am not sure how could i fix

TravMurav: drop the file and check if lk2nd still boots on the device?

luka177: Well it for sure will , but i thought those are 2 different revisions

TravMurav: Unless you can confirm that there is a device that really needs to match mtp to work, please drop. This really is not something samsung ever used in my experience, and is just from the default qcom code most likely

luka177: done

dts/msm8226/msm8226-samsung-ms01-r4.dts
TravMurav: I'd still appreciate having some match rule for this, could you share the cmdline you get in the lk log maybe?

PR #345, opened 2024-04-02
dts: msm8916: fortuna/rossa: rename from gprime/cprime

wonderfulShrineMaidenOfParadise:
To avoid breaking panels, samsung,gprime-panel and samsung,cprime-panel are not renamed for now.


TravMurav:
Thanks!

PR #347, opened 2024-04-02
dts: msm8974: Add support for Samsung Galaxy Note 3 (hlte)

FrostI7Alex:
It's a draft for now, as I won't be online for a while (till May).

Why haven't I squashed? Later... Don't want to set up Git on devices I use atm.

IMG_20240402_200901

[lk_log.txt](https://github.com/msm8916-mainline/lk2nd/files/14841425/lk_log.txt)

### DIFF
--- a/dts/msm8974/msm8974-samsung-hlte-r07.dts
+++ b/dts/msm8974/msm8974-samsung-hlte-r07.dts
@@ -17,7 +17,7 @@
 		lk2nd,match-bootloader = "N9005*";
 
 		lk2nd,keys =
-			<KEY_HOME		PM_GPIO(3) PM_GPIO_PULL_UP_1_5>,
+			<KEY_HOME	PM_GPIO(3) PM_GPIO_PULL_UP_1_5>,
 			<KEY_VOLUMEDOWN	PM_GPIO(2) PM_GPIO_PULL_UP_1_5>,
 			<KEY_VOLUMEUP	PM_GPIO(5) PM_GPIO_PULL_UP_1_5>;
 
--- a/dts/msm8974/msm8974-samsung-hlte-r07.dts
+++ b/dts/msm8974/msm8974-samsung-hlte-r07.dts
@@ -7,12 +7,9 @@
 
 / {
 	// This is used by the bootloader to find the correct DTB
-	qcom,msm-id = <0x7E01FF01 7 0>, <0xB801FF01 7 0>, <0xB901FF01 7 0>;
+	qcom,msm-id = <0x7E01FF01 7 0>;
 
 	hlte {
-		qcom,board-id = <1 0x0>;
-		qcom,msm-id = <126 0x10001>;
-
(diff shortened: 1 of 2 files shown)


dts/msm8974/msm8974-samsung-hlte-r07.dts
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton.dtsi>
+#include <lk2nd.h>
+
+/ {
+	// This is used by the bootloader to find the correct DTB
+	qcom,msm-id = <0x7E01FF01 7 0>, <0xB801FF01 7 0>, <0xB901FF01 7 0>;
```
TravMurav, line 10:
```suggestion
	qcom,msm-id = <0x7E01FF01 7 0>;
```

the three in here were: msm8974(126), apq8074(184), msm8274(185), per the highest byte and qcom,ids.h. I don't think you need apq and the 3g-only chip here...

dts/msm8974/msm8974-samsung-hlte-r07.dts
```diff
@@ -0,0 +1,40 @@
...
+
+	hlte {
+		qcom,board-id = <1 0x0>;
+		qcom,msm-id = <126 0x10001>;
```
TravMurav, line 14:
```suggestion
```

this looks bogus and I really don't think any samsung dts would have this pair...

dts/msm8974/msm8974-samsung-hlte-r07.dts
FrostI7Alex: Got it.
Hadn't looked through what they were back then; just kept all three as they were in the dts.

dts/msm8974/msm8974-samsung-hlte-r07.dts
FrostI7Alex: That's what I thought. Looked too _generic_ to me, although my mind went "who knows?" at that time.
Those were based on lk2nd's log and my quick run-through of the documentation for board-id. However, I just can't find anything on board-id for non-Pro MSM8974.

TravMurav: lk2nd has logged what it has read from the hardware fuses using the qcom's code that most oems also use. However samsung is one of the oems who doesn't use that code and instead has their own msm-id scheme with the long 4 byte value + rev number. If you want to make sure downstream works, you /do/ need to use those properties to override lk2nd's guess but you need to set the desired values from downstream dts, i.e. like here:

https://github.com/msm8916-mainline/lk2nd/blob/b02d7582ef456ba1e0297d8a2beb3e098d1d8ce9/dts/msm8916/msm8916-samsung-r08.dts#L48-L49

which is usually only necessary if the rev you need is different from the rev that samsung's bl picks in lk2nd dts, since lk2nd tries to find and reuse samsung's bootloader choice when it can. (not sure how robust is that though)

PR #348, opened 2024-04-03
dts: msm8916: add support for Motorola Moto X 2014 2nd gen.

Michal-Szczepaniak:
If this PR is towards wrong branch please point me to correct one and I will rebase changes

PR #349, opened 2024-04-04
dts: msm8974: add support for Motorola Moto X 2014 2nd gen.

Michal-Szczepaniak:
I'm opening new PR because i couldn't change source branch. I applied CR fixes.

Original PR: #348


TravMurav:
Thanks!

PR #351, opened 2024-04-17
dts: msm8226: Add Motorola Moto G 2014 (titan)

daviewales:
Note: My device boots with the 4b variant. I included the other variants in case they are useful, but let me know if you want me to remove them from the PR.

dts details from here: https://github.com/LineageOS/android_kernel_motorola_msm8226/tree/cm-14.1/arch/arm/boot/dts


TravMurav:
Thanks!

PR #355, opened 2024-05-30
dts: xiaomi-vince: fix dts issues

barni2000:
Xiaomi Redmi 5 Plus (xiaomi-vince) is an qrd device.


TravMurav:
Thanks!

barni2000:
I have accidentally replaced touchscreen compatible. I will open a new MR about it. rmi4 is the primarily supported touchscreen.

PR #356, opened 2024-05-30
lk2nd: smp: msm8953: configure clocks for both clusters

barni2000:


lk2nd/smp/clock-msm8953.c
```diff
@@ -0,0 +1,35 @@
...
+#include <clock.h>
+#include <platform/iomap.h>
+#include <reg.h>
+
+#define MHZ				1000000
+
+static bool platform_has_second_cluster(void)
+{
+	switch (board_platform_id()) {
+		case MSM8953:
```
TravMurav, line 15: apq8053?

also are there any that don't have the second cluster? Or it's "just in case"?

lk2nd/smp/clock-msm8953.c
```diff
@@ -0,0 +1,35 @@
...
+			return true;
+		default:
+			return false;
+	};
+}
+
+static int lk2nd_8953_clock_setup(void *dtb, const char *cmdline,
+				  enum boot_type boot_type)
+{
+
```
stephan-gh, line 27: I'd add a check here to skip this for downstream and lk2nd like in clock-msm8916.c:

```c
	/* Downstream sets this itself */
	if (boot_type & (BOOT_DOWNSTREAM | BOOT_LK2ND))
		return 0;

File: lk2nd/smp/clock-msm8953.c (hunk 1 above)
Comment (barni2000) [reply lk2nd/smp/clock-msm8953.c:15]: I think it just an left over from clock-msm8916.c

File: lk2nd/smp/clock-msm8953.c [hunk 3]
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/* Copyright (c) 2022, Stephan Gerhold <stephan@gerhold.net> */
```
barni2000, line 2: @stephan-gh is this fine? it is basing from clock-msm8916.c what was created by you so guess i should keep the copyright header from it or idk.

lk2nd/smp/clock-msm8953.c
```diff
@@ -0,0 +1,24 @@
...
+static int lk2nd_8953_clock_setup(void *dtb, const char *cmdline,
+				  enum boot_type boot_type)
+{
+
+	/* Downstream sets this itself */
+	if (boot_type & (BOOT_DOWNSTREAM | BOOT_LK2ND))
+		return 0;
+
+	// select clock for secondary cluster
+	clk_get_set_enable("a53ssmux_c0", 800 * MHZ, 1);
```
stephan-gh, line 21:
```suggestion
	clk_get_set_enable("a53ssmux_c0", 800 * MHZ, 0);
```

Is it necessary to enable the clock here? I think this should happen automatically when Linux is booted. Make sure to test that it still works correctly with this change.

TravMurav:
Thanks!

PR #357, opened 2024-05-31
dts: xiaomi-vince: set back rmi4 as default touchscreen

barni2000:
Set back touchscreen-compatible to syna,rmi4-i2c because this is the primarily supported.

I would keep novatek compatible as a comment let users to switch it if they needed it.

### DIFF
--- a/lk2nd/device/dts/msm8953/msm8953-xiaomi-vince.dts
+++ b/lk2nd/device/dts/msm8953/msm8953-xiaomi-vince.dts
@@ -22,23 +22,28 @@
 
 			qcom,mdss_dsi_td4310_fhdplus_video_e7 {
 				compatible = "xiaomi,td4310-fhdplus-e7";
-				touchscreen-compatible = "novatek,nt36525-i2c";
+				touchscreen-compatible = "syna,rmi4-i2c";
+				// touchscreen-compatible = "novatek,nt36525-i2c";
 			};
 			qcom,mdss_dsi_td4310_fhdplus_video_e7_g55 {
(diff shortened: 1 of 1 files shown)


lk2nd/device/dts/msm8953/msm8953-xiaomi-vince.dts
```diff
@@ -22,23 +22,28 @@
 
 			qcom,mdss_dsi_td4310_fhdplus_video_e7 {
 				compatible = "xiaomi,td4310-fhdplus-e7";
-				touchscreen-compatible = "novatek,nt36525-i2c";
+				touchscreen-compatible = "syna,rmi4-i2c";
+				// touchscreen-compatible = "novatek,nt36525-i2c";
```
TravMurav, line 26: I guess I should've noticed they're all the same in the last pr. If you always pick the syna, why novatek exists in the dts? Or they both exist but not tied to the panel?

barni2000: Yes they are not tied to the panel there are same panel with different touch variant so we choosen one touch variant as primary and the other can be enabled by the users.
https://wiki.postmarketos.org/wiki/Xiaomi_Redmi_5_Plus_(xiaomi-vince)#Touchscreen

In long term solution i2c probe, or dt overlays or an extra extlinux parameter could be better.

TravMurav: I see, makes sense.

PR #358, opened 2024-06-04
dts: msm8909-1gb-qrd-skuc: add Acer Liquid Z330 (acer-t01)

wonderfulShrineMaidenOfParadise:


lk2nd/device/dts/msm8909/msm8909-1gb-qrd-skuc.dts
```diff
@@ -0,0 +1,64 @@
...
+
+&lk2nd {
+	acer-t01 {
+		model = "Acer Liquid Z330";
+		compatible = "acer,t01";
+
+		qcom,msm-id = <QCOM_ID_MSM8909 0>;
+		qcom,board-id = <0x1000b 0x8>, <0x1000b 0x1008>,
+				<0x11000b 0x8>, <0x11000b 0x1008>,
+				<0x21000b 0x8>, <0x21000b 0x1008>;
```
TravMurav, line 19: This override looks weird. Do you need it?

lk2nd/device/dts/msm8909/msm8909-1gb-qrd-skuc.dts
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
TravMurav, line 1: Would you mind 
```suggestion
// SPDX-License-Identifier: BSD-3-Clause
```

lk2nd/device/dts/msm8909/msm8909-1gb-qrd-skuc.dts
wonderfulShrineMaidenOfParadise:
```
79321fb1a5dc9631c496dc4371728067e9a658c195364c4440a063c3ee71bdf8  plat_f5-var_1000b-sub_1008.dtb
79321fb1a5dc9631c496dc4371728067e9a658c195364c4440a063c3ee71bdf8  plat_f5-var_1000b-sub_8.dtb
79321fb1a5dc9631c496dc4371728067e9a658c195364c4440a063c3ee71bdf8  plat_f5-var_11000b-sub_1008.dtb
79321fb1a5dc9631c496dc4371728067e9a658c195364c4440a063c3ee71bdf8  plat_f5-var_11000b-sub_8.dtb
79321fb1a5dc9631c496dc4371728067e9a658c195364c4440a063c3ee71bdf8  plat_f5-var_21000b-sub_1008.dtb
79321fb1a5dc9631c496dc4371728067e9a658c195364c4440a063c3ee71bdf8  plat_f5-var_21000b-sub_8.dtb
```

I would assume all of them are valid, even identical.

TravMurav: if you want lk2nd itself to boot on all those variants, you should put those values in the `/ {};` for OEM bootloader to read. lk2nd by default tries to figure out correct board-id automatically, and setting msm/board id in the device node like this is for cases when automatic detection is wrong (i.e. on samsung), so we need to tell lk2nd exactly which msm-id/board-id it should use, which should be a single value, not 6.

This is not used for mainline and if you can check if downstream boots via lk2nd without those props then it's not needed at all.

TravMurav:
Thanks!

PR #359, opened 2024-06-06
dts: msm8952: Add Motorola Moto G5 (cedric)

a-andre:
For some reason lk2nd fails to boot if `$(LOCAL_DIR)/msm8937-mtp.dtb` is included in `rules.mk`.

### DIFF
--- a/lk2nd/device/dts/msm8952/msm8937-motorola-cedric.dts
+++ b/lk2nd/device/dts/msm8952/msm8937-motorola-cedric.dts
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/*
+ * To build for motorola-cedric, add LK2ND_DTBS="msm8937-motorola-cedric.dtb"
+ * to your make cmdline.
+ * cedric does not work with all dtbs enabled; the bootloader gets upset and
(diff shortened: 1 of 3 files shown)


lk2nd/device/dts/msm8952/msm8937-motorola-cedric.dts
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
```
TravMurav, line 5: Perhaps add a comment like this and suggest a quirk of building with a single dtb for now then?

https://github.com/msm8916-mainline/lk2nd/blob/145fd710570af873d0b25a77f1cfb7d00a6b590b/lk2nd/device/dts/msm8916/msm8939-htc-m8qlul.dts#L6-L11

a-andre: I have chosen to add `msm8937-motorola-cedric.dtb` to `rules.mk` and to filter with `LK2ND_DTBS`.

When using `LK2ND_QCDTBS="msm8937-motorola-cedric.dtb"` instead, `LK2ND_ADTBS=""` would also be needed, to get rid of `msm8937-mtp.dtb`.

TravMurav:
Thanks!

PR #362, opened 2024-06-17
dts: msm8909-1gb-qrd-skuc: add FarEasTone Smart 506

bolilingmeng89006:


lk2nd/device/dts/msm8909/msm8909-1gb-qrd-skuc.dts
```diff
@@ -56,4 +56,24 @@
 			};
 		};
 	};
+
+	fareastone-smart506 {
+		model = "FarEasTone Smart 506";
```
TravMurav, line 61: Could you please also add it to the devices.md file in the same commit?

lk2nd/device/dts/msm8909/msm8909-1gb-qrd-skuc.dts
```diff
@@ -56,4 +56,24 @@
...
+		compatible = "fareastone,smart506";
+
+		lk2nd,dtb-files = "msm8909-fareastone-smart506";
+		lk2nd,match-panel;
```
TravMurav, line 65: Any chance there is a better match rule for this device? Could you check cmdline lk2nd gets for any of the bootloader/device or other custom values it passes that we could use?

TravMurav: Ah, I see, then could you please add a comment like here with correct make arguments so it's easier for the next person, and drop the match property completely?

https://github.com/msm8916-mainline/lk2nd/blob/ff96cb9ff649099dd0a1481204b17994e9568ecb/lk2nd/device/dts/msm8916/msm8916-512mb-mtp.dts#L58-L61

```suggestion
```

TravMurav:
Thanks!

PR #376, opened 2024-06-22
dts: xiaomi-markw: fix board-id

### LINKED ISSUE #375: xiaomi-markw (msm8953) doesn't boot: dtb not found
markw has board-id 0x1000b so it looks like it's not mtp but qrd. lk2nd boots if I set `qcom,board-id = <QCOM_BOARD_ID(QRD, 1, 0) 1>;`

vipaoL:
fix #375


TravMurav:
Thanks!

PR #380, opened 2024-06-29
dts: msm8953: Add Lenovo ThinkSmart View (CD-18781Y)

kaechele:
Menu navigation is currently not possible as the only two buttons available on this device are Volume Up and Volume Down. Menu items can be browsed but not selected as there is no button mapped to KEY_HOME.

Note that bootloader unlock is not functional on the ThinkSmart View. But the device will boot any image as long as it is AVB1 signed. For simple and quick signing `magiskboot sign` can be used on `lk2nd.img`.

### DIFF
--- a/lk2nd/device/dts/msm8953/apq8053-lenovo-cd-18781y.dts
+++ b/lk2nd/device/dts/msm8953/apq8053-lenovo-cd-18781y.dts
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/*
+ * While the bootloader can be unlocked using `fastboot oem-unlock` the
+ * device's lk bootloader will crash as long as the bootloader remains
+ * unlocked. Therefor:
+ * - Secure Boot can not be disabled
+ *   However, the device will happily boot any AVB1 signed image. Tools
+ *   such as magiskboot can be used to sign lk2nd.img and make it bootable.
+ * - lk2nd cannot be flashed in fastboot mode
+ *   Instead, use EDL to write the signed lk2nd.img to the boot partition.
+ */
+
+/ {
+	qcom,msm-id = <QCOM_ID_APQ8053 0>;
+	qcom,board-id= <0x1010520 0>;
+};
+
+&lk2nd {
+	model = "Lenovo ThinkSmart View";
+	compatible = "lenovo,cd-18781y";
+
+	lk2nd,dtb-files = "apq8053-lenovo-cd-18781y";
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		/* HACK: map KEY_VOLUMEUP to non-existent button so the actual
+		 * Volume Up button will successfully re-map to KEY_POWER below.
+		 */
+		volume-up-unmap {
+			lk2nd,code = <KEY_VOLUMEUP>;
+			gpios = <&pmic_pon 0 0>;
+		};
+
+		/*
+		 * Remap Volume Up to KEY_POWER to allow selecting menu items as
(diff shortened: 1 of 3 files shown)


lk2nd/device/dts/msm8953/apq8053-lenovo-cd-18781y.dts
```diff
@@ -0,0 +1,31 @@
...
+/ {
+	qcom,msm-id = <QCOM_ID_APQ8053 0>;
+	qcom,board-id= <0x1010520 0>;
+};
+
+&lk2nd {
+	model = "Lenovo ThinkSmart View";
+	compatible = "lenovo,cd-18781y";
+
+    lk2nd,dtb-files = "apq8053-lenovo-cd-18781y";
```
TravMurav, line 17: This is spaces not tabs, I'd appreciate if everything is consistent and uses tabs.

lk2nd/device/dts/msm8953/apq8053-lenovo-cd-18781y.dts
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
TravMurav, line 1: I'd appreciate new submissions being `BSD-3-Clause` like what upstream Linux prefers

lk2nd/device/dts/msm8953/apq8053-lenovo-cd-18781y.dts
```diff
@@ -0,0 +1,31 @@
...
+	qcom,msm-id = <QCOM_ID_APQ8053 0>;
+	qcom,board-id= <0x1010520 0>;
+};
+
+&lk2nd {
+	model = "Lenovo ThinkSmart View";
+	compatible = "lenovo,cd-18781y";
+
+    lk2nd,dtb-files = "apq8053-lenovo-cd-18781y";
+
```
TravMurav, line 18: > the only two buttons available on this device are Volume Up and Volume Down.

Then you may want to map vol+ to KEY_POWER keycode using gpio-keys, the menu loops around so it's fine but the message on the screen will be slightly incorrect. Better than having no menu though

TravMurav (commented):
That's a quite funny device, please also edit the devices.md list in the same commit to add it, Probably want to add some extra notes that the device is "quirky" and document how to fixup the lk2nd.img to let it boot it, may also add a note that the vol+ is mapped to power key for lk2nd. See how other devices document this.

lk2nd/device/dts/msm8953/apq8053-lenovo-cd-18781y.dts
kaechele: I've tried it and it doesn't actually work. Even with the `gpio-keys` configuration in there the Volume Up key still simply navigates through the menu.

kaechele:
Thanks for the review. I've addressed most comments. Key remapping doesn't seem to work. I've committed the configuration I used in the dts file for review purposes, but it doesn't actually function as intended.

lk2nd/device/dts/msm8953/apq8053-lenovo-cd-18781y.dts
TravMurav: Ah, sorry, it's a subtle bug/quirk in the gpio-keys/menu implementation: it checks for the volume keys first and since there is no node for vol+, it uses the default key, which then suppresses the power keycode... 

As a hack, I think you can add a vol+ key with `gpios = <&pmic_pon 0 0>;` to suppress the keycode with an non-existing key, please add a HACK: comment near it if you do

kaechele: That worked, thanks!

TravMurav:
Thanks!

PR #381, opened 2024-06-29
lk2nd: build: add option to sign bootimg

kaechele:
Some targets require the boot image to be AVB1 signed. This change allows specifying the SIGN_BOOTIMG flag at build time to enable signing. By default signing is done using the Android verity test keys from the AOSP tree.
The user can specify their own certificate and key using the BOOTIMG_CERT and BOOTIMG_KEY flags.

Note that the signing script requires Python with the pyasn1, pyasn1_modules and pycryptodome libraries installed.

Credits to Bjoern Kerler for the MIT-licensed bootsignature.py script.


kaechele:
Tested and working on Lenovo ThinkSmart View (CD-18781Y / APQ8053 / MSM8953).

lk2nd/project/lk2nd.mk
```diff
@@ -36,3 +36,7 @@ DEFINES := $(filter-out ENABLE_XPU_VIOLATION=1, $(DEFINES))
 # Build Android boot image
 OUTBOOTIMG := $(BUILDDIR)/lk2nd.img
 MKBOOTIMG_CMDLINE := lk2nd
+
+SIGN_BOOTIMG ?= 0
+BOOTIMG_CERT ?= lk2nd/certs/verity.x509.pem
+BOOTIMG_KEY ?= lk2nd/certs/verity.pk8
```
TravMurav, line 42: Nice! Could you write a short documentation about this in building.md + maybe in a separate commit update the thinksmart comment to explain what `make` line one should use to get a bootable image?

kaechele: Thanks! I've incorporated your suggestions in the latest push.

kaechele:
Ideally, this could be considered when updating the lk2nd package in postmarketOS when this becomes part of a release. That way `pmbootstrap install` would generate properly signed images for the ThinkSmart View.

TravMurav:
Thanks!

> That way pmbootstrap install would generate properly signed images for the ThinkSmart View.

pmbootstrap installs lk2nd.img that is built by pmOS packaging, so to get that, one would need to package a "special" lk2nd for that device (or enable signing for all devices and validate it doesn't break things)

kaechele:
I was thinking a sub-package could be used, e.g. `lk2nd-msm8953-signed`, similar to what is done for `htc-memul`.

TravMurav:
Yes, that's pretty much what I meant by "special" :)

PR #382, opened 2024-06-30
target: msm8953: Add support for second mmc device

barni2000:


TravMurav:
This is still a draft so I assume you will change the commit, but this is a direct backport of the "old" lk2nd fs-boot change, but in the new one we have the fs code explicitly open the sd card only when it's needed. This allows us to at least avoid random crashes with weird old sd-cards so one can always get to the fastboot menu, even if something went wrong.

I'd strongly prefer this to be in-line with the other targets :)

barni2000:
> This is still a draft so I assume you will change the commit, but this is a direct backport of the "old" lk2nd fs-boot change, but in the new one we have the fs code explicitly open the sd card only when it's needed. This allows us to at least avoid random crashes with weird old sd-cards so one can always get to the fastboot menu, even if something went wrong.
> 
> I'd strongly prefer this to be in-line with the other targets :)
> 
> (...) that is, by implementing target_get_sd_mmc() like i.e. this: [6d8312b](https://github.com/msm8916-mainline/lk2nd/commit/6d8312b89f9308149ca4e70c799b19bcfd40a1cf)

msm8953 mmc init a little bit different from older soc, but i will try to modify it.

TravMurav:
Thanks!

PR #383, opened 2024-06-30
target: msm8952: Enable sdcard fs support

barni2000:


barni2000:
@TravMurav Is it seems fine?

TravMurav:
Thanks!

PR #384, opened 2024-06-30
Add MSM8952 cpu clock definitions 

barni2000:
Add cpu clock definitions for make cpufreq happy.

I need to validate some values so it will be draft until.


barni2000:
@TravMurav Is it seems fine?

lk2nd/smp/clock-msm8952.c
```diff
@@ -0,0 +1,91 @@
...
+			return true;
+		default:
+			/* MSM8917,SDM429 have only one CPU cluster */
+			return false;
+	};
+}
+
+static void setup_little_cluster_clocks(void)
+{
+	uint32_t reg = (board_platform_id() == SDM439? 0x44024665: 0x4c015765);
```
TravMurav, line 42: Do those register values have definitions? I'd prefer having them decoded with macros if possible...

TravMurav (commented):
Sorry for the delay, seems like this got lost from my inbox somehow...

lk2nd/smp/clock-msm8952.c
TravMurav: Seems like those are just magic values in downstream... Don't bother decoding then, just rename the variable to `val` please

lk2nd/smp/clock-msm8952.c
```diff
@@ -0,0 +1,91 @@
...
+	/* Initialize PLL configuration to avoid extra code in Linux */
+	writel(PLL_USER_CTL_PLLOUT_LV_MAIN | PLL_USER_CTL_PLLOUT_LV_EARLY,
+				 APCS_C0_PLL_BASE + PLL_USER_CTL);
+	writel(reg, APCS_C0_PLL_BASE + PLL_CONFIG_CTL);
+}
+
+static void setup_big_cluster_clocks(void)
+{
+	if (!platform_has_second_cluster())
+		return;
```
TravMurav, line 62: Any reason this is duplicated in each setup_... instead of in the clock_setup() below? (i.e. just bail out on socs that don't have the second cluster instead of calling setup_...()?

barni2000: I can move it to clock_setup

lk2nd/smp/clock-msm8952.c
```diff
@@ -0,0 +1,89 @@
...
+	if (boot_type & (BOOT_DOWNSTREAM | BOOT_LK2ND))
+		return 0;
+
+	if (!platform_has_second_cluster())
+		return 0;
+
+	switch (board_platform_id()) {
+		case SDM439:
+		case MSM8937:
+		case MSM8940:
```
stephan-gh, line 76: If you look at https://github.com/msm8916-mainline/lk2nd/blob/main/platform/msm8952/platform.c#L205-L369 there are a couple of `case`s missing here, e.g. all the APQ and SDA variants (SoC without modem). Please add them here or alternatively use the `platform_is_msm8937()`, `platform_is_sdm439()` etc checks directly if that is easier.

lk2nd/smp/clock-msm8952.c
```diff
@@ -0,0 +1,89 @@
...
+			setup_little_cluster_clocks();
+			break;
+		case MSM8952:
```
stephan-gh, line 79: Hm I'm looking at https://github.com/caf-backup/quic_la_kernel_msm/blob/0a4ac95ab92c74ad88e2d9c07c7392ea761f61a8/arch/arm/boot/dts/qcom/msm8952.dtsi#L195 and msm8952 uses the 8939 cpu clock driver, i.e. I think it should use `setup_little_cluster_clocks();`. Are you sure this is right?

barni2000: I don't know much about MSM8952 so i am not sure.

lk2nd/smp/clock-msm8952.c
```diff
@@ -0,0 +1,63 @@
...
+static int lk2nd_smp_clock_setup(void *dtb, const char *cmdline,
+				  enum boot_type boot_type)
+{
+	/* Downstream sets this itself */
+	if (boot_type & (BOOT_DOWNSTREAM | BOOT_LK2ND))
+		return 0;
+
+	if (platform_is_msm8937() || platform_is_msm8952() || platform_is_sdm439())
+		setup_little_cluster_clocks();
+	else if (platform_is_msm8956() || platform_is_msm8976_v_1_1())
```
stephan-gh, line 57:
```suggestion
	else if (platform_is_msm8956())
```

8976 is covered already:

```c
int platform_is_msm8956(void)
{
	uint32_t platform = board_platform_id();
	uint32_t ret = 0;

	switch(platform)
	{
		case MSM8956:
		case APQ8056:
		case APQ8076:
		case MSM8976:
			ret = 1;
			break;
		default:
			ret = 0;
	};

	return ret;
}
```

lk2nd/smp/clock-msm8952.c
```diff
@@ -0,0 +1,63 @@
...
+
+#define PLL_USER_CTL_PLLOUT_LV_MAIN	BIT(0)
+#define PLL_USER_CTL_PLLOUT_LV_AUX	BIT(1)
+#define PLL_USER_CTL_PLLOUT_LV_AUX2	BIT(2)
+#define PLL_USER_CTL_PLLOUT_LV_EARLY	BIT(3)
+#define PLL_USER_CTL_MN_EN		BIT(24)
+
+static void setup_little_cluster_clocks(void)
+{
+	uint32_t val = (platform_is_sdm439() ? 0x44024665: 0x4c015765);
```
stephan-gh, line 27: nitpick:

```suggestion
	uint32_t val = (platform_is_sdm439() ? 0x44024665 : 0x4c015765);
```

TravMurav:
Thanks!

PR #388, opened 2024-07-12
on7lte-chn fixes

jja2000:
In the transfer from the old way of working to the new one, the match-bootloader line for on7lte-chn got mangled.

Retested a bit with removing the line from msm8929 and now it seems to work weirdly enough. Don't know what happened there when I got lk2nd working....


TravMurav:
Thanks!

PR #389, opened 2024-07-13
dts: msm8974: Add support for Samsung Galaxy Note 3

FrostI7Alex:
Adds initial support for the Samsung Galaxy Note 3 (hlte).
And also, a few patches by @amessier (#314) are in here as they weren't picked up during the transition to the new fork.

_Finally got my lazy ass to somewhat finish it..._

[lk2nd.log](https://github.com/user-attachments/files/16202239/lk2nd.log)


lk2nd/device/dts/msm8974/rules.mk
```diff
@@ -8,6 +8,7 @@ QCDTBS += \
 	$(LOCAL_DIR)/msm8974-htc-m8.dtb \
 	$(LOCAL_DIR)/msm8974-lge-d855.dtb \
 	$(LOCAL_DIR)/msm8974-lge-hammerhead.dtb \
+	$(LOCAL_DIR)/msm8974-samsung-hlte.dtb \
```
TravMurav, line 11: Hm, would you mind adding the device and the new msm-id to the common samsung.dts? For the new lk2nd we decided to squash all the rev files together since it usually causes a mess otherwise...

FrostI7Alex: Hmm... That's what happens when I do that.
[lk2nd.log](https://github.com/user-attachments/files/16202790/lk2nd.log)
(Unable to boot anything in state, as you can guess from this log.)

Maybe it's still worth keeping different models in different files? Note 3's variants in one, S5's in other? After all, there's still some difference between Pro and non-Pro variant of MSM8974.

TravMurav: If you want to make sure lk2nd can boot downstream, you need to add override values in the device node, that is, you'd add `qcom,msm-id = <0x7E01FF01 7 0>;` (or whatever is correct for your downstream) into the hlte node, then lk2nd will use that value to discover the dtb in downstream images.

FrostI7Alex: Alright, I'll try to do so, then.

FrostI7Alex: Well, that just works.

lk2nd/device/dts/msm8974/samsung.dts
```diff
@@ -4,11 +4,45 @@
...
+			up {
+				lk2nd,code = <KEY_VOLUMEUP>;
+				gpios = <&pmic 5 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			};
+		};
+
+		panel {
+			compatible = "samsung,s6e3fa0", "lk2nd,panel";
+			replace-compatible;
+		};
```
TravMurav, line 43: Something got lost?

FrostI7Alex: Got rid off the downstream panels, leaving only the mainline one.

TravMurav: If you don't need panel selection, drop the node please, the only use for panel {} is when you want to boot mainline and let lk2nd update the dt with the correct display compatible. (that is, when multiple display variants can exist in the same device model) If there is no subnodes, nothing will happen.

FrostI7Alex: Done.

TravMurav:
Thanks!

PR #391, opened 2024-07-26
dts: msm8953: Add Asus Zenfone 3 ZE520KL/ZE552KL (zenfone3)

Smankusors:
both ZE520KL and ZE552KL are using the same device tree on the downstream kernel, at least that's when I'm looking at the kernel source at the LineageOS.

---

btw, I have a question about `lk2nd.dtb-files`

```
lk2nd,dtb-files = "msm8953-asus-zenfone3", "ze552kl-sr1_msm8953-mtp";
```

the first one I followed the mainline kernel convention, while the second one is from the downstream kernel. Wdyt? Without these, the lk2nd still boots normally, though.

### DIFF
--- a/lk2nd/device/dts/msm8953/msm8953-asus-zenfone3.dts
+++ b/lk2nd/device/dts/msm8953/msm8953-asus-zenfone3.dts
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+/dts-v1/;
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8953 0>;
+	qcom,board-id = <21 0>;
+};
+
+&lk2nd {
+	model = "Asus Zenfone 3";
+	compatible = "asus,zenfone3";
+	lk2nd,dtb-files = "msm8953-asus-zenfone3", "ze552kl-sr1_msm8953-mtp";
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		volume-down {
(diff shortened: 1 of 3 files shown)


lk2nd/device/dts/msm8953/msm8953-asus-zenfone3.dts
```diff
@@ -0,0 +1,26 @@
...
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8953 0>;
+	qcom,board-id = <21 0>;
+};
+
+&lk2nd {
+	model = "Asus Zenfone 3";
+	compatible = "asus,zenfone3";
+	lk2nd,dtb-files = "msm8953-asus-zenfone3", "ze552kl-sr1_msm8953-mtp";
```
TravMurav, line 16:
```suggestion
	lk2nd,dtb-files = "msm8953-asus-zenfone3";
```

This is only useful for mainline, and if someone ever decided to reuse this tooling for downstream they're better just renaming the dtb file since for some devices the file name would clash anyway, when oem just modified the mtp/qrd/whatever file...

lk2nd/device/dts/msm8953/msm8953-asus-zenfone3.dts
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
TravMurav, line 1:
```suggestion
// SPDX-License-Identifier: BSD-3-Clause
```

I'd prefer BSD-3 for new submissions since this is what upstream linux generally wants to see for dts files, and lk2nd nowdays tries to follow that a little bit

TravMurav:
Thanks!

PR #392, opened 2024-07-29
dts: msm8953: add panel support for zenfone3

Smankusors:
Actually, the mainline kernel needs this to identify the appropriate driver to use.

### DIFF
--- a/lk2nd/device/dts/msm8953/msm8953-asus-zenfone3.dts
+++ b/lk2nd/device/dts/msm8953/msm8953-asus-zenfone3.dts
@@ -11,16 +11,67 @@
 };
 
 &lk2nd {
-	model = "Asus Zenfone 3";
-	compatible = "asus,zenfone3";
-	lk2nd,dtb-files = "msm8953-asus-zenfone3";
+	ze520kl {
+		model = "Asus Zenfone 3 (ZE520KL)";
+		compatible = "asus,zenfone3", "asus,ze520kl";
+		lk2nd,dtb-files = "msm8953-asus-zenfone3";
+		lk2nd,match-cmdline = "*HW_ID=ZE520KL*";
 
-	gpio-keys {
-		compatible = "gpio-keys";
+		gpio-keys {
+			compatible = "gpio-keys";
 
-		volume-down {
-			lk2nd,code = <KEY_VOLUMEDOWN>;
-			gpios = <&tlmm 13 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			volume-down {
+				lk2nd,code = <KEY_VOLUMEDOWN>;
+				gpios = <&tlmm 13 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			};
+		};
+
+		panel {
+			compatible = "asus,zenfone3-panel", "lk2nd,panel";
+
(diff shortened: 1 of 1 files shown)


lk2nd/device/dts/msm8953/msm8953-asus-zenfone3.dts
```diff
@@ -23,4 +23,32 @@
...
+			compatible = "asus,boe5p2_ili7807b";
+		};
+
+		qcom,mdss_dsi_txd5p5_nt35596_1080p_video {
+			compatible = "asus,txd5p5_nt35596";
+		};
+
+		qcom,mdss_dsi_lce5p5_otm1901a_1080p_video {
+			compatible = "asus,lce5p5_otm1901a";
+		};
```
TravMurav, line 52: Something tells me that "5p5" and "5p2" are display panel sizes, it seems like this device is 5.2?

```suggestion
		qcom,mdss_dsi_tm5p2_r63350_1080p_video {
			compatible = "asus,tm5p2_r63350";
		};

		qcom,mdss_dsi_boe5p2_ili7807b_1080p_video {
			compatible = "asus,boe5p2_ili7807b";
		};
```

TravMurav: ah great those are two different phones...

lk2nd/device/dts/msm8953/msm8953-asus-zenfone3.dts
```diff
@@ -23,4 +23,32 @@
...
+
+		qcom,mdss_dsi_ctc5p5_ili7807b_1080p_video {
+			compatible = "asus,ctc5p5_ili7807b";
+		};
+
+		qcom,mdss_dsi_tm5p5_r63350_1080p_video {
+			compatible = "asus,tm5p5_r63350";
+		};
+
+		qcom,mdss_dsi_tm5p2_r63350_1080p_video {
```
TravMurav, line 38: Considering that they are two different phones (5.5 display and 5.2 display), are there any other differences or the display is the only one, and we can expect the hardware design to be the same? If not, I'd say we should probably split this into two nodes and use match-display here to separate 5.5 and 5.2...

Smankusors: @TravMurav there are about seven variants for each screen size...
https://github.com/LineageOS/android_kernel_asus_msm8953/blob/d4ab72afd66672e7b3e49378e5eaf874c8e09440/include/linux/kernel.h#L22-L38

I don't really plan to port all the features of the Zenfone 3, as I only own the ZE520KL_MP. Should I create 14 nodes, or just 2 nodes (ze520kl and ze552kl)?

TravMurav: Please only add those you can verify working

Smankusors: ok I splitted them into two nodes based on the cmdline, wdyt?

TravMurav: This looks good I think but please squash the commits

TravMurav:
Thanks!

PR #393, opened 2024-07-30
dts: msm8952: Add support for HMD Global Nokia 6

FrostI7Alex:
Adds <i>some</i> support for the HMD Global Nokia 6 (ple).
The only weird thing I've noticed is that the volume keys are swapped around.

<i>The IDs seem to be correct, but still can't boot the downstream.</i>
Tried both, ADTB and QCDTB, but the result seems to be the same.

[lk2nd.log](https://github.com/user-attachments/files/16437903/lk2nd.log)

### DIFF
--- a/Documentation/devices.md
+++ b/Documentation/devices.md
@@ -71,6 +71,7 @@
 ### lk2nd-msm8952
 
 - BQ X5 Plus (Longcheer L9360)
+- HMD Global Nokia 6 (ple)
 - Leeco s2
 - Motorola Moto G5 (cedric) (quirky - see comment in `lk2nd/device/dts/msm8952/msm8937-motorola-cedric.dts`)
 - Redmi Note 3 Pro (kenzo)
--- a/lk2nd/device/dts/msm8952/msm8937-nokia-ple.dts
+++ b/lk2nd/device/dts/msm8952/msm8937-nokia-ple.dts
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8937 0x2000>,
+		      <QCOM_ID_MSM8937 0x3000>;
+	qcom,board-id = <0x9B 0>;
+};
+
+&lk2nd {
+	ple {
+		model = "HMD Global Nokia 6 (ple)";
+		compatible = "nokia,ple";
+		lk2nd,match-device = "PLE";
+
+		lk2nd,dtb-files = "msm8937-nokia-ple";
+
+		gpio-keys {
+			compatible = "gpio-keys";
+			up {
+				lk2nd,code = <KEY_VOLUMEUP>;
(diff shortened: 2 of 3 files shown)


Documentation/devices.md
```diff
@@ -73,6 +73,7 @@
 - BQ X5 Plus (Longcheer L9360)
 - Leeco s2
 - Motorola Moto G5 (cedric) (quirky - see comment in `lk2nd/device/dts/msm8952/msm8937-motorola-cedric.dts`)
+- HMD Global Nokia 6 (ple)
```
TravMurav, line 76:
```suggestion
- BQ X5 Plus (Longcheer L9360)
- HMD Global Nokia 6 (ple)
- Leeco s2
- Motorola Moto G5 (cedric) (quirky - see comment in `lk2nd/device/dts/msm8952/msm8937-motorola-cedric.dts`)
```

I guess you sorted it with N for Nokia but let's keep the list sorted fully alphabetically :)

lk2nd/device/dts/msm8952/msm8937-nokia-ple.dts
```diff
@@ -0,0 +1,47 @@
...
+		      <QCOM_ID_MSM8937 0x3000>;
+	qcom,board-id = <0x9B 0>;
+};
+
+&lk2nd {
+	ple {
+		model = "HMD Global Nokia 6 (ple)";
+		compatible = "nokia,ple";
+		qcom,msm-id = <QCOM_ID_MSM8937 0x2000>,
+			      <QCOM_ID_MSM8937 0x3000>;
```
TravMurav, line 17: This override shouldn't change anything unless the original bootloader is completely bogus, and lk2nd will always only use the first value in the list anyway.

lk2nd/device/dts/msm8952/msm8937-nokia-ple.dts
```diff
@@ -0,0 +1,47 @@
...
+			};
+		};
+
+		panel {
+			compatible = "nokia,ple-panel", "lk2nd,panel";
+
+			fih,mdss_dsi_hx8394a_720p_video {
+				compatible = "nokia,ple-hx8394a";
+			};
+			fih,mdss_dsi_ili7807e_1080p_video {
```
TravMurav, line 36: is this device 720p or 1080p?

Documentation/devices.md
FrostI7Alex: Yeah, you guessed correctly; just forgot about the "HMD Global" part when sorting it.

lk2nd/device/dts/msm8952/msm8937-nokia-ple.dts
FrostI7Alex: Gotcha.

lk2nd/device/dts/msm8952/msm8937-nokia-ple.dts
FrostI7Alex: <i>* Looks through the device tree *</i>
Oh, oops, my mistake.

TravMurav:
Thanks!

PR #394, opened 2024-07-30
dts: msm8952: Add support for Huawei Honor 7C

FrostI7Alex:
Adds <i>some initial</i> support for the Huawei Honor 7C (aum-l41).
The main issue is that it has separate ramdisk and kernel partitions in place of a normal boot partition. And so, there's a need for some more shenanigans to make it work correctly.

<s><i>Haven't tested if it actually boots anything yet.</i></s>
<s>Trying to boot the downstream results in a freeze, whether it's flashed onto a new boot partition or booted without flashing.</s>

[lk2nd.log](https://github.com/user-attachments/files/16431450/lk2ndaum.log)

### DIFF
--- a/lk2nd/device/dts/msm8952/msm8937-huawei-aum.dts
+++ b/lk2nd/device/dts/msm8952/msm8937-huawei-aum.dts
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/*
+ * To flash lk2nd onto the device, you need to flash lk2nd as the
+ * kernel and flash an empty image in place of the ramdisk.
+ * To create an empty image use the following command:
+ * mkbootimg --kernel /dev/null --ramdisk /dev/null --base 0x80000000 --kernel_offset 0x00008000 --ramdisk_offset 0x02000000 --second_offset 0x00f00000 --tags_offset 0x00000100 --pagesize 2048 --header_version 0 -o ramdisk.img"
+ *
+ * As this device has no boot partition, you'll have to create
+ * one yourself.
+ * To create the partition follow these steps:
+ * ./parted /dev/block/mmcblk0
+ * unit kB
+ * resizepart 55 31209799kB
+ * mkpart boot ext2 31209799kB 100%
+ * quit
+ * dd if=/dev/block/mmcblk0p41 of=/dev/block/mmcblk0p56 bs=1 seek=446 count=64
+ */
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8937 0x8192>;
+	qcom,board-id = <8331 0>;
+};
+
+&lk2nd {
+	aum {
+		model = "Huawei Honor 7C (aum-l41)";
+		compatible = "huawei,aum";
+		lk2nd,match-panel;
+
+		lk2nd,dtb-files = "msm8937-huawei-aum";
+
+		gpio-keys {
+			compatible = "gpio-keys";
+			up {
+				lk2nd,code = <KEY_VOLUMEUP>;
+				gpios = <&tlmm 91 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			};
+		};
+
+		panel {
+			compatible = "huawei,aum-panel", "lk2nd,panel";
+
+			lcdkit_aum_l41_va_djn_cpt_ili9881c_5p7_hd_video {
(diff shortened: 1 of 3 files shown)


lk2nd/device/dts/msm8952/msm8937-huawei-aum.dts
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/*
+ * To flash lk2nd onto the device, you have to manually splice the
+ * resulting image into two separate images: ramdisk and kernel.
+ */
```
TravMurav, line 9: I'd appreciate if you could give a bit more detailed instructions here (i.e. how do you generate the kernel image, I think you can ignore the ramdisk part since lk2nd doesn't use it anyway)

FrostI7Alex: Not needing to deal with the ramdisk simplifies it a bit, then.

FrostI7Alex: Let's see... Would that be good enough?
```
/*
 * To flash lk2nd onto the device, you need to flash lk2nd as the
 * kernel and flash an empty image in place of the ramdisk.
 * To create an empty image use the following command:
 * mkbootimg --kernel /dev/null --ramdisk /dev/null --base 0x80000000 --kernel_offset 0x00008000 --ramdisk_offset 0x02000000 --second_offset 0x00f00000 --tags_offset 0x00000100 --pagesize 2048 --header_version 0 -o ramdisk.img"
 */
```

FrostI7Alex: The only thing left is to create the boot partition... The only way to boot anything right now is using <b>fastboot boot</b>.

FrostI7Alex: And as for the partition:
```
/*
 * As this device has no boot partition, you'll have to create
 * one yourself.
 * To create the partition follow these steps:
 * ./parted /dev/block/mmcblk0
 * unit kB
 * resizepart 55 31209799kB
 * mkpart boot ext2 31209799kB 100%
 * quit
 * dd if=/dev/block/mmcblk0p55 of=/dev/block/mmcblk0p56 bs=1 seek=446 count=64
 */
```

lk2nd/device/dts/msm8952/msm8937-huawei-aum.dts
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/*
+ * To flash lk2nd onto the device, you need to flash lk2nd as the
+ * kernel and flash an empty image in place of the ramdisk.
+ * To create an empty image use the following command:
+ * mkbootimg --kernel /dev/null --ramdisk /dev/null --base 0x80000000 --kernel_offset 0x00008000 --ramdisk_offset 0x02000000 --second_offset 0x00f00000 --tags_offset 0x00000100 --pagesize 2048 --header_version 0 -o ramdisk.img"
```
TravMurav, line 10: ugh do they really use two bootimg files for kernel and ramdisk? what if you just completely ignore the ramdisk partition and keep whatever downstream had in it?

Do I understand correctly that lk2nd.img can be just flashed to "kernel"?

lk2nd/device/dts/msm8952/msm8937-huawei-aum.dts
TravMurav: fwiw you don't need boot.img at all if your only goal is to boot mainline :)

lk2nd/device/dts/msm8952/msm8937-huawei-aum.dts
FrostI7Alex: It can be just flashed to "kernel." I just am not sure if it won't cause anything if there's that ramdisk partition there, and trying to erase it just throws you into eRecovery (TWRP in my case).

lk2nd/device/dts/msm8952/msm8937-huawei-aum.dts
FrostI7Alex: Via extlinux.conf, I know. But, imo, the existence of that boot partition simplifies quite a few things with this phone.

TravMurav:
Thanks!

PR #395, opened 2024-08-04
dts: apq8053-lenovo-cd-18781y: change touchscreen-compatible for FT8201

kaechele:
Mainline settled on using the focaltech,ft8201 format instead of the edt prefix for LCMs that aren't actually made by EDT.

Link: https://lore.kernel.org/linux-input/20240804031310.331871-1-felix@kaechele.ca/T/


TravMurav:
Thanks! Nice to see things moving along upstream too :)

PR #399, opened 2024-08-10
dts: msm8909: add mobvoi-catfish

CE1CECL:


lk2nd/device/dts/msm8909/rules.mk
```diff
@@ -6,3 +6,6 @@ QCDTBS += \
 	$(LOCAL_DIR)/msm8905-qrd-skub.dtb \
 	$(LOCAL_DIR)/msm8909-1gb-qrd-skuc.dtb \
 	$(LOCAL_DIR)/msm8909-mtp.dtb \
+
+ADTBS += \
```
wonderfulShrineMaidenOfParadise, line 10: Appended DTB? Doesn't it work as QCDTB?

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
```diff
@@ -0,0 +1,32 @@
...
+                };
+                gpio-keys {
+                        compatible = "gpio-keys";
+                        volume-down {
+                                gpios = <&tlmm 98 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+                                lk2nd,code = <KEY_VOLUMEDOWN>;
+                        };
+                };
+        };
+};
```
wonderfulShrineMaidenOfParadise, line 32:
```suggestion
// SPDX-License-Identifier: GPL-2.0-only

#include <skeleton64.dtsi>
#include <lk2nd.dtsi>

/ {
	model = "Qualcomm APQ8009W";
	qcom,msm-id = <QCOM_ID_APQ8009 0>, <QCOM_ID_APQ8009W 0>;
	qcom,board-id = <QCOM_BOARD_ID_MTP 0x105>;
};

&lk2nd {
	mobvoi-catfish {
		model = "Mobvoi TicWatch Pro (WF12096)";
		compatible = "mobvoi,catfish";
		lk2nd,dtb-files = "mobvoi,catfish";
		lk2nd,match-panel;

		panel {
			compatible = "mobvoi,catfish-panel";

			qcom,mdss_dsi_rm67162_wqvga_cmd {
				compatible = "mobvoi,rm67162";
			};
		};

		gpio-keys {
			compatible = "gpio-keys";

			volume-down {
				gpios = <&tlmm 98 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
				lk2nd,code = <KEY_VOLUMEDOWN>;
			};
		};
	};
};
```

Minor coding style fixes:
- Use tabs instead of spaces
- Newline between properties and nodes.

lk2nd/device/dts/msm8909/rules.mk
CE1CECL: No: Stock fastboot gives "dtb not found", unless you want it in both?

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
CE1CECL: Thanks! (The spaces instead of tabs was a mistake when I copied it from another PC)

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	model = "Qualcomm APQ8009W";
+	qcom,msm-id = <QCOM_ID_APQ8009 0>, <QCOM_ID_APQ8009W 0>;
```
wonderfulShrineMaidenOfParadise, line 8: Looks a bit of weird when you have 2 or more.

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	model = "Qualcomm APQ8009W";
```
TravMurav, line 7: Do you need the "model" to boot?

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
```diff
@@ -0,0 +1,36 @@
...
+	qcom,msm-id = <QCOM_ID_APQ8009 0>, <QCOM_ID_APQ8009W 0>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0x105>;
+};
+
+&lk2nd {
+	mobvoi-catfish {
+		model = "Mobvoi TicWatch Pro (WF12096)";
+		compatible = "mobvoi,catfish";
+		lk2nd,dtb-files = "mobvoi,catfish";
```
TravMurav, line 16: This looks very wrong and is not a usual format for the dtb filenames. If you don't know, or there is not yet a canonical dtb name for the device, this better be a "fixme" comment like on few other devices

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
TravMurav, line 1:
```suggestion
// SPDX-License-Identifier: BSD-3-Clause
```

I'd prefer new submissions to be bsd3, which is also what's now preferred by linux for new dts.

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
CE1CECL: > Looks a bit of weird when you have 2 or more.

The device would reboot if the `model` didn't exist in that location. In fact, I could remove the `model = "Mobvoi Tic...` and it would work, but to make the project cleaner, I did model= twice, one for to be able to boot, and the other to match with the rest of lk2nd

Unless you are looking at `qcom,msm-id `? This came from the stock dtb.

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
CE1CECL: Unfortunately, otherwise the device reboots and fastboot returns "No such device"

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
TravMurav: it was about two msm-ids I believel, do you really need to match non-w variant of the soc if this device is w variant?

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
TravMurav: I see, could you then
```suggestion
	model = "Qualcomm APQ8009W"; /* Required by the oem bootloader. */
```

CE1CECL: I by mistake press shift+enter on the word "Add" but I don't know if this commit message matters or not? If it does I can revert it.

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
CE1CECL: [catfish-stock-boot-dts.zip](https://github.com/user-attachments/files/16571596/catfish-stock-boot-dts.zip)
I'm not sure what you are trying to ask, but I uploaded the stock dts (decompiled already) if you like to take a look. There are no ref's no mobvoi, so it may need a FIXME.

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
TravMurav: You will need to squash all changes into one commit when you're done (and probably drop co-developed tags for review suggestions like this)

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
TravMurav: the lk2nd,dtb-files prop is for booting mainline linux with dtb selected by lk2nd. There is no "qcom/mobvoi,catfish.dts" in mainline and this shouldn't be a correct name. If you already have/plan to make a mainline dts, use that name; otherwise put a comment like this:

https://github.com/msm8916-mainline/lk2nd/blob/fa2a7615bfaa458caf528b5489eb6f032485bd7e/lk2nd/device/dts/msm8916/msm8916-asus-z00e.dts#L15

CE1CECL: Ok, I did a FIXME, but is the skeleton64.dtsi alright? Or should it be skeleton32.dtsi?

TravMurav: doesn't currently matter but prefer correct one, so 32 for 32bit soc

TravMurav (commented):
This looks good to me now, but please squash all commits into one

CE1CECL:
> squash all commits into one

May I ask what commands you'd recommend to use to do this?

CE1CECL:
I wonder if https://github.com/ChrisEric1/lk2nd-new/blob/e38c6482583d7416c8676aa174e4cb69d1dd7117/lk2nd/device/dts/msm8909/apq8009w-wtp.dts#L33 is ok or not? I'm looking at the `(GPIO_ACTIVE_LOW | GPIO_PULL_UP)` part, I kinda just guessed but it works however.

I got the `98` from this line `gpios = <0xa0 0x62 0x1>;`, the 0x62 hex2dec is 98, &tlmm because the section doesn't ref anything about pmic and only tlmm.

```
		gpio_keys {
			compatible = "gpio-keys";
			input-name = "gpio-keys";
			pinctrl-names = "tlmm_gpio_key_active", "tlmm_gpio_key_suspend";
			pinctrl-0 = <0xf0>;
			pinctrl-1 = <0xf1>;

			multi_function_key {
				label = "multi_function_key";
				gpios = <0xa0 0x62 0x1>;
				linux,input-type = <0x1>;
				linux,code = <0x8b>;
				gpio-key,wakeup;
				debounce-interval = <0xf>;
			};
		};
```

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
```diff
@@ -0,0 +1,38 @@
...
+	qcom,msm-id = <QCOM_ID_APQ8009W 0>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0x105>;
+};
+
+&lk2nd {
+	mobvoi-catfish {
+		model = "Mobvoi TicWatch Pro (WF12096)";
+		compatible = "mobvoi,catfish";
+
+		// FIXME: lk2nd,dtb-files = "mobvoi-catfish";
```
wonderfulShrineMaidenOfParadise, line 17:
```suggestion
		// FIXME: lk2nd,dtb-files = "apq8009w-mobvoi-catfish";
```

TravMurav:
Thanks! (The button looks correct, you'd need to resolve pinctrl phandles to see the pull/active flags but they're almost always same everywhere)

PR #403, opened 2024-08-15
Add support for OPPO MSM8937 devices

rcky844:
These devices were released during which the official notice from CCP regarding locking down bootloader unlock access was released, so earlier versions of their bootloaders weren't impossible to crack, but requires some hacks to work. Please check the individual dts files. Thanks.


rcky844:
Note that I have wondered if creating a certificate with OPPO's details will work. But for now, the method specified in the comment block is the most reliable and widely used. Tools have already been developed to abuse this vulnerability.
https://github.com/affggh/oppo_fake_signature

TravMurav:
Thanks!

PR #407, opened 2024-08-23
dts: msm8953: Add panel selection for Xiaomi Redmi 5 (rosy)

barni2000:
Add missing panel selection for Redmi 5.


TravMurav:
Thanks!

PR #409, opened 2024-08-24
platform: msm_shared: Add SDM630/636/660 chip IDs

minlexx:
Add SDM630/660 SoC family chip IDs.

This allows to implement target_baseband_detect() similarly to how [msm8996 did it](https://lxr.trvn.ru/lk2nd/latest/source/target/msm8996/init.c#L512).


TravMurav:
Thanks!

PR #413, opened 2024-08-26
msm8952: Add support for couple msm8917/msm8937/msm8940 devices

barni2000:
This PR adds support for:
- Moto G5S
- Redmi 3S
- Redmi 4X
- Redmi 5A


lk2nd/device/dts/msm8952/msm8937-xiaomi-land.dts
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
TravMurav, line 1:
```suggestion
// SPDX-License-Identifier: BSD-3-Clause
```

I'd prefer bsd3 for new dt submissions, like what linux prefers nowdays. (for all files here)

PR #415, opened 2024-08-27
dts: msm8226: Add support for Motorola Moto G 2014 LTE (thea)

barni2000:
Add Moto G 2014 LTE support it uses gpio-keys for vol-down same should be added for other motos i think.


TravMurav:
Thanks

PR #416, opened 2024-08-27
dts: msm8952: Fix cedric/montana detection with all dtbs enabled

a-andre:
With the model property added, cedric can be detected with all dtbs enabled, but `msm8937-mtp.dtb`.

For some reason motorola's bootloader prefers `msm8937-mtp.dtb` over `msm8937-motorola-cedric.dtb`. So move the matching description for cedric to the former, to fix detection of cedric with all dtbs enabled.


FieryFlames:
Looks good to me, but we should move Montana too provided it's selecting the same DTB (Montana has the same issue)

TravMurav:
Thanks!

PR #418, opened 2024-09-01
dts: msm8952: Add support for Motorola Moto G6 Play (jeter)

Bstaja:
Add device tree for the Motorola Moto G6 Play (jeter).

scr


TravMurav:
Thanks!

PR #421, opened 2024-09-05
Docs: devices: add Motorola Moto G6

FieryFlames:
🧍‍♂️i forgot to do this


TravMurav:
Yeah, I also sometimes miss this in review... Thanks!

FieryFlames:
> Yeah, I also sometimes miss this in review... Thanks!

I wonder if you could do a PR action that checks all device entries in the dtbs exist in the docs

TravMurav:
> I wonder if you could do a PR action that checks all device entries in the dtbs exist in the docs

it's probably possible but too annoying considering its freeform list we have, even pmOS one that checks the wiki breaks all the time with weird changes people do, and that one is formal... Probably easier to just not forget about it xD

PR #433, opened 2024-10-16
doc/dts: msm8974: Remove hammerhead from qcdtbs and document the quirks

barni2000:
LG Nexus 5 (hammerhead) should not be added for QCDTBS list it breaks other devices. It works with ADTBS and QCDTBS list should be empty.

@z3ntu @minlexx

### DIFF
--- a/lk2nd/device/dts/msm8974/rules.mk
+++ b/lk2nd/device/dts/msm8974/rules.mk
@@ -7,7 +7,6 @@ QCDTBS += \
 	$(LOCAL_DIR)/msm8974pro-ac-pm8941-mtp.dtb \
 	$(LOCAL_DIR)/msm8974-htc-m8.dtb \
 	$(LOCAL_DIR)/msm8974-lge-d855.dtb \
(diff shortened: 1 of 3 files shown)


lk2nd/device/dts/msm8974/rules.mk
```diff
@@ -7,7 +7,6 @@ QCDTBS += \
 	$(LOCAL_DIR)/msm8974pro-ac-pm8941-mtp.dtb \
 	$(LOCAL_DIR)/msm8974-htc-m8.dtb \
 	$(LOCAL_DIR)/msm8974-lge-d855.dtb \
-	$(LOCAL_DIR)/msm8974-lge-hammerhead.dtb \
```
wonderfulShrineMaidenOfParadise, line 10: I believe that hammerhead uses QCDTBS. Will check it later.

barni2000: It works with adtb I have checked

bolilingmeng89006: Confirmed working. Thanks.

PR #435, opened 2024-10-26
fix typo

exkc:


TravMurav:
I've reworded the commit message according to our usual rules, thanks

PR #439, opened 2024-11-02
lk2nd: device: menu: display SECURE BOOT availability

exkc:


exkc:
i want a bit of testing

lk2nd/device/menu/menu.c
```diff
@@ -140,6 +140,7 @@ void display_fastboot_menu(void)
 	int y, y_menu, old_scale, incr;
 	unsigned int sel = 0, i;
 	bool armv8 = is_scm_armv8_support();
+	bool secure_boot = is_secure_boot_enable();
```
exkc, line 143: not sure how accurate is that (need more tesing)

TravMurav: this is really /not/ accurate, reporting both false positives and false negatives. We confirmed this recently in the chat on multiple devices.

PR #440, opened 2024-11-04
Add long-press navigation mode

### LINKED ISSUE #432: Allow navigating menu using only power key (smartwatches)
On smartwatches it's usual to only have a single button, the power button. But on these smartwatches it's therefore not possible to navigate the menus right now.

It would be great if it's possible (via dts config?) to switch this over to short power press = navigate down (currently volume down), and long press = select (currently power) so it's possible to navigate the menus.

On some smartwatch bootloaders the vendor has implemented touchscreen support but I imagine this is out of scope, and I personally wouldn't really spend much time to get the touchscreen to work in lk2nd.

TravMurav:
Implement optional single-key navigation scheme: short keypress navigates menu items down, and a long press (>1s) selects the item.

fixes #432

@z3ntu would you mind checking that I've picked the correct smartwatches to use single-key? not sure if one of those has multiple keys...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ target/*/tools/mkheader
 cscope.*
 ncscope.*
 
+# clangd
+.cache
+compile_commands.json
--- a/Documentation/dt-bindings.md
+++ b/Documentation/dt-bindings.md
@@ -151,6 +151,16 @@ them in the device node and lk2nd will use them instead of autodetecting.
 	qcom,board-id = <0xCE08FF01 1>;
 ```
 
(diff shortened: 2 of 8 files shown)


z3ntu:
sparrow, sturgeon & lenok should be good, those are my three msm8226 smartwatches.

I've yet to test this, but awesome to see this implemented!

TravMurav:
Thanks! I've tested this on lenok while writing the thing, feel free to lmk if there are any issues

z3ntu:
Will do, thanks!

PR #443, opened 2024-11-11
dts: msm8216-samsung: add e53g

gAtrium:
This commit adds E5 Duos (E500H) model node to the msm8216-samsung dts. Also lists the device under Documentation/devices.md.
Can't test if the panel works but another user reports that there's display already.

```
$ fastboot getvar all
....
(bootloader) lk2nd:model:Samsung Galaxy E5 (SM-E500H)
(bootloader) lk2nd:compatible:samsung,e35g

```


wonderfulShrineMaidenOfParadise (commented):
Reword it to `dts: msm8216-samsung: add e53g`

TravMurav:
Thanks!

PR #452, opened 2024-11-24
dts: msm8916-samsung: Add support for another J5 2016 variant (j5xnlteb)

semfault:


lk2nd/device/dts/msm8916/msm8916-samsung.dts
```diff
@@ -490,6 +490,25 @@
 
 	/* rev 4 */
 
+	j5xnlteb {
```
wonderfulShrineMaidenOfParadise, line 493: Why?

```suggestion
	j5xnltedx {
```

wonderfulShrineMaidenOfParadise (changes requested):
Remember to update `Documentation/devices.md` as well.

lk2nd/device/dts/msm8916/msm8916-samsung.dts
semfault: oops, was writing the codename from the top of my head

PR #460, opened 2024-12-09
Add lk1st support for Huawei Mediapad T1-A21W

hicode002:
In fact, T1-A21W is the wifi-only version of T1-A21L.
This device does not use tps65132,but shared the same gpios with tps65132
GPIO 32 is used to enable the regulator of LCD
GPIO 97 is used to enable the regulator of the backlight
And huawei uses custom dcs commands to set brightness. 
```
0xf5, your brightness, 0x15, 0x80
```

### DIFF
--- a/app/aboot/aboot.c
+++ b/app/aboot/aboot.c
@@ -4051,7 +4051,6 @@ normal_boot:
 		if (target_is_emmc_boot())
 		{
 			if (!boot_into_recovery) {
-				/* Try to boot from first fs we can find */
 				ssize_t loaded_file = fsboot_boot_first(target_get_scratch_address(), target_get_max_flash_size());
 
 				if (loaded_file > 0)
@@ -4114,6 +4113,7 @@ normal_boot:
 #if FBCON_DISPLAY_MSG
 	display_fastboot_menu();
 #endif
+	
 }
 
 uint32_t get_page_size()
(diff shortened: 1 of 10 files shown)


app/aboot/aboot.c
```diff
@@ -4051,7 +4051,6 @@ void aboot_init(const struct app_descriptor *app)
 		if (target_is_emmc_boot())
 		{
 			if (!boot_into_recovery) {
-				/* Try to boot from first fs we can find */
```
wonderfulShrineMaidenOfParadise, line 4054: Try not to add irrelevant changes.

platform/msm_shared/display_menu.c
```diff
@@ -345,7 +345,7 @@ void display_bootverify_option_menu_renew(struct select_msg_info *msg_info)
 		FBCON_COMMON_MSG, big_factor);
 	display_fbcon_menu_message("Press volume keys to navigate, and "\
 		"press power key to select\n\n", FBCON_COMMON_MSG, common_factor);
-
+	display_fbcon_menu_message("Welcome to HICODE002 BOOTLOADER!\n", FBCON_RED_MSG, common_factor);
```
wonderfulShrineMaidenOfParadise, line 348: Remember to drop customised parts. You can split these into another commit, which would be easier to manage.

wonderfulShrineMaidenOfParadise (changes requested):
Configure your `~/.gitconfig`.
And fix the author with `git commit --amend --author="Shaobai Wang <whyscz@sohu.com>"`

wonderfulShrineMaidenOfParadise:
`legacy` should have been achieved and not supposed to be added more changes.

app/aboot/aboot.c
hicode002: > Try not to add irrelevant changes.

OK

platform/msm_shared/display_menu.c
hicode002: > Refe

Deleted

PR #464, opened 2024-12-16
dts: msm8952: Add support for Motorola Moto G6 Play (XT1922-X, jeter)

andrewgigena:
Hello everyone, this pull request introduces device tree support for the Motorola Moto G6 Play (codenamed "jeter") on the MSM8952 platform, which utilizes the MSM8920 CPU (Snapdragon 427).

The DTS was handwritten based on downstream DTS extracted from [unofficial TWRP images found on XDA](https://xdaforums.com/t/g6-play-unofficial-twrp-xt1922-x.4038873/) and the [official TWRP of G6 (non-play variant)](https://twrp.me/motorola/motorolamotog6.html). Its functionality was tested and verified by @pryzhak, the device owner, via the Matrix community. The `gpio-keys` were explicitly set exactly as the downstream DTS, maybe we can strip down "down" and "power" keys after some more testing with @pryzhak.

Build command (needs bootsignature.py dependencies installed):
```bash
make TOOLCHAIN_PREFIX=arm-none-eabi- SIGN_BOOTIMG=1 LK2ND_ADTBS="msm8920-motorola-jeter.dtb" LK2ND_DTBS="" lk2nd-msm8952  
```

Below is a reference image for the device:
image

### DIFF
--- a/lk2nd/device/dts/msm8952/msm8920-motorola-jeter.dts
+++ b/lk2nd/device/dts/msm8952/msm8920-motorola-jeter.dts
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8920 0x00>;
+	qcom,board-id = <0x43 0x8000>,
+			<0x43 0x83a0>;
+
+	/* 
+	 * model is required by bootloader to pick dtb.
+	 * The bootloader also crashes if model isn't present in every dtb. (see lk2nd.dtsi for more info)
+	 */
+	
+	model = "jeter";
+};
+
+&lk2nd {
+	model = "Motorola Moto G6 Play (jeter) (MSM8920)";
+	compatible = "motorola,jeter";
+	lk2nd,match-device = "jeter";
+
+	lk2nd,dtb-files = "msm8920-motorola-jeter";
+};
--- a/lk2nd/device/dts/msm8952/msm8937-motorola-jeter.dts
+++ b/lk2nd/device/dts/msm8952/msm8937-motorola-jeter.dts
@@ -6,7 +6,8 @@
 / {
 	qcom,msm-id = <QCOM_ID_MSM8937 0x00>;
 	qcom,board-id = <0x4A 0x8100>,
-			<0x4A 0x8000>;
+			<0x4A 0x8000>,
+			<0x4B 0x8000>;
 
 	/* 
 	 * model is required by bootloader to pick dtb.
@@ -17,7 +18,7 @@
 };
 
 &lk2nd {
-	model = "Motorola Moto G6 Play (jeter)";
+	model = "Motorola Moto G6 Play (jeter) (MSM8937)";
 	compatible = "motorola,jeter";
 	lk2nd,match-device = "jeter";
 
--- a/lk2nd/device/dts/msm8952/rules.mk
+++ b/lk2nd/device/dts/msm8952/rules.mk
@@ -5,6 +5,7 @@ ADTBS += \
 	$(LOCAL_DIR)/msm8917-mtp.dtb \
 	$(LOCAL_DIR)/msm8917-xiaomi-rolex.dtb \
 	$(LOCAL_DIR)/msm8917-xiaomi-riva.dtb \
+	$(LOCAL_DIR)/msm8920-motorola-jeter.dtb \
 	$(LOCAL_DIR)/msm8937-huawei-aum.dtb \
 	$(LOCAL_DIR)/msm8937-motorola-jeter.dtb \
 	$(LOCAL_DIR)/msm8937-mtp.dtb \


lk2nd/device/dts/msm8952/msm8920-motorola-jeter.dts
```diff
@@ -0,0 +1,46 @@
...
+
+		down {
+			lk2nd,code = <KEY_VOLUMEDOWN>;
+			gpios = <&pmic_pon GPIO_PMIC_RESIN 1>;
+		};
+
+		power {
+			lk2nd,code = <KEY_POWER>;
+			gpios = <&pmic_pon GPIO_PMIC_RESIN 0>;
+		};
```
TravMurav, line 43: The third value for pmic_pon is non-sensical as it represents flags, which don't matter for pmic_pon. You set both keycodes to the same resin key (which is, usually, vol-) Perhaps just keeping override for vol+ is enough? those other two seem to be default anyway.

lk2nd/device/dts/msm8952/msm8920-motorola-jeter.dts
```diff
@@ -0,0 +1,46 @@
...
+	};
+
```
TravMurav, line 45:
```suggestion
```

lk2nd/device/dts/msm8952/msm8937-motorola-jeter.dts
```diff
@@ -6,20 +6,42 @@
...
 &lk2nd {
-	model = "Motorola Moto G6 Play (jeter)";
+	model = "Motorola Moto G6 Play (jeter) (MSM8937)";
 	compatible = "motorola,jeter";
 	lk2nd,match-device = "jeter";
 
 	lk2nd,dtb-files = "msm8937-motorola-jeter";
+
+	gpio-keys {
+		compatible = "gpio-keys";
```
TravMurav, line 29: the whole node looks like defaults for 8937 platform. Are you sure you need this? I prefer to omit keycode overrides if they do nothing.

lk2nd/device/dts/msm8952/rules.mk
```diff
@@ -19,4 +20,3 @@ ADTBS += \
 
 DTBS += \
     $(LOCAL_DIR)/msm8917-huawei-agassi.dtb \
-
```
TravMurav, line 22: I'd appreciate if you can drop this hunk from your diff, I'd prefer to avoid unnecessary churn of empty lines.

lk2nd/device/dts/msm8952/msm8937-motorola-jeter.dts
andrewgigena: Not sure, I am waiting for @pryzhak for feedback to remove gpio-keys from both DTS to see if their keys still work, I added them before the device even booted up.

lk2nd/device/dts/msm8952/rules.mk
andrewgigena: That was a mistake, fixed.

andrewgigena:
@TravMurav,

I have unchecked the draft option, so the PR is now ready to merge following the reviews. I have implemented the changes you suggested, and @pryzhak has tested it thoroughly. Everything appears to be functioning correctly across all menu options. However, the recovery option is still buggy—it returns to lk2nd instead of entering recovery mode as expected.

We verified the regulators, and the SD card's LDO11 is enabled, allowing booting from it (although this has not been tested). The buttons are functioning as intended.

Regarding the next v21, do you know if there are plans to autobuild for these devices? I am concerned it might not boot on his device due to the signing and single DTB option.

lk2nd/device/dts/msm8952/msm8920-motorola-jeter.dts
```diff
@@ -0,0 +1,26 @@
...
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8920 0x00>;
+	qcom,board-id = <0x43 0x8000>,
+			<0x43 0x83a0>;
+
+	/* 
+	 * model is required by bootloader to pick dtb.
+	 * The bootloader also crashes if model isn't present in every dtb. (see lk2nd.dtsi for more info)
+	 * Build with: make TOOLCHAIN_PREFIX=arm-none-eabi- SIGN_BOOTIMG=1 LK2ND_ADTBS="msm8920-motorola-jeter.dtb" LK2ND_DTBS="" lk2nd-msm8952
```
TravMurav, line 14: Can you also update the devices.md file to note this one is quirky?

TravMurav (commented):
> Regarding the next v21, do you know if there are plans to autobuild for these devices? I am concerned it might not boot on his device due to the signing and single DTB option.

For those cases my ""policy"" is to let the user to build quirky devices. Unfortunately, if you look at the devices.md file and the amount of "quirky" notes, if we were to build a custom image for each of those, it would be even bigger nightmare than letting user to run a command on their own. 

Considering the thing was added without the quirky note, perhaps there is some configuration in which signing/single dtb is not needed? Could you confirm again that it doesn't boot with either signing or single-dtb disabled?

andrewgigena:
We tried again this DTS with different build flags, at the end the device was not quirky, it just casually coincided that the device DTS worked at the same time I did a build with some board-id modifications and tried with signing + single dtb. I will push the changes one more time.

TravMurav:
Thanks!

PR #467, opened 2024-12-20
Feature: Add optional delay before booting via `fastboot boot`

NekoCWD:
Useful when a kernel panic occurs before a debugger is attached


TravMurav:
Thanks!

PR #468, opened 2024-12-20
Huawei G7 Plus / G8 / GX8 / Maimang 4: add Panel

bzy-080408:
added panel for these devices  
panel driver will be push to the repo later

### DIFF
--- a/lk2nd/device/dts/msm8916/msm8939-huawei-rio.dts
+++ b/lk2nd/device/dts/msm8916/msm8939-huawei-rio.dts
@@ -16,4 +16,18 @@
 	compatible = "huawei,rio";
 
 	lk2nd,dtb-files = "msm8939-huawei-rio";
+
+	panel {
+		compatible = "huawei,rio-panel";
+		
+		qcom,mdss_dsi_boe_otm1901_5p5_1080p_video {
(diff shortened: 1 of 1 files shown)


lk2nd/device/dts/msm8916/msm8939-huawei-rio.dts
```diff
@@ -12,8 +12,22 @@
 };
 
 &lk2nd {
-	model = "Huawei G7 Plus / G8 / GX8";
+	model = "Huawei G7 Plus / G8 / GX8 / Maimang 4";
```
wonderfulShrineMaidenOfParadise, line 15: > Maimang 4

Would be better if it's written in the commit message, or split into another commit.

wonderfulShrineMaidenOfParadise (changes requested):

lk2nd/device/dts/msm8916/msm8939-huawei-rio.dts
wonderfulShrineMaidenOfParadise: Other than that, looks good to me.

wonderfulShrineMaidenOfParadise (approved):

TravMurav:
Thanks!

PR #475, opened 2025-01-04
dts: msm8952: xiaomi-rolex: fix panel selection

barni2000:
Downstream node names was miss typed what cause panel selection not worked as expected.


TravMurav:
Thanks!

PR #477, opened 2025-01-11
dts: msm8953: apq8053-lenovo-cd-18781y: update note on quirks

kaechele:
Commit 72e9abacf8812298851ac4c9137fceb0af56bfa4 introduced QCTD images to the MSM8953 builds.
The Lenovo ThinkSmart View (CD-18781Y) will not boot lk2nd images that are built with a QCDT image appended. Add this to the notes in the dts.


TravMurav:
Thanks!

PR #478, opened 2025-01-12
menu: Add a custom menu key description feature

kaechele:
This adds a feature to specify custom key descriptions for the navigation hint in the menu.
This is mostly useful for devices that need to re-map keys, such as https://github.com/msm8916-mainline/lk2nd/blob/main/lk2nd/device/dts/msm8953/apq8053-lenovo-cd-18781y.dts

Please let me know if this creates ginormous memory leaks and security issues, as my C is certainly getting better but it's still not great. I'm aware that you'd eventually have to call `free()` on a `strdup()` but it seems this is done nowhere in the code for all the other fields of the `lk2nd_dev` struct that have something `strdup`'d into it via `lk2nd_device2nd_parse_cmdline` (for example).

Before and after, showing the incorrect key descriptions on the left device (without this patch) and the correct key descriptions on the device on the right:
tsv-lk2nd-custom-keys

### DIFF
--- a/lk2nd/device/device.c
+++ b/lk2nd/device/device.c
@@ -126,6 +126,14 @@ static void parse_dtb(const void *dtb)
 	if (len >= 0)
 		lk2nd_dev.single_key = true;
 
+	val = fdt_stringlist_get(dtb, node, "lk2nd,menu-key-strings", 0, &len);
+	if (val && len > 0)
+		lk2nd_dev.menu_keys.navigate = strndup(val, len);
+
+	val = fdt_stringlist_get(dtb, node, "lk2nd,menu-key-strings", 1, &len);
+	if (val && len > 0)
+		lk2nd_dev.menu_keys.select = strndup(val, len);
+
 	dprintf(INFO, "Detected device: %s (compatible: %s)\n",
 		lk2nd_dev.model, lk2nd_dev.compatible);
 
--- a/lk2nd/device/menu/menu.c
+++ b/lk2nd/device/menu/menu.c
@@ -203,8 +203,10 @@ void display_fastboot_menu(void)
 		fbcon_puts_ln(SILVER, y, incr, true, "Short press to navigate.");
 		fbcon_puts_ln(SILVER, y, incr, true, "Long press to select.");
 	} else {
-		fbcon_puts_ln(SILVER, y, incr, true, "Volume keys to navigate.");
-		fbcon_puts_ln(SILVER, y, incr, true, "Power key to select.");
+		fbcon_printf_ln(SILVER, y, incr, true, "%s to navigate.",
+				(lk2nd_dev.menu_keys.navigate ? lk2nd_dev.menu_keys.navigate : "Volume keys"));
+		fbcon_printf_ln(SILVER, y, incr, true, "%s to select.",
+				(lk2nd_dev.menu_keys.select ? lk2nd_dev.menu_keys.select : "Power key"));
 	}
 
 	/*
--- a/Documentation/dt-bindings.md
+++ b/Documentation/dt-bindings.md
@@ -161,6 +161,31 @@ used to switch the navigation scheme to short/long press of a single key.
 	lk2nd,single-key-navigation;
 ```
 
+### Custom menu navigation key hint
+
+If your device uses re-mapped keys you can customize the hint that is shown in
+the lk2nd boot menu.
+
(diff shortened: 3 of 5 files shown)


lk2nd/device/menu/menu.c
```diff
@@ -203,8 +203,15 @@ void display_fastboot_menu(void)
 		fbcon_puts_ln(SILVER, y, incr, true, "Short press to navigate.");
 		fbcon_puts_ln(SILVER, y, incr, true, "Long press to select.");
 	} else {
-		fbcon_puts_ln(SILVER, y, incr, true, "Volume keys to navigate.");
-		fbcon_puts_ln(SILVER, y, incr, true, "Power key to select.");
+		if(lk2nd_dev.menu_keys.navigate)
+			fbcon_printf_ln(SILVER, y, incr, true, "%s to navigate.", lk2nd_dev.menu_keys.navigate);
+		else
+			fbcon_puts_ln(SILVER, y, incr, true, "Volume keys to navigate.");
```
TravMurav, line 209: nit but I think writing it like this is a little bit easier to read:

```suggestion
		fbcon_printf_ln(SILVER, y, incr, true, "%s to navigate.",
			    (lk2nd_dev.menu_keys.navigate ? lk2nd_dev.menu_keys.navigate : "Volume keys"));
```

lk2nd/device/device.c
```diff
@@ -126,6 +127,13 @@ static void parse_dtb(const void *dtb)
 	if (len >= 0)
 		lk2nd_dev.single_key = true;
 
+	strings = (const char *const *)
+		lkfdt_stringlist_get_all(dtb, node, "lk2nd,menu-key-strings", &len);
```
TravMurav, line 131: This function allocates an array and strdups strings into it already, in this case you will leak them all since you don't save the whole array as done above. First thought would be to copy pointers from the array as-is and free the outer array then but you'd need to handle extra values if len>2 then...

I'd say here you should use something like `blah = strdup(fdt_stringlist_get(...))` and it seem to return NULL if the prop doesn't exist or doesn't have a string of given index

TravMurav:
>  I'm aware that you'd eventually have to call free() on a strdup() but it seems this is done nowhere in the code for all the other fields of the lk2nd_dev

FWIW stuff in lk2nd_dev is never freed since that struct (and it's contents) is expected to live for the rest of the lifetime of the lk2nd, so it's gone when lk2nd is gone xD

lk2nd/device/menu/menu.c
TravMurav: (to note, I implied fixing both this and below one, and maybe align the second line more nicely as my GH message editor doesn't have monospace enabled)

kaechele: Yes, I figured :) I already have it fixed and formatted on my local branch. Thanks!

lk2nd/device/menu/menu.c
```diff
@@ -203,8 +203,10 @@ void display_fastboot_menu(void)
 		fbcon_puts_ln(SILVER, y, incr, true, "Short press to navigate.");
 		fbcon_puts_ln(SILVER, y, incr, true, "Long press to select.");
 	} else {
-		fbcon_puts_ln(SILVER, y, incr, true, "Volume keys to navigate.");
-		fbcon_puts_ln(SILVER, y, incr, true, "Power key to select.");
+		fbcon_printf_ln(SILVER, y, incr, true, "%s to navigate.",
+				(lk2nd_dev.menu_keys.navigate ? lk2nd_dev.menu_keys.navigate : "Volume keys"));
+		fbcon_printf_ln(SILVER, y, incr, true, "%s to select.",
+				lk2nd_dev.menu_keys.select ? lk2nd_dev.menu_keys.select : "Power key");
```
TravMurav, line 209: nit: either both with brackets or both without

```suggestion
				(lk2nd_dev.menu_keys.select ? lk2nd_dev.menu_keys.select : "Power key"));
```

kaechele:
I tested this with:
1. `lk2nd,menu-key-strings` unset
    > Volume keys to navigate. Power key to select.
2. `lk2nd,menu-key-strings = "TEST";`
    > TEST to navigate. Power key to select.
3. `lk2nd,menu-key-strings = "Volume Down", "Volume Up";`
    > Volume Down to navigate. Volume Up to select.

TravMurav:
Thanks!

PR #479, opened 2025-01-12
dts: msm8939-huawei-rio: fix panel selection

bzy-080408:
fixup panel selection for huawei rio

### DIFF
--- a/lk2nd/device/dts/msm8916/msm8939-huawei-rio.dts
+++ b/lk2nd/device/dts/msm8916/msm8939-huawei-rio.dts
@@ -18,7 +18,7 @@
 	lk2nd,dtb-files = "msm8939-huawei-rio";
 
 	panel {
-		compatible = "huawei,rio-panel";
+		compatible = "huawei,rio-panel", "lk2nd,panel";;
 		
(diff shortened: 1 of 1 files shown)


lk2nd/device/dts/msm8916/msm8939-huawei-rio.dts
```diff
@@ -18,7 +18,7 @@
 	lk2nd,dtb-files = "msm8939-huawei-rio";
 
 	panel {
-		compatible = "huawei,rio-panel";
+		compatible = "huawei,rio-panel", "lk2nd,panel";;
```
TravMurav, line 21: Thanks for noticing this! Would you mind fixing the commit so it actually compiles on CI?

```suggestion
		compatible = "huawei,rio-panel", "lk2nd,panel";
```

lk.log
```diff
@@ -0,0 +1,27 @@
+[0] welcome to lk
```
TravMurav, line 1: Seems like you've accidentally checked in this file?

bzy-080408: looks like yes...

TravMurav:
Thanks!

PR #480, opened 2025-01-13
dts: msm8916: Add XiaoXun JZ0145 v33

rcky844:
The OEM of this device is derived from the name of the store that sells these. Although the stock firmware identifies them as "UFI", the JZ series of WiFi modem stick are not the same as the UFI ones.

Match cmdline with a dsi panel unique to this device.


rcky844:
Please wait until https://github.com/msm8916-mainline/linux/pull/386 is merged to merge this.

rcky844:
@TravMurav This is ready then.

TravMurav:
Thanks!

PR #481, opened 2025-01-15
dts: msm8960: add Samsung Galaxy Ace 3 LTE

user0-07161:
This makes lk2nd boot functionally on the samsung-loganrelte. Screen doesn't refresh for some reason. Anything else works.


TravMurav:
Thanks!

PR #483, opened 2025-01-19
dts: msm8952: Add Motorola Moto E5 (nora)

valpackett:
Nora is an MTP device (even though Android DTs have different board-id, ref: [msm8917-nora-p0.dts](https://github.com/motoe5/android_kernel_motorola_msm8937/blob/lineage-17.1/arch/arm/boot/dts/qcom/msm8917-nora-p0.dts) [msm8920-nora-p0.dts](https://github.com/motoe5/android_kernel_motorola_msm8937/blob/lineage-17.1/arch/arm/boot/dts/qcom/msm8920-nora-p0.dts); same deal as #416 – maybe all these IDs should be put into the device node as in the "msm-id override" doc section) available in MSM8917 and MSM8920 variants.

IMG_20250119_040934143_HDR

upd: tianma 571 → ili9881h is the only IC i could find in [commits](https://github.com/motoe5/android_kernel_motorola_msm8937/commit/b0adda93fcc70860f9cdc1f28fa15cad8a1378c9)

### DIFF
--- a/Documentation/devices.md
+++ b/Documentation/devices.md
@@ -80,6 +80,7 @@
 - Huawei MediaPad T3 10 (ags- l09/l03/w09) (quirky - see comment in `lk2nd/device/dts/msm8952/msm8917-huawei-agassi.dts`)
 - Leeco s2
 - Lenovo K5 Play (l38011)
+- Motorola Moto E5 (nora)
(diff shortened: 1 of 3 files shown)


TravMurav:
Thanks!

PR #484, opened 2025-01-26
dts: msm8952: motorola-nora: align panel compatibles with dev tree

valpackett:
We have decided not to list the IC for the only variant where we know it. (https://github.com/msm89x7-mainline/linux-panel-drivers/pull/5)


TravMurav:
Thanks!

Issue #482, opened 2025-01-18
`touchscreen-compatible` cause crashing when `status` property is missng


barni2000:
I have found out `touchscreen-compatible` makes lk2nd to crash when a touchscreen node have not got
the `status` property.

Crashing:
```
        touchscreen@5d {
                compatible = "goodix,gt911";
                reg = <0x5d>;
                interrupts-extended = <&tlmm 65 IRQ_TYPE_LEVEL_LOW>;
                reset-gpios = <&tlmm 64 GPIO_ACTIVE_HIGH>;
                AVDD28-supply = <&pm8937_l10>;
                VDDIO-supply = <&pm8937_l5>;

                touchscreen-size-x = <720>;
                touchscreen-size-y = <1280>;
        };
```

Not Crashing:
```
        touchscreen@5d {
                compatible = "goodix,gt911";
                reg = <0x5d>;
                interrupts-extended = <&tlmm 65 IRQ_TYPE_LEVEL_LOW>;
                reset-gpios = <&tlmm 64 GPIO_ACTIVE_HIGH>;
                AVDD28-supply = <&pm8937_l10>;
                VDDIO-supply = <&pm8937_l5>;

                touchscreen-size-x = <720>;
                touchscreen-size-y = <1280>;

                status = "okay";
        };
```

barni2000:
This is an possible fix
```
diff --git a/lk2nd/device/panel.c b/lk2nd/device/panel.c
index 602438824..e643939cb 100644
--- a/lk2nd/device/panel.c
+++ b/lk2nd/device/panel.c
@@ -99,7 +99,7 @@ static int lk2nd_panel_dt_update(void *dtb, const char *cmdline,
                if (ret < 0) {
                        dprintf(CRITICAL, "Failed to enable %s touchscreen: %d\n",
                                panel->ts_compatible, ret);
-                       return ret;
+                       return 0;
                }
        }
```

TravMurav:
I'd argue this is a bit problematic in case that the said dt has multiple touchscreen nodes with one being "enabled by default", then if we "enable" some other node, we suddenly have two conflicting nodes. failing to boot on missing status at least highlights that possibility.

i.e. I'd argue that if multiple touchscreen nodes exist, they all should be status=disabled, and if there is only one, there is no need in lk2nd touching it at all (so touchscreen-compatible prop can be dropped from lk2nd dt)

I somewhat fail to understand the different usecase here, what do you think?

[...]

PR #487, opened 2025-01-27
lk2nd: device: panel: remove failure on status is missing

barni2000:
Remove failure when ts-compatible is in use and status property is missing from touchscreen node. Make a warning instead of fail.

Issues:
- Message is not appearing in the logs yet.
- Maybe it would be great show this warning on splash.

Fixes: #482

### DIFF
--- a/lk2nd/device/panel.c
+++ b/lk2nd/device/panel.c
@@ -97,9 +97,14 @@ static int lk2nd_panel_dt_update(void *dtb, const char *cmdline,
 
 		ret = fdt_nop_property(dtb, node, "status");
 		if (ret < 0) {
-			dprintf(CRITICAL, "Failed to enable %s touchscreen: %d\n",
-				panel->ts_compatible, ret);
-			return ret;
+			dprintf(CRITICAL,
+				"WARNING!!!!\n"
+				"Failed to enable %s touchscreen: %d\n"
+				"Please check your device tree.\n"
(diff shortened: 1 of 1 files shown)


lk2nd/device/panel.c
```diff
@@ -97,9 +97,14 @@ static int lk2nd_panel_dt_update(void *dtb, const char *cmdline,
...
-			dprintf(CRITICAL, "Failed to enable %s touchscreen: %d\n",
-				panel->ts_compatible, ret);
-			return ret;
+			dprintf(CRITICAL,
+				"WARNING!!!!\n\
+				Failed to enable %s touchscreen: %d\n\
+				Please check your device tree.\n\
+				Maybe a status property is missing from touchscreen nodes.\n",
+				panel->ts_compatible,
+				ret);
```
TravMurav, line 106: FWIW I'd use multiple dprintf so lk2nd prepends it's [000] timestamp to each, otherwise I'd use multiple string literals instead of escaping newline and adding 4 tabs to the string
```
dprintf(CRITICAL,
         "foo\n"
         "bar\n");
```

barni2000:
@TravMurav I have released the draft state, idk if it is possible to notify the users on splash if there is any warning message, maybe it would be a good approach in the future to make them check the logs.

TravMurav:
Thanks!

PR #492, opened 2025-02-07
msm_shared_ dwc: Stop device in dwc_device_reset

FieryFlames:
Based on:
https://github.com/torvalds/linux/blob/master/drivers/usb/dwc3/core.c#L319
https://github.com/torvalds/linux/commit/f4fd84ae0765a80494b28c43b756a95100351a94


TravMurav:
Thanks!

PR #493, opened 2025-02-07
msm8953: add support for Samsung Galaxy Tab A (2018, 10.5) LTE 

leandrofriedrich:
This MR is for adding support for the Samsung SM-T595 Tablet.

Status:
it boots, things like fastboot flash and fastboot boot work

Quirks: 
sometimes USB breaks, even after cold booting, sometimes i noticed it happening when a Kernel initialized usb, and going to Download mode fixed it (survives cold boots), but i can't constantly reproduce it

### DIFF
--- a/lk2nd/device/dts/msm8953/sdm450-samsung-r04.dts
+++ b/lk2nd/device/dts/msm8953/sdm450-samsung-r04.dts
@@ -70,4 +70,13 @@
 
 		lk2nd,dtb-files = "sdm450-samsung-a6plte-r4";
 	};	
+
+	gta2xllte-generic {
+		model = "Samsung Tab A2 XL LTE Rev04";
+		compatible = "samsung,gta2xllte";
+		lk2nd,match-bootloader = "T595*";
+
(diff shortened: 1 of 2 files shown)


lk2nd/device/dts/msm8953/sdm450-samsung-r04.dts
```diff
@@ -70,4 +70,13 @@
 
 		lk2nd,dtb-files = "sdm450-samsung-a6plte-r4";
 	};	
+
+	gta2xllte-generic {
+                model = "Samsung Tab A2 XL LTE Rev04";
```
TravMurav, line 75: Hi! Sorry for the delay, would you mind reconfiguring your editor to use spaces instead of tabs, so it's in line with the rest of the file?

leandrofriedrich: Hello, thanks for letting me know about this, i am on a new (to me) machine so things might not be as i or others expect, it should be fixed now!

TravMurav:
Thanks!

PR #496, opened 2025-02-08
Add support for LG G5 (H850)

DeltaSporos:
Working device tree for the LG G5 (H850, EU model.)

Confirmed that menu navigation keys work, and am able to boot a TWRP image using fastboot boot.


TravMurav:
Thanks!

PR #499, opened 2025-02-16
Fix UFI buttons

185264646:
The button should be used to enter fastboot mode rather than recevory mode. Fix it.


TravMurav:
In old lk2nd branch we had a commit that reassigned HOME to enable fastboot, but seems like I forgot to re-add this commit into the new branch. I've added it now to `main`, could you check if that fixes things? I'm sorry for this

185264646:
> In old lk2nd branch we had a commit that reassigned HOME to enable fastboot, but seems like I forgot to re-add this commit into the new branch. I've added it now to `main`, could you check if that fixes things? I'm sorry for this

Seems a valid fix. I'll close this pr.

PR #502, opened 2025-02-25
lk2nd: build: create a fake ramdisk

J0SH1X:
Create a 1 byte large fake ramdisk that is just 1 zero.

Some devices (in my case Nexus 6P) will not boot a Android bootimage if the bootimage has no ramdisk in it, it can be anything any garbage or whatever just has to be present


J0SH1X:
Tested on newly supported device Huawei Angler - now boots with the generated lk2nd.img instead of falling back to fastboot

Tested on old supported device Asus Zenfone 2 laster - still boots no difference to before.

TravMurav:
It's interesting how we survived without this for so long, but considering usually boot.img contains a ramdisk, I hope this won't break anything... Thanks!

PR #504, opened 2025-02-27
[DNM] dts: msm8994: Add support for LG Nexus 5X (Bullhead)

Dominduchami:


Dominduchami:
Just found out that 5X accepts the lg g4 dtb for some reason, will need to investigate further, changing the pr name for now until this is resolved

lk2nd/device/dts/msm8994/msm8992-lg-bullhead.dts
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+    qcom,msm-id = <QCOM_ID_MSM8992 0>, <QCOM_ID_APQ8092 0>;
```
TravMurav, line 7: APQ phone?

J0SH1X: Nah it's msm8992 with internal modem.

The bootloader seems to favor LG G4 dts over it's own one tho.

That happens cause LG build the Nexus 5X and it reuses their G4's LK.

Doesn't lk2nd have like that match thingy we can set to match to bullheads cmdline ?

TravMurav: Ah, I see, this and lg-h815 both use 2916 dtb, just that one is written as hex... Then you need to combine them into a single file and introduce matching rules for both devices, yes. So bullhead and h815 will share a file.

But here I was asking which of two msm-id are correct: MSM8992 (LTE) or APQ8092 (WiFI only soc)?

Dominduchami: > Ah, I see, this and lg-h815 both use 2916 dtb, just that one is written as hex... Then you need to combine them into a single file and introduce matching rules for both devices, yes. So bullhead and h815 will share a file.
Is there an example of that being done  before?
> 
> But here I was asking which of two msm-id are correct: MSM8992 (LTE) or APQ8092 (WiFI only soc)?
Should be msm8992, although the stock dtb seems to specify both?

PR #507, opened 2025-03-05
Better support for A/B targets for lk1st and lk2nd 

J0SH1X:
Im pretty bored theese days and wanted to test A/B slots on my lumia 950XL

The phone can now use slotted partitions with vAB and dynamic partitions,

LK1ST can now display the current active slot and also switch the slot


lk2nd/device/menu/menu.c
```diff
@@ -15,7 +15,6 @@
 #include <lk2nd/device/keys.h>
 #include <lk2nd/util/minmax.h>
 #include <lk2nd/version.h>
-
```
TravMurav, line 18: Let's not add extra churn, please keep the empty line as-is

lk2nd/device/menu/menu.c
```diff
@@ -127,6 +126,13 @@ static void opt_recoery(void)    { reboot_device(RECOVERY_MODE); }
 static void opt_bootloader(void) { reboot_device(FASTBOOT_MODE); }
 static void opt_edl(void)        { reboot_device(EMERGENCY_DLOAD); }
 static void opt_shutdown(void)   { shutdown_device(); }
+#if SLOTS_SUPPORTED
+static void opt_switch_slot(void){
```
TravMurav, line 130:
```suggestion
static void opt_switch_slot(void)
{
```

lk2nd/device/menu/menu.c
```diff
@@ -147,6 +156,11 @@ static struct {
 		y += incr; \
 	} while(0)
 
+#define fbcon_printfn(color, y, incr, x...) \
+	do { \
+		fbcon_printf(color, y, x); \
+	} while(0)
+
```
TravMurav, line 163: Why not just call fbcon_printf directly?

lk2nd/device/menu/menu.c
```diff
@@ -228,6 +242,11 @@ void display_fastboot_menu(void)
 	if (lk2nd_dev.bootloader)
 		fbcon_printf_ln(SILVER, y, incr, false, " Bootloader:  %s", lk2nd_dev.bootloader);
 #endif
+#if SLOTS_SUPPORTED
+		int current_active_slot = partition_find_active_slot();
+		fbcon_printfn(SILVER, y, incr, false, " Active Slot:  ");
+		fbcon_printf_ln(current_active_slot ? GREEN : RED, y, incr, false, "              %s", current_active_slot ? "B" : "A");
+#endif
```
TravMurav, line 249: It's unclear to me why A/B should be color-coded, nor why B should be green/"good" and A should be red/"bad"

platform/msm_shared/include/ab_partition_parser.h
```diff
@@ -79,7 +79,11 @@ int partition_fill_partition_meta(char has_slot_pname[][MAX_GET_VAR_NAME_SIZE],
...
-{
-	return target_is_emmc_boot() && _partition_multislot_is_supported();
+static inline bool partition_multislot_is_supported(void){
+#if SLOTS_SUPPORTED
+	return true;
+#else
+	return false;
+#endif
 }
+
```
TravMurav, line 89: Do I understand correctly that `SLOTS_SUPPORTED` "forces" slots on instead of relying on caf lk's code that detects the presence of blah_a and blah_b? In this case, if this is just an lk2nd extension, please rename it to something like `LK2ND_FORCE_AB_SLOTS` and keep existing behavior if it's not enabled. Do I understand correctly that you need it due to dynamic partitions?

(...) thinking of this more I'm not sure how to integrate this into the menu nicely....

app/aboot/rules.mk
```diff
@@ -13,7 +13,6 @@ OBJS += \
 
 ifeq ($(ABOOT_STANDALONE), 1)
 DEFINES += ABOOT_STANDALONE=1
-OBJS := $(filter-out $(LOCAL_DIR)/recovery.o, $(OBJS))
```
TravMurav, line 16: Do we really need the recovery code?

lk2nd/device/menu/menu.c
J0SH1X: It just looks cool, lol.

No color is good/bad for me it just looks cool as said but we can keep it silver if you like it more

platform/msm_shared/include/ab_partition_parser.h
J0SH1X: I need to look into this again either the detection code was just plainly broken or didn't work for me it was some weeks ago I fixed the code.

Iirc it was badly ported over from edk2's aboot impl (probably taken from shift phone xbl)

PR #512, opened 2025-03-17
lk2nd: device: Add an LED driver

185264646:
I use lk1st for most of the time and find it really annoying to has no LED indication in lk1st. So here is the driver to resolve the issue. The code is based on keymap.c, though much simpler and provides no API for other modules. You need to add a node to the dts like below:
```dts
leds {
    compatible = "gpio-leds";
    red {
        gpios = <&tlmm 3 GPIO_ACTIVE_LOW>;
        default-state = "on" // or "off", "keep"
    };
};
```

Tested on a downstream zhihe device with lk2nd.

Not all error paths are tested yet. Need review.

### DIFF
--- a/Documentation/dt-bindings.md
+++ b/Documentation/dt-bindings.md
@@ -247,3 +247,34 @@ new keycode to the actual key.
 		};
 	};
 ```
+
+### GPIO leds
+
+lk2nd supports the `gpio-leds` node. This node follows upstream binding except
+that most properties are not supported. Currently only `gpios` and
+`default-state` are supported in lk2nd.
+
+```
+gpio-leds {
+    compatible = "gpio-leds";
+
+    red {
+        gpios = <&tlmm 3 GPIO_ACTIVE_HIGH>;
+        default-state = "on";
+    };
+
+    green {
+        gpios = <&tlmm 4 GPIO_ACTIVE_HIGH>;
+        default-state = "off";
+    };
+};
+```
+
+This driver is mainly for lk1st since no previous boot-loader is capable to
+light up an LED in this case. This makes it unclear what state currently the
+device is in. If you wish to make sure some leds are enabled or disabled by
+lk1st on boot, you can use this driver.
+
+lk2nd doesn't implement a fully-fledged led subsystem. The only purpose for
+this driver is for debugging and the LED behavior consistancy between lk1st
+and lk2nd.
--- a/lk2nd/device/leds.c
+++ b/lk2nd/device/leds.c
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/* Copyright (c) 2025 Yang Xiwen <forbidden405@outlook.com> */
+
+#include <debug.h>
+#include <libfdt.h>
+
+#include <lk2nd/hw/gpio.h>
+
+#include "device.h"
+
+static int lk2nd_leds_init(const void *dtb, int node)
+{
+	int i = 0, subnode, len;
+	struct gpiol_desc desc;
+	const char *state;
+
+	dprintf(SPEW, " Initializing LEDs\n");
+	dprintf(SPEW, " | label | gpio        | state |\n");
+
+	fdt_for_each_subnode(subnode, dtb, node) {
+		state = fdt_getprop(dtb, subnode, "default-state", &len);
+		if (len < 0 || state[len - 1] != '\0') {
+			dprintf(CRITICAL, "leds: Failed to get property \"default-state\". Broken dtb?\n");
+			continue;
+		}
+
+		if (!strcmp(state, "on")) {
+			gpiol_get(dtb, subnode, NULL, &desc, GPIOL_FLAGS_OUT_ASSERTED);
+		} else if (!strcmp(state, "off")) {
+			gpiol_get(dtb, subnode, NULL, &desc, GPIOL_FLAGS_OUT_DEASSERTED);
+		} else if (!strcmp(state, "keep")) {
+			/* do nothing */
+		} else {
+			dprintf(CRITICAL, "leds: Unknown \"default-state\" value: %s\n", state);
+			continue;
+		}
+
+		i++;
+
+		dprintf(SPEW, " | %5s | 0x%02x 0x%04x | %5s |\n",
+			fdt_get_name(dtb, subnode, NULL), desc.dev, desc.pin, state);
+	}
+	dprintf(INFO, "leds: %d leds configured\n", i);
+
+	if (subnode < 0 && subnode != -FDT_ERR_NOTFOUND) {
+		dprintf(CRITICAL, "leds: Failed to parse subnodes: %d\n", subnode);
+		return subnode;
+	}
+
+	return 0;
+}
+
+LK2ND_DEVICE_INIT("gpio-leds", lk2nd_leds_init);
--- a/lk2nd/device/rules.mk
+++ b/lk2nd/device/rules.mk
@@ -18,6 +18,7 @@ OBJS += \
 	$(LK2ND_DEVICE_OBJ) \
 	$(LOCAL_DIR)/panel.o \
 	$(LOCAL_DIR)/keys.o \
+	$(LOCAL_DIR)/leds.o \
 
 ifneq ($(LK2ND_COMPATIBLE),)
 DEFINES += LK2ND_COMPATIBLE="$(LK2ND_COMPATIBLE)"


lk2nd/device/leds.c
```diff
@@ -0,0 +1,80 @@
...
+
+		if (!strcmp(state, "on"))
+			estate = LED_STATE_ON;
+		else if (!strcmp(state, "off"))
+			estate = LED_STATE_OFF;
+		else if (!strcmp(state, "keep"))
+			estate = LED_STATE_KEEP;
+
+		switch (estate) {
+		case LED_STATE_ON:
```
TravMurav, line 51: TBH considering this state value isn't used anywhere else, I'd just fold this switch and if/elif above and set the gpio immediately in "on/off"

lk2nd/device/leds.c
```diff
@@ -0,0 +1,80 @@
...
+			/* Simply leave this pin alone */
+			break;
+		case LED_STATE_UNKNOWN:
+			dprintf(CRITICAL, "leds: Unknown \"default-state\" value: %s\n", state);
+			continue;
+		}
+
+		i++;
+
+		dprintf(SPEW, " | %5s | 0x%02x 0x%04x |\n",
```
TravMurav, line 67: I'd also put the "state" here I guess but for such a simple driver that doesn't actually store anything, could as well just omit the state table completely. Line below would say "x leds configured" and that would be enough IMO.

lk2nd/device/leds.c
```diff
@@ -0,0 +1,80 @@
...
+
+#include <debug.h>
+
+#include <libfdt.h>
+
+#include <lk2nd/hw/gpio.h>
+
+#include "device.h"
+
+#define LK2ND_MAX_LEDS 16
```
TravMurav, line 12: Could remove the constraint since there is no table those leds are saved to.

lk2nd/device/leds.c
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/* Copyright (c) 2025 Yang Xiwen <forbidden405@outlook.com> */
+
+#include <debug.h>
+
```
TravMurav, line 5: this is very nitty but 
```suggestion
```

lk2nd/device/leds.c
```diff
@@ -0,0 +1,80 @@
...
+	if (subnode < 0 && subnode != -FDT_ERR_NOTFOUND) {
+		dprintf(CRITICAL, "leds: Failed to parse subnodes: %d\n", subnode);
+		return subnode;
+	}
+
+	return 0;
+}
+
+LK2ND_DEVICE_INIT("gpio-leds", lk2nd_leds_init);
```
TravMurav, line 80: I'd really appreciate if this was documented in dt-bindings.md like gpio-keys, no need to write much, just that the driver only sets the led on/off(/keep) on boot for debugging purposes and that it follows upstream binding in the form of only two properties we support here.

lk2nd/device/leds.c
```diff
@@ -0,0 +1,80 @@
...
+
+static int lk2nd_leds_init(const void *dtb, int node)
+{
+	int i = 0, subnode, len;
+	struct gpiol_desc desc;
+	const char *state;
+	enum led_state estate = LED_STATE_UNKNOWN;
+
+	dprintf(SPEW, " Initializing LEDs\n");
+	dprintf(SPEW, " | label | code  | gpio        |\n");
```
TravMurav, line 29: this header doesn't match the table output below ("code" is useless for leds)

lk2nd/device/leds.c
185264646: ok. I've also removed the enum.

lk2nd/device/leds.c
185264646: removed code, added state

lk2nd/device/leds.c
185264646: I've written a simple one. But maybe it still needs more detail so i leave this unresolved.

185264646:
There's some questions:
1. should we document all LEDs? For clarify i think at least those are on by default should be there. But what about other LEDs?
2. should we use "keep" or "on"/"off"? I prefer not using "keep" for clarity.
3. should we following vendor uboot convention or use a custom one? Both seems reasonable. With a different LED state,we can clearly know it's in lk2nd or vendor lk. But linux needs to follow vendor settings.

I wish these questions to be documented too. Not only the binding.

Documentation/dt-bindings.md
```diff
@@ -247,3 +247,40 @@ new keycode to the actual key.
 		};
 	};
 ```
+
+### GPIO leds
+
+We also support the `gpio-leds` node. This node follows upstream binding except
```
TravMurav, line 253:
```suggestion
lk2nd supports the `gpio-leds` node. This node follows upstream binding except
```

My personal preference is to write documentation from the point of the software, not the people making it

Documentation/dt-bindings.md
```diff
@@ -247,3 +247,40 @@ new keycode to the actual key.
...
+
+This driver is mainly for lk1st since no previous boot-loader is capable to
+light up an LED in this case. This makes it unclear what state currently the
+device is in.  It's recommended to document all the LEDs here if lk1st is
+available for your device.
+
+Unlike linux, we don't want to introduce an LED subsystem. There is no public
+API for LED operations like turning LEDs on/off. And there is no LED trigger
+(yet). The only purpose for this driver is for debugging and the LED behavior
+consistancy between lk1st and lk2nd.
```
TravMurav, line 286:
```suggestion
lk2nd doesn't implement a fully-fledged led subsystem. The only purpose for
this driver is for debugging and the LED behavior consistancy between lk1st
and lk2nd.
```

lk2nd/device/leds.c
```diff
@@ -0,0 +1,53 @@
...
+			continue;
+		}
+
+		if (!strcmp(state, "on"))
+			gpiol_get(dtb, subnode, NULL, &desc, GPIOL_FLAGS_OUT_ASSERTED);
+		else if (!strcmp(state, "off"))
+			gpiol_get(dtb, subnode, NULL, &desc, GPIOL_FLAGS_OUT_DEASSERTED);
+		else if (!strcmp(state, "keep"))
+			; /* do nothing */
+		else {
```
TravMurav, line 33: I'd strongly prefer if/elseif/else constructions to either all have brackets, or none to have. Let's put brackets in all cases here.

Documentation/dt-bindings.md
```diff
@@ -247,3 +247,40 @@ new keycode to the actual key.
...
+    red {
+        gpios = <&tlmm 3 GPIO_ACTIVE_HIGH>;
+        default-state = "on";
+    };
+
+    green {
+        gpios = <&tlmm 4 GPIO_ACTIVE_HIGH>;
+        default-state = "keep";
+    };
+
```
TravMurav, line 270: > should we use "keep" or "on"/"off"? I prefer not using "keep" for clarity.

I feel like "keep" makes no sense for lk1st and such a simple driver, since the only goal for us is to set initial state of the led (most often to "on"). I suggest just omitting it here and in the driver (will just be a warning anyway)

Documentation/dt-bindings.md
```diff
@@ -247,3 +247,40 @@ new keycode to the actual key.
...
+        gpios = <&tlmm 4 GPIO_ACTIVE_HIGH>;
+        default-state = "off";
+    };
+};
+```
+
+This driver is mainly for lk1st since no previous boot-loader is capable to
+light up an LED in this case. This makes it unclear what state currently the
+device is in.  It's recommended to document all the LEDs here if lk1st is
+available for your device.
```
TravMurav, line 281: > "It's recommended to document"

I'd actually word this more softly, like "if you wish to make sure some leds are enabled or disabled by lk1st on boot, you can use this driver"

TravMurav (commented):
> should we document all LEDs? For clarify i think at least those are on by default should be there. But what about other LEDs?

I think of lk2nd's dtb as "utilitarian" rather than descriptive: if something allows us to reach a goal (i.e. enable a led that wasn't enabled) - we add it, if it's only descriptive - there is no use of the node for us. I'd prefer only leds which state needs to be set to be added to lk2nd dtb.

> should we use "keep" or "on"/"off"? I prefer not using "keep" for clarity.

answered on the file but from above, "keep" node is same as no node IMO.


> should we following vendor lk convention or use a custom one lighting? Both seems reasonable. With a different LED state,we can clearly know it's in lk2nd or vendor lk. But linux needs to follow vendor settings afaik. We want to avoid LED flip of course.

I'd actually prefer to /not/ have strong rules on how specific devices should be enabled, I prefer "stakeholders" of specific devices to decide what's best, i.e. in case of zhihe devices - if you think it's best to keep it same as what oem did - then it's best.

Documentation/dt-bindings.md
185264646: It is a preperty supported by  linux gpio-driver. I think it's better to be a no-op without a warning. It's different with some other random definitely wrong things. But l think it's also okay to omit it anyway.

Documentation/dt-bindings.md
185264646: Seems reasonable.

lk2nd/device/leds.c
185264646: I still prefer to make 'keep' a valid no-op rather than an error to remain consistancy with linux dt-binding.

lk2nd/device/leds.c
```diff
@@ -0,0 +1,53 @@
...
+			dprintf(CRITICAL, "leds: Failed to get property \"default-state\". Broken dtb?\n");
+			continue;
+		}
+
+		if (!strcmp(state, "on")) {
+			gpiol_get(dtb, subnode, NULL, &desc, GPIOL_FLAGS_OUT_ASSERTED);
+		} else if (!strcmp(state, "off")) {
+			gpiol_get(dtb, subnode, NULL, &desc, GPIOL_FLAGS_OUT_DEASSERTED);
+		} else if (!strcmp(state, "keep")) {
+			; /* do nothing */
```
TravMurav, line 32:
```suggestion
			/* do nothing */
```

TravMurav (commented):
One nit, otherwise this looks good to me

PR #516, opened 2025-04-11
Add Samsung Galaxy J5 2016 U+ (SM-J510L)

cherryband:
SM-J510L is a S. Korean U+ carrier variant of SM-J510x. AFAICT no additional tweaks are needed other than just adding it as a supported device in the DTS.

There are other variants such as J510S/K (for different carriers) but this is the only hardware I have and thus have tested on.


TravMurav:
Thanks!

PR #517, opened 2025-04-11
Better Android support, plus a bit of improvements

me-cafebabe:
Please check out each commits for details.

### DIFF
--- a/app/aboot/aboot.c
+++ b/app/aboot/aboot.c
@@ -3233,7 +3233,9 @@ void cmd_boot(const char *arg, void *data, unsigned sz)
 	unsigned int kernel_size = 0;
 	enum boot_type boot_type = 0;
 #if DEVICE_TREE
+	void * image_buf = NULL;
 	uint8_t dtb_copied = 0;
+	unsigned dtb_image_size = 0;
 	unsigned int scratch_offset = 0;
 #endif
 #if VERIFIED_BOOT_2
@@ -3248,6 +3250,9 @@ void cmd_boot(const char *arg, void *data, unsigned sz)
 #ifdef MDTP_SUPPORT
         static bool is_mdtp_activated = 0;
 #endif /* MDTP_SUPPORT */
+#endif
+#ifdef OSVERSION_IN_BOOTIMAGE
+	uint32_t dtb_image_offset = 0;
 #endif
 
 #if FBCON_DISPLAY_MSG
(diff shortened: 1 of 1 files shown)


lk2nd/project/lk2nd.mk
```diff
@@ -11,11 +11,14 @@ ifneq ($(ENABLE_FBCON_DISPLAY_MSG),1)
...
 endif
 
-# Use part of the "boot" partition for the lk2nd boot image. The real Android
-# boot image can be placed in the partition with 512 KiB offset.
-LK2ND_PARTITION_BASE ?= boot
-LK2ND_PARTITION_NAME ?= lk2nd
-LK2ND_PARTITION_SIZE ?= 512*1024
+# Use part of the "boot" and "recovery" partition for the lk2nd boot image.
+# The real Android boot image can be placed in the partition with 512 KiB offset.
+LK2ND_BOOT_PARTITION_BASE ?= boot
```
TravMurav, line 16: If the goal is making an image that would boot from recovery, I wonder if it's simpler to document how to compile and use an image with `make LK2ND_BOOT_PARTITION_BASE=recovery` instead of introducing complicated logic of checking and retrying images on multiple partitions that may or may not exist on the device...

app/aboot/recovery.c
```diff
@@ -428,8 +428,8 @@ int _emmc_recovery_init(void)
 		boot_into_recovery = 1;
 	}
 
-	if (!strcmp("update-radio",msg->command))
-	{
+#if !ABOOT_STANDALONE
+	if (!strcmp("update-radio",msg->command)) {
```
TravMurav, line 432: I'd appreciate if we could avoid churn here, even if the old code is slightly dumber. I'd prefer this only add the `#if` and `#endif` and doesn't reformat the code inbetween

lk2nd/project/lk2nd.mk
TravMurav: Discussed in chat: The mainline android + mainline recovery would both need to go via lk2nd and if android reboots into recovery, oem aboot will boot recovery partiiton, so lk2nd needs to be installed to both.

app/aboot/aboot.c
```diff
@@ -4754,7 +4813,7 @@ void cmd_reboot_fastboot(const char *arg, void *data, unsigned sz)
 		return;
 	}
 	fastboot_okay("");
-	reboot_device(REBOOT_MODE_UNKNOWN);
+	reboot_device(RECOVERY_MODE);
```
TravMurav, line 4816: I don't exactly understand how "reboot fastboot" ends up with "recovery mode"

app/aboot/aboot.c
```diff
@@ -5711,7 +5770,7 @@ void aboot_init(const struct app_descriptor *app)
 		}
 		else
 		{
-		if (!IS_ENABLED(ABOOT_STANDALONE))
+		if (!IS_ENABLED(ABOOT_STANDALONE)) // TODO: detect_android_from_flash()
```
TravMurav, line 5773:
```suggestion
		if (!IS_ENABLED(ABOOT_STANDALONE))
```

I sincerely doubt anyone would care to boot mainline android on flash devices, let's avoid code churn here

lk2nd/device/menu/menu.c
```diff
@@ -123,7 +124,10 @@ static uint16_t wait_key(void)
 
 static void opt_continue(void)   { cmd_continue(NULL, NULL, 0); }
 static void opt_reboot(void)     { reboot_device(0); }
-static void opt_recovery(void)   { reboot_device(RECOVERY_MODE); }
+static void opt_recovery(void)   {
+	boot_into_recovery = 1;
+	cmd_continue(NULL, NULL, 0);
+}
```
TravMurav, line 130: I'd prefer this to be confined to the function scope
```suggestion
static void opt_recovery(void)
{
	extern unsigned boot_into_recovery;

	boot_into_recovery = 1;
	cmd_continue(NULL, NULL, 0);
}
```

app/aboot/aboot.c
TravMurav: sure thing but I still sincerely doubt you'd run LOS on NAND targets like MDM modem chipsets so I don't see the need for a todo here, and I strongly prefer to have minimal changes to CAF code when possible

app/aboot/aboot.c
TravMurav: Ah, right... So new android needs the fastbootd for the virtual parititons mess I guess. It makes sense I guess though not sure if it'd break the usecase of rebooting into oem bootloader/lk1st fastboot. Need to think about it...

PR #522, opened 2025-04-23
dts: msm8960: add Samsung Galaxy Tab 2 10.1 (SGH-I497)

mhamzahkhan:
This is quite an old device. It's an AT&T specific variant of the Galaxy Tab 2 10.1 that was given to me years ago and has been collecting dust so I figured it would be fun to play with.


TravMurav:
Thanks! It's interesting that this is a qcom variant and there is another one with OMAP as well...

PR #528, opened 2025-05-04
gcdb: display: Print framebuffer address on panel initialization

ItsVixano:
gcdb: display: Print framebuffer address on panel initialization

Matches cont-splash logging.

Change-Id: I62bbeef05b317dbd68af1924080bc32f3b210cfc

### DIFF
--- a/dev/gcdb/display/gcdb_display.c
+++ b/dev/gcdb/display/gcdb_display.c
@@ -599,14 +599,19 @@ int gcdb_display_init(const char *panel_name, uint32_t rev, void *base)
 		panel.panel_info.dfps.dfps_fb_base = base;
 		base += DFPS_PLL_CODES_SIZE;
 		panel.fb.base = base;
-		dprintf(SPEW, "dfps base=0x%p,d, fb_base=0x%p!\n",
-				panel.panel_info.dfps.dfps_fb_base, base);
-
 		panel.fb.width =  panel.panel_info.xres;
 		panel.fb.height =  panel.panel_info.yres;
 		panel.fb.stride =  panel.panel_info.xres;
 		panel.fb.bpp =  panel.panel_info.bpp;
 		panel.fb.format = panel.panel_info.mipi.dst_format;
+
+		dprintf(INFO, "DSI panel initialized: base: %p, stride: %d, "
+			"img: %dx%d, format: %#x, bpp: %d\n",
+			panel.fb.base, (panel.fb.stride * panel.fb.bpp / 8),
+			panel.fb.width, panel.fb.height,
+			panel.fb.format, panel.fb.bpp
+		);
+
 	} else if (pan_type == PANEL_TYPE_EDP) {
 		mdss_edp_panel_init(&(panel.panel_info));
 		/* prepare func is set up at edp_panel_init */


TravMurav:
Hi! Could you help me understand how is this different from reading those values in lk2nd log?

```
$ fastboot oem log && fastboot get_staged /dev/stdout
(...)
[140] MDP continuous splash detected: pipe RGB_0, base: 0xf100000, stride: 1440, src: 480x854 (0,0), img: 480x854, out: 480x854 (0,0), format: 0x2243f (bpp: 3)
(...)
```

TravMurav:
In addition, if you want to add simplefb node to mainline, is lk2nd's automatic fixup enabled via `lk2nd.pass-simplefb` cmdline param deficient? That should automatically generate the required dt nodes when booting, based on the lk2nd's detected mdp settings

ItsVixano:
> Hi! Could you help me understand how is this different from reading those values in lk2nd log?
> 
> ```
> $ fastboot oem log && fastboot get_staged /dev/stdout
> (...)
> [140] MDP continuous splash detected: pipe RGB_0, base: 0xf100000, stride: 1440, src: 480x854 (0,0), img: 480x854, out: 480x854 (0,0), format: 0x2243f (bpp: 3)
> (...)
> ```

Unfortunately i didn't had such log (unless is due on how i configured it for my device?)

```
TOOLCHAIN_PREFIX := arm-none-eabi-
PROJECT := lk2nd-msm8226
LK2ND_BUNDLE_DTB := msm8926-sony-eagle.dtb
LK2ND_DISPLAY := innolux_qhd_video
```

TravMurav:
> LK2ND_DISPLAY := innolux_qhd_video

Does that device disable the display when it leaves oem bootloader? usually lk2nd picks up the already set-up mdp, which is what that log is for, with a panel-specific driver lk2nd will decide on the settings on it's own. `lk2nd.pass-simplefb` should work regardless of the display driver though.

ItsVixano:
> > LK2ND_DISPLAY := innolux_qhd_video
> 
> Does that device disable the display when it leaves oem bootloader? usually lk2nd picks up the already set-up mdp, which is what that log is for, with a panel-specific driver lk2nd will decide on the settings on it's own. `lk2nd.pass-simplefb` should work regardless of the display driver though.

From what i've seen, the OEM bootloader mode doesn't initialize the screen (but it shows a blue LED indicating the device is in said mode).

I've tried already to boot lk2nd without that var, with no avail. I could send some logs later this day if required.

TravMurav:
I see. Then I'd suggest to add similar logging message into the panel init driver. I think you'd just need to promote this message to INFO and extend it to be a similar format to cont-splash one. I think this is simpler and more obvious to keep the debugging information like this in the same place (lk2nd log) than to add a debug command just to probe static information.

https://github.com/msm8916-mainline/lk2nd/blob/a4bfd07abc31a6b2ad2c0a797ef80de629171456/dev/gcdb/display/gcdb_display.c#L602-L603

https://github.com/msm8916-mainline/lk2nd/blob/a4bfd07abc31a6b2ad2c0a797ef80de629171456/lk2nd/display/cont-splash/pipe.c#L94-L101

ItsVixano:
It should be enough i think

```
...
[140] DSI panel initialized: base: 0x3201000, stride: 1620, img: 540x960, format: 0x3, bpp: 24
...
```

TravMurav:
Thanks!

PR #531, opened 2025-05-12
lk2nd: dts: msm8916: Add EE Harrier Mini (harrier_mini)

halal-beef:
Based on the msm8916 board.

Currently a draft as of right now as the volume keys are inverted and defining the correct GPIOs for them (that hasn't been pushed here for the following reason) seem to not work, the volume up key will work as expected but the rest stop responding.

### DIFF
--- a/Documentation/devices.md
+++ b/Documentation/devices.md
@@ -12,6 +12,7 @@
 - BQ Aquaris M5 - piccolo
 - BQ Aquaris X5 - paella, picmt
 - DragonBoard 410c - apq8016-sbc
+- EE Harrier Mini - harrier-mini
 - GPLUS FL8005A
 - HTC One M8s - m8qlul (quirky - see comment in `lk2nd/device/dts/msm8916/msm8939-htc-m8qlul.dts`)
 - Huawei Ascend G7 - G7-L01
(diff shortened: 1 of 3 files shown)


lk2nd/device/dts/msm8916/msm8916-ee-harrier_mini.dts
```diff
@@ -0,0 +1,41 @@
...
+
+	soc: soc@0 {
+	};
+};
+
+&lk2nd {
+	model = "EE Harrier Mini";
+	compatible = "ee,harrier_mini";
+
+	lk2nd,dtb-files = "msm8916-ee-harrier_mini";
```
TravMurav, line 40: These compatible and dtb filename should match mainline linux, and upstream won't accept underscores.

halal-beef:
Current progress in terms of keys, defining volume up alone kind of works, itll go up and down at the same time and the volume down key will still go upwards in the ui, upon defining the gpio for volume down, the volume down and power key will stop responding (though in further testing the power key will respond if pressed alongside with the volume up key)

current definition so far is:

```
        gpio-keys {
                compatible = "gpio-keys";

                volume-down {
                        lk2nd,code = <KEY_VOLUMEDOWN>;
                        gpios = <&tlmm 107 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
                };

                volume-up {
                        lk2nd,code = <KEY_VOLUMEUP>;
                        gpios = <&pmic_pon GPIO_PMIC_RESIN (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
                };
        };
```

if this looks incorrect, please let me know as im honestly clueless

halal-beef:
Fixed! thanks to @bengris32

PR #532, opened 2025-05-13
lk2nd: keys: Conditionally swap volume keys

ItsVixano:
Despite defining the correct GPIO values for the volume keys, there are cases where those values won't apply properly. Let's introduce a DT property that swaps the default keycode for these cases.

Change-Id: Ibfb5bba5d8a78ff71bada54d06d8c469a20d9949


ItsVixano:
Not sure if there are better ways to handle it.

Usage:

```
        gpio-keys {
                compatible = "gpio-keys";
                lk2nd,swap-volume-keys;
        };
```

ItsVixano:
No longer needed, thanks for the review regardless!

PR #539, opened 2025-05-22
dts: msm8960: add Casio G'zOne (CA-201L)

peremen:
The device itself is quite old: https://wiki.postmarketos.org/wiki/Casio_G%27zOne_(casio-ca201l)

The cmdline set by the stock bootloader does not contain the device model/codename, so relying on the other parameter instead. Although the dts is not yet included in the mainline kernel, the filename will likely be as indicated here.

Caveats: the panel is mounted upside down, and seems like the color format is different from default. So display is incorrect at this moment, but bootloader functionality is tested. But keys are working as expected.

### DIFF
--- a/lk2nd/device/dts/msm8960/bundle.dts
+++ b/lk2nd/device/dts/msm8960/bundle.dts
@@ -116,4 +116,26 @@
 			};
 		};
 	};
+	casio-ca201l {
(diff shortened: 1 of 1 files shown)


TravMurav:
Thanks!

PR #540, opened 2025-05-23
lk2nd: dts: msm8953: Add Billion Capture+ (rimob)

cristianc:
Add support for msm8953 based billion-rimob.

Looking at downstream dts (no kernel sources, extracted dtb from running system), there seem to be  a few panel variants. Though, they seem to be part of the reference board (so I've not added panel node).


TravMurav:
Thanks!

PR #560, opened 2025-07-05
fix Moto `sanders` DTB

### LINKED ISSUE #559: Moto G5s Plus (msm8953) fails to boot with lk2nd – "Error: failed to load kernel"
Hello,
I’m trying to use lk2nd on my Moto G5s Plus (msm8953), but it fails to boot the kernel. When I power on the device, I get the following error on screen:

Start Up Failed:

Your device didn't start up successfully.
Use the Software Repair Assistant on computer to repair your device.
Connect your device to your computer to get the Software Repair Assistant.

AP Fastboot Flash Mode (Secure)
Error: failed to load kernel!
Fastboot Reason: Fall-through from charger boot mode
USB connected

This issue happens right after flashing lk2nd as the boot

Image

playday3008:
Closes: #559 

- Stock `boot.img` of Motorola G5s Plus (`sanders`) have:
  - compressed QCDT via LZ4
    - added documentation for repacking stock image
  - vendor specific `qcom,board-id` values
    - can be handled by `lk2nd`, by splitting dts into separate dt's
- Wrong format of DT was used for Motorola G5s Plus (`sanders`) (this closes #559)


playday3008:
This PR is complete (in my opinion)

stephan-gh:
Can you separate the new features into a separate PR so that there is only the actual fix for #559 in this PR?

I appreciate the work you did to enable booting downstream boot images, but I'm afraid I'm not entirely sure yet if we can accept that change. Generally, booting downstream with lk2nd is only supported on a "least effort" basis. It's good if it works, but the problem is that there are tons of subtly incompatible vendor modifications to the boot image, DTB selection or QCDT format and it would be an extreme effort to support all of those. With the 150+ devices we support nowadays, there is also a significant risk that implementing the modifications of one vendor breaks booting on the devices from another vendor.

In this case for example I'm concerned that changing the `struct dt_table` definition (splitting up `version` into multiple fields) might break booting on some other devices. There is also quite some maintenance overhead to add a copy of the lz4 library just to decompress the DTBs for downstream on some Motorola devices...

What is the use case you have in mind for these changes? How often would one really boot downstream with lk2nd on these devices?

playday3008:
> significant risk that implementing the modifications of one vendor breaks booting on the devices from another vendor.

True, but.
LZ4 decompression only occurs when DT starts with `LZ4F_MAGICNUMBER`, which is impossible for normal DT, no matter how broken vendor made it, when LZ4 does not occur, nothing changes in code flow, as any changes are guarded by `magic` check.
Motorola specific check only returns true, when `mmi_unit_info` is not null, which can only be non-null on Motorola devices. So again, code flow does not change for non Motorola devices.

> In this case for example I'm concerned that changing the `struct dt_table` definition (splitting up `version` into multiple fields) might break booting on some other devices.

About splitting `struct dt_table`, this might break something only when used on big endian QCDT, which I'm not even sure exists, as it's structure inside a file, and not inside a code.

>  There is also quite some maintenance overhead to add a copy of the lz4 library just to decompress the DTBs for downstream on some Motorola devices...

Adding lz4 increases the size of the final binary for everyone, and a non-significant amount of people actually needs this. But I'm not sure that we know about all instances where adding lz4 would actually resolve a problem that somebody had. So, I think that adding lz4 have still more advantages than disadvantages. Also, I didn't add a specific commit of lz4, but just tag 1.10.0 which can be observed in `lz4.h`, and changes that had to be made to make it work in lk2nd are in a separate commit, so we know what has to be done to lz4 if we wanted to update it someday.

https://github.com/msm8916-mainline/lk2nd/blob/20c6ebbb2e3004005c5daf3bdc306a675770e61a/lib/lz4/lz4.h#L131-L133

> What is the use case you have in mind for these changes? How often would one really boot downstream with lk2nd on these devices?

It's very useful to have "dual boot" with lk2nd, because we can have stock kernel in phone, and WIP mainline from sdcard. Which allows anybody to still have a working phone while working on mainline kernel to support said phone. (Not only mainline, maybe some custom kernel from stock, IDK, I just provided an example that relevant to me)

> Can you separate the new features into a separate PR so that there is only the actual fix for https://github.com/msm8916-mainline/lk2nd/issues/559 in this PR?

If you really-really want only this change, just cherry-pick d2d0b98cc10d82dc7db8495ae17b55d8bd961546 commit into main, this will also automatically close the issue. But please reconsider.

playday3008:
> That is true, but the problem I meant was that there are multiple vendors who modified this structure differently. I'm pretty sure we've had some weird QCDT table from OPPO before that we decided against supporting. If we start adding all sorts of different formats it will become really difficult to maintain (as in: required effort, but also making sure none of the many vendor-specific modifications break when making changes).

I've had an idea of instead of changing `struct dt_table` structure, just do `table-> version & 0xFF` everywhere, but changing struct looked less painful.

> The `is_stock_match()` function you add for DTB selection could perhaps reuse the `lk2nd_dt_override` functionality. It does more or less the same already if you would fill it from motorola-unit-info.

Actually, I wish I knew how to utilize that functionality, I couldn't find any example or anything that will explain how and with what `lk2nd_dt_override` is getting populated.

> Couldn't you also repack the downstream image to fit the expectations of lk2nd better? If you decompress the LZ4 QCDT, we could already drop that extra dependency. That would be easy to do with a simple script.

I could repack it in the way that will not require lz4 dependency in lk2nd, but QCDT structure will still be `extended`, that needs to be accounted for.

> You could probably also drop the extra field from the `struct dt_table` while repacking the boot image. Given that lk2nd boots with just `QCDTBS +=`, apparently Motorola's bootloader supports the non-customized QCDT format as well? So the repacked/differently built boot images would likely also work with the stock bootloader?

You mean changing QCDT structure itself in `boot.img`? Like removing `name` field from it? IDK how painlessly do that (means automate it, I know how to do it manually)

<details><summary>Example</summary>
<p>

From:
image

To:
image


</p>
</details> 

> lk2nd is already extremely complex to maintain with so many supported devices

Yeah, I wanted to completely rewrite it someday in Zig, and I gave up on 1%.

playday3008:
@stephan-gh That's better?

barni2000:
Other motos are not splitted by revisions maybe it is also not necessary for sanders.

playday3008:
> Other motos are not splitted by revisions maybe it is also not necessary for sanders.

It's necessary for a stock boot.img, for non-stock it doesn't matter.

playday3008:
As lk2nd overriding feature only works with dts where and/or `qcom,board-id`, `qcom,msm-id`, `qcom,pmic-id` are single, here's code below that checks for it.
https://github.com/msm8916-mainline/lk2nd/blob/0cf77849fccdf059e6dfc5e2f4308702fa20666b/platform/msm_shared/dev_tree.c#L970-L974
And overriding feature is required for stock boot.img to boot successfully.
Overriding feature requires complete rewrite to be able to handle overriding without splitting dtb's.

raphasil:
Hi everyone, thanks for the hard work on bringing support for the Sanders! 

I've been testing this PR on my Moto G5S Plus (XT1802, `retbr`), but I'm consistently running into a `dtb not found` error when trying to boot. 

Here are the relevant parts of my `fastboot oem log`:

`[130] Detected device: Motorola Moto G5s Plus (sanders) (compatible: motorola,sanders)`
`[140] Unexpected DTB selected by bootloader, need to override`
`[150] Motorola unit info v3: prod_id=0x004b, rev=0x8400, serial=0xef1fa26900000000, machine='', barcode='0044505380', carrier='retbr', baseband='', device='sanders', radio='Brazil' (0x0), powerup_reason=0x00004000`
`...`
`[13300] DTB offset is incorrect, kernel image does not have appended DTB`

**What I tried locally:**
I noticed my device has `prod_id=0x004b` and `rev=0x8400`. I tried locally editing the `msm8953-motorola-sanders-p*.dts` files to append my board ID like this:
`qcom,board-id = <0x4b 0>, <0x4b 0x8400>, <0x4c 0>, <8 0>;`

However, even after a clean build of `lk2nd-msm8953` and booting it, I still get the exact same `DTB offset is incorrect` error when trying to pass a mainline kernel image (without an appended DTB). 

Is there a specific way the `0x004b` hardware revision needs to be mapped in the code, or am I missing a step to force lk2nd to pick up the correct DTB? 

Any guidance or patches I can test would be greatly appreciated. Let me know if you need full logs or testing!

TravMurav:
>  [13300] DTB offset is incorrect, kernel image does not have appended DTB

> trying to pass a mainline kernel image (without an appended DTB).

I'm confused. You say you're booting an android boot image \*\*without an appended DTB\*\* and you get an error that your kernel image \*\*doesn't have appended DTB\*\*...

What are you trying to do?

raphasil:
Ah, sorry for the confusion! Let me clarify my goal.

I am trying to boot a mainline/postmarketOS kernel. My understanding was that `lk2nd` should provide the correct DTB from its own internal package based on the device matching (`prod_id=0x004b`), which is why I intentionally passed a clean boot image without an appended DTB.

The issue seems to be that `lk2nd` is failing to match my device with any of its internal DTBs (even after I tried locally patching the `qcom,board-id` with `<0x4b 0>` in the `.dts` files and booting that new `lk.bin-dtb`). 

Because it fails to find an internal match, it falls back to looking for an appended DTB in my boot image, which triggers that error since I didn't append one.

My main question is: how can I get `lk2nd` to successfully match and use its *internal* DTB for this `0x004b` hardware revision? Did I patch the `board-id` incorrectly, or is there another place in the code I need to modify for this specific revision?

TravMurav:
> My understanding was that lk2nd should provide the correct DTB from its own internal package

lk2nd doesn't include board dtbs for any devices it supports, lk2nd's dtb is it's own format only used internally and useless for linux or any other os.

raphasil:
🫣. Thank you so much for clearing that up! That was a massive fundamental misunderstanding on my part. 

I was under the impression that `lk2nd` provided the mainline DTB to the kernel directly from its own build, which is why I was deliberately trying to boot a clean kernel without an appended DTB. 

Now it makes perfect sense!

I will go back to appending the proper mainline postmarketOS DTB to my `vmlinuz` and boot that. Sorry for the noise, and thanks again for the quick reality check.

PR #561, opened 2025-07-06
lk2nd: Run lk2nd_init() late again, after display initialization

wonderfulShrineMaidenOfParadise:
This is needed for the panel detection stuff to work with lk1st since the panel is not known yet until the display is turned on.

### DIFF
--- a/app/aboot/aboot.c
+++ b/app/aboot/aboot.c
@@ -5535,10 +5535,6 @@ void aboot_init(const struct app_descriptor *app)
 		}
 	}
 
-#if WITH_LK2ND
-	lk2nd_init();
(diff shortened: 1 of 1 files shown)


wonderfulShrineMaidenOfParadise:
This will partially undo 4d2d95805d15b3f743acaea456bc13f91168d6a0

stephan-gh:
We can't do this since it would break `lk2nd/device/2nd/spi-display.c`, which needs to run before display initialization. Since lk1st does not actually support panel detection (you need to specify the panel to use on the make command line), I think we should add some code to populate the panel name earlier.

PR #564, opened 2025-07-27
dts: msm8939-samsung: fix qcom,board-id typo

bolilingmeng89006:


lk2nd/device/dts/msm8916/msm8939-samsung.dts
```diff
@@ -7,7 +7,7 @@
 
 / {
 	qcom,msm-id = <QCOM_ID_MSM8939 0>;
-	qcom,board-id = <0xEF08FF01 1>;
+	qcom,board-id = <0xEF08FF1 1>;
```
TravMurav, line 10: Was a bit confused by this but apparently samsung either actually messed this up compared to most other boards or most other boards are weirdly laid out given the comment seems to define right bits... What a mess

https://github.com/msm8916-mainline/linux-downstream/blob/cc5ef89ef8cae990a812f331046e1d3b9a150ef6/arch/arm/boot/dts/samsung/msm8939/msm8939-sec-a7-eur-r06.dts#L19-L27

TravMurav:
Thanks!

PR #568, opened 2025-08-10
Add menu option to skip lk2nd_boot

BetaRays:
This can be useful when dual-booting Android on the internal storage and postmarketOS on an SD card without physically removing it or using fastboot boot with the flashed Android boot image. This can be useful considering Android is required to charge the battery (safely) for some devices.

### DIFF
--- a/app/aboot/aboot.c
+++ b/app/aboot/aboot.c
@@ -4680,6 +4680,27 @@ void cmd_continue(const char *arg, void *data, unsigned sz)
 	}
 }
 
+#if WITH_LK2ND_BOOT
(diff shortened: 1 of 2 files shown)


Biswa96:
The reboot option already boots to Android.

BetaRays:
> The reboot option already boots to Android.
It doesn’t seem to for me. It goes back to lk2nd and if I do nothing, boots what’s on the SD card.

Biswa96:
Understood. I've not flahsed lk2nd in one device and it reboots to Android.

PR #570, opened 2025-08-17
Building related enhancements, and try to match `compatible` on kernel dtbs

me-cafebabe:
For example, in `dev_tree_appended()` path, `platform_dt_absolute_match()` filters the dtbs with matching ids, and this `compatible` matching logic will prefer the first of these filtered dtbs with matching `compatible`

The `compatible` matching logic is useful when the user wants to boot from `boot.img` containing mainline kernel and multiple devices' dtb (which shares `qcom,board-id`)

### DIFF
--- a/lk2nd/device/dts/msm8953/msm8953-xiaomi-common.dts
+++ b/lk2nd/device/dts/msm8953/msm8953-xiaomi-common.dts
@@ -36,7 +36,7 @@
 	sakura {
 		/* Xiaomi Redmi 6 Pro (sakura) and Xiaomi Mi A2 Lite (daisy) are software compatible */
 		model = "Xiaomi Redmi 6 Pro";
-		compatible = "xiaomi,sakura";
+		compatible = "xiaomi,sakura", "xiaomi,daisy";
 		lk2nd,match-bootloader = "*DAISY*";
 
 		lk2nd,dtb-files = "msm8953-xiaomi-daisy";


lk2nd/device/dts/msm8953/msm8953-xiaomi-common.dts
```diff
@@ -37,6 +37,7 @@
 		/* Xiaomi Redmi 6 Pro (sakura) and Xiaomi Mi A2 Lite (daisy) are software compatible */
 		model = "Xiaomi Redmi 6 Pro";
 		compatible = "xiaomi,sakura";
+		dtb_compatible = "xiaomi,daisy";
```
TravMurav, line 40: Can we just put daisy into the array of compatible values? This would indeed make the matching logic a bit more complicated since one'd need to iterate over all compatibles, but the compatible in the line above was kinda meant for that already. I'd strongly prefer to not have two props that do the same thing.

TravMurav (commented):
Hi! First of all, I'm sorry for not replying to this earlier.

Overall the idea sounds fine to me, but see one comment below. Aside from that:

> Fixes linking errors when building an image with a more serious
toolchain

Could you help me understand in which case you get linking errors? i.e. what's the workflow that fails for you without adding the ifdefs?

app/aboot/aboot.c
```diff
@@ -1487,7 +1489,9 @@ static void verify_signed_bootimg(uint32_t bootimg_addr, uint32_t bootimg_size)
 #if !VERIFIED_BOOT
 	if(device.is_tampered)
 	{
+	#if !ABOOT_STANDALONE
```
Biswa96, line 1492: This was added in d90cfca1a77ad81ee0fe9c6dd0f2e85a9a276ec7 commit.

lk2nd/device/device.c
```diff
@@ -106,12 +114,16 @@ static void parse_dtb(const void *dtb)
...
-	val = fdt_getprop(dtb, node, "compatible", &len);
+	val = fdt_stringlist_get(dtb, node, "compatible", 0, &len);
 	if (val && len > 0)
 		lk2nd_dev.compatible = strndup(val, len);
 	else
 		dprintf(CRITICAL, "Failed to read 'compatible': %d\n", len);
 
+	val = fdt_stringlist_get(dtb, node, "compatible", 1, &len);
+	if (val && len > 0)
+		lk2nd_dev.dtb_compatible = strndup(val, len);
```
TravMurav, line 125: While I see how it's tempting to just hack things together to make your single usecase work, I think this makes the code more confusing since it's hard to understand what is the meaning of which prop, and it would violate expectations one would have of DT matching.

I.e. you're currently creating two seemingly unrelated fields in the struct, but they now suddenly are an "array" of at most two fields, used for dtb mathing. Except contrary to the usual dtb logic, if the second (less specific, fallback) compatible is present, it would be preferred, even if a more specific compatible is available.

I'd strongly prefer this to align to common dt logic, i.e. I'd expect the `lk2nd_dev.compatible` be changed to something like `lk2nd_dev.compatibles` with a setup similar to `.dtbfiles`, and your selection logic to iterate over all of them to find the best pick.

This will unfortunately require a slight refactoring where all uses of `lk2nd_dev.compatible` will have to be changed to `lk2nd_dev.compatibles[0]`, with a bit more care taken with assignments to create a correct pointer structure.

What would I do to refactor for this implementation:

1. Study the usage of `lk2nd_dev.compatible`, I believe it's mainly used in lk2nd device module for picking the dtb in case of lk1st->lk2nd boot. Based on the usage, I'd change the variable to a `char**` like `dtbfiles`, use `lkfdt_stringlist_get_all` to assign it in the parser, and pointer to a string literal where it's assigned at compile time/from cmdline. Then refactor all read usages like suggested above. (I'd put it as a separate refactor commit)
2. Based on that work of refactoring lk2nd compatible list, I'd create a new function `int lk2nd_is_compatible(char *compatible)` that would compare a provided compatible against a list and return either -1 (no match) or an index of the provided compatible (i.e. sakura=0, daisy=1), which could be used to rank dtbs in a container by best match (i.e. loop over them and find the lowest index - that's the one to boot). This would allow you to add sakura.dtb later if you have to and keep the logic working. This funciton could also be reusable later if we ever want to implement other system containers like BLS#2 (UKI) for example.

PR #572, opened 2025-08-19
WIP: Moto G7 (river) Support

ElliotLurie:
Simple device tree files that *should* enable support for the Motorola Moto G7, but the boot loops. 

If anyone has access to this device with a serial cable, that would be very helpful for debugging.


lk2nd/device/dts/msm8953/sdm632-motorola-river.dts
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
TravMurav, line 1:
```suggestion
// SPDX-License-Identifier: BSD-3-Clause
```

if possible, we try to align to the modern Linux practice for dts.

lk2nd/device/dts/msm8953/sdm632-motorola-river.dts
```diff
@@ -0,0 +1,29 @@
...
+
+/dts-v1/;
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+#include <motorola-carrier-channel-ids.dtsi>
+
+/ {
+	model = "Qualcomm Technologies, Inc. SDM632 + PMI632 MTP S3";
```
TravMurav, line 10: which of the lines in `/ {}` are actually required for the device to boot?

PR #579, opened 2025-08-28
Add support for setting OS version and patch level and add new Palm Phone device

xc-racer99:
The Palm phone bootloader has a few quirks, so need to add some code to set the OS version and patch level in the generated lk2nd image.  Then add support for the device itself.  It only has a single power button, so use the short/long press control for it.

### DIFF
--- a/lk2nd/device/dts/msm8952/msm8940-palm-pepito.dts
+++ b/lk2nd/device/dts/msm8952/msm8940-palm-pepito.dts
@@ -1,0 +1,31 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/*
+ * There are a few quirks to be aware of
+ *
+ * - Fastboot is not available on this device, lk2nd must be flashed via EDL.
+ *   Some EDL programs (or versions thereof) do not work properly with available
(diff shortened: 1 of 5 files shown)


lk2nd/device/dts/msm8952/msm8940-palm-pepito.dts
```diff
@@ -0,0 +1,31 @@
...
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8940 0>;
+	qcom,board-id = <0xaa000008 0x00>;
+};
+
+&lk2nd {
+	model = "Palm Phone (pepito)";
+	compatible = "palm,pepito";
+
+	lk2nd,dtb-files = "qcom,msm8940-mtp", "qcom,msm8940", "qcom,mtp";
```
TravMurav, line 28: This line should define a dtb filename for this device, not a list of compatibles (especially downstream ones), which would be used for booting a generic linux image with extlinux.conf. If you have some work to run linux on this device, please put the name of your dtb here 

https://github.com/msm8916-mainline/lk2nd/blob/4801a1e9e0c0f26c99dce2df9f5dc970fa2cf8e5/Documentation/dt-bindings.md?plain=1#L67-L69

TravMurav (commented):
Hi! Sorry for the delay. The changes look good to me apart from one comment below

PR #590, opened 2025-09-29
dts: msm8226: add support for LG Wine Smart D486 (vfp)

egor835:
This device doesn't have fastboot by default, so i thought it would be useful

IMG_20250929_125439

### DIFF
--- a/lk2nd/device/dts/msm8226/msm8926-lg-vfp.dts
+++ b/lk2nd/device/dts/msm8226/msm8926-lg-vfp.dts
@@ -4,7 +4,7 @@
 #include <lk2nd.dtsi>
 
 / {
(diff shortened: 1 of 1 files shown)


lk2nd/device/dts/msm8226/msm8926-lg-vfp.dts
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton32.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <200 0>;
```
TravMurav, line 7:
```suggestion
	qcom,msm-id = <QCOM_ID_MSM8926 0>;
```

PR #596, opened 2025-10-18
[DRAFT] implement multiboot menu based on extlinux

hitechshell:
test1
test2


hitechshell:
Hmm, okay. Unfortunately, porting u-boot will take significantly more time, but I understand your point. Perhaps when I have a little more free time, I'll implement the lk2nd-like UI (it looks really convenient) for u-boot.

Perhaps we should close this PR as not-planned?

Issue #597, opened 2025-10-18
need a some way for distinguish sdm439-xiaomi-pine and sdm439-nokia-panther


hitechshell:
both device has same `qcom,msm-id` and `qcom,board-id`

hitechshell:
yes, but lk2nd currently show me "nokia panther" on my xiaomi-pine.
if i reorder `lk2nd/device/dts/msm8952/rules.mk` than it will show "Redmi 7A".

exkc:
humm maybe add match resolution and match cmdline to lk2nd

barni2000:
I think match panel not checking the whole panel name

barni2000:
Here is a workaround:
`make TOOLCHAIN_PREFIX=arm-none-eabi- LK2ND_ADTBS='sdm439-xiaomi-pine.dtb' lk2nd-msm8952`
I am trying to find a solution but it seems `lk2nd,match-panel` is not working and adding `lk2nd,match-bootloader` to nokia is also not working.

barni2000:
The last released image also working https://github.com/msm8916-mainline/lk2nd/releases/tag/21.0

PR #609, opened 2025-11-22
dts: msm8952: Merge nokia-panther with xiaomi-pine

barni2000:
nokia-panther and xiaomi-pine have same board-id so they should be in the same devicetree to be able to lk2nd,match-panel select the correct device.

Fixes: #597


TravMurav:
Thanks!

PR #617, opened 2025-12-16
dts: msm8939: asus-z00t: add a fixed regulator for sdcard

erikas9987:
Add an external fixed regulator for powering the SD card slot. This fixes an issue where lk2nd would fail to communicate with SD card slot because no suitable voltages are available:
```
[140] Error: Command timeout error
[140] Failure getting OCR response from MMC Card
[140] MMC card failed to respond, try for SD card
[150] Error: Command timeout error
[150] The response for CMD8 does not match the supplied value
[150] Failed to initialize SD card
[160] Failed detecting MMC/SDC @ slot2
[160] SD card MMC is unavailable.
```
This would make booting the kernel from an SD card impossible.

### DIFF
--- a/lk2nd/device/dts/msm8916/msm8939-asus-z00t.dts
+++ b/lk2nd/device/dts/msm8916/msm8939-asus-z00t.dts
@@ -4,7 +4,7 @@
 #include <lk2nd.dtsi>
 
 / {
-	qcom,msm-id = <QCOM_ID_MSM8939 0>, <QCOM_ID_MSM8939 30000>;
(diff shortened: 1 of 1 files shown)


erikas9987:
Only [this commit](https://github.com/msm8916-mainline/lk2nd/pull/617/commits/405dee7a548ce720bacd234a8ca63806ee869e50) is relevant. Not sure how others ended up here...

TravMurav:
The changes look good to me but this can't be merged due to history of your branch being weird...

Please rebase the two commits you need to merge on top of the master branch, there is a merge commit there currently which breaks merge at least...

For the future, you should probably use feature branches for PRs instead of the main branch to avoid issues like this...

TravMurav:
Thanks!

PR #618, opened 2025-12-16
dts: msm8953: Add support for Meizu M15(m1871)

PanicNotPanic:
增加魅族M15(m1871)机型 dts

### DIFF
--- a/Documentation/devices.md
+++ b/Documentation/devices.md
@@ -9,6 +9,7 @@
 - Asus Zenfone 2 Laser (1080p) - Z00T
 - Asus Zenfone 2 Laser ZE500KL - Z00E
 - Asus Zenfone Max ZC550KL (2016) - Z010D
+- Asus ZenPad 8.0 - P024
 - BQ Aquaris M5 - piccolo
 - BQ Aquaris X5 - paella, picmt
 - DragonBoard 410c - apq8016-sbc
--- a/lk2nd/device/dts/msm8916/msm8916-mtp.dts
+++ b/lk2nd/device/dts/msm8916/msm8916-mtp.dts
@@ -37,6 +37,34 @@
 		};
 	};
 
+	asus-p024 {
+		model = "Asus ZenPad 8.0";
+		compatible = "asus,p024";
+		lk2nd,match-panel;
+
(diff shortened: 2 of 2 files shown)


lk2nd/device/dts/msm8953/msm8953-meizu-m1871.dts
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
TravMurav: We prefer BSD-3-Clause for new submissions, unless this is inherited from somewhere.

lk2nd/device/dts/msm8953/msm8953-meizu-m1871.dts
```diff
@@ -0,0 +1,51 @@
...
+		};
+
+		qcom,mdss_dsi_sharp_1080p_video {
+			compatible = "meizu,m1871-sharp";
+		};
+
+		qcom,mdss_dsi_djn_1080p_video {
+			compatible = "meizu,m1871-djn";
+		};
+	};
```
TravMurav: While we usually try to list all panels possible in hardware, this list somehow looks weird to me... For example, there is a mix of cmd and video mode panels (I'd expect only one or another to be used), also there is the adv7533 which is a bridge chip, which I'd not expect to find in a phone...

Would you mind checking with caf kernel (or oem kernel if you have sources) which panels are actually added by the oem? and remove those which were left over from qualcomm reference designs.

PanicNotPanic: The device tree I extracted from OEM firmware is located in this [link](https://github.com/PanicNotPanic/meizu-m1871-device-tree/blob/main/device.dts). I am not sure if the Qualcomm reference design code exists in it, but it does have multiple cmd mode panels and video mode panels

我从oem firmware中提取出的 device tree 在这个[链接](https://github.com/PanicNotPanic/meizu-m1871-device-tree/blob/main/device.dts) , 我不确定高通参考设计代码是否在其中存在, 但他确实存在多个cmd模式面板和video模式面板

TravMurav:
There is still one unresolved review comment. Other than that it would be preferable if you can squash all your changes into your first commit.

PR #619, opened 2025-12-25
dts: msm8996: Add support for LG G6

AKAsaliza:


lk2nd/device/dts/msm8996/msm8996-lg-h990.dts
```diff
@@ -4,7 +4,8 @@
 #include <lk2nd.dtsi>
 
 / {
-	qcom,msm-id = <QCOM_ID_MSM8996 0x10000>;
+	model = "LG V20";	
+	qcom,msm-id = <QCOM_ID_MSM8996 0x30001>;
```
wonderfulShrineMaidenOfParadise: What's the point to modify h990 here?

AKAsaliza: Fixed missing model name display and updated qcom/msm-id for the H990 device.

TravMurav: Would you please separate the change to two commits then, one "fixing" h990 and the one adding h870?

PR #620, opened 2025-12-25
dts: msm8952: Add Motorola Moto G5 (cedric) panel support

Hacker1245:
Added tianma panel support only as it's the one I could test.

### DIFF
--- a/lk2nd/device/dts/msm8952/msm8937-mtp.dts
+++ b/lk2nd/device/dts/msm8952/msm8937-mtp.dts
@@ -15,6 +15,18 @@
 		lk2nd,match-device = "cedric";
 
 		lk2nd,dtb-files = "msm8937-motorola-cedric";
+		
+		panel {
+			compatible = "motorola,cedric-panel", "lk2nd,panel";
+			
+			qcom,mdss_dsi_mot_tianma_497_1080p_video_v0 {
+				compatible = "motorola,cedric-nt35596-tianma";
+			};
+			
+			qcom,mdss_dsi_mot_inx_497_1080p_video_v0 {
+				compatible = "motorola,cedric-nt35596-inx";
+			};
+		};
 	};
 
 	hannah {


lk2nd/device/dts/msm8952/msm8937-mtp.dts
```diff
@@ -15,6 +15,14 @@
 		lk2nd,match-device = "cedric";
```
wonderfulShrineMaidenOfParadise:
```suggestion
```

Try to use only one of `lk2nd,match-*`

lk2nd/device/dts/msm8952/msm8937-mtp.dts
```diff
@@ -15,6 +15,14 @@
...
 
 		lk2nd,dtb-files = "msm8937-motorola-cedric";
+		
+		lk2nd,match-panel;
+  		panel {
+   			compatible = "motorola,cedric-panel", "lk2nd,panel";
+   			qcom,mdss_dsi_mot_tianma_497_1080p_video_v0 {
+    				compatible = "motorola,cedric-nt35596-tianma";
+   			};
+  		};
```
wonderfulShrineMaidenOfParadise:
```suggestion

		lk2nd,match-panel;

		panel {
			compatible = "motorola,cedric-panel", "lk2nd,panel";

			qcom,mdss_dsi_mot_tianma_497_1080p_video_v0 {
				compatible = "motorola,cedric-nt35596-tianma";
			};
		};
```

- Drop trailing white spaces.
- Use tabs for indents.
- Some newline nitpicks by the way.

lk2nd/device/dts/msm8952/msm8937-mtp.dts
```diff
@@ -12,9 +12,22 @@
 	cedric {
 		model = "Motorola Moto G5 (cedric)";
 		compatible = "motorola,cedric";
-		lk2nd,match-device = "cedric";
```
TravMurav: I'm confused. I'd expect device model matching be way more reliable than panel matching (which should only be used for the last resort, when nothing else can be used.) Why are you switching the matching here?

wonderfulShrineMaidenOfParadise: Oops. I suggested to use either of them instead of both in earlier comments.

TravMurav:
> Do I squash the commits into one?

Yes, please squash the commits

PR #621, opened 2025-12-28
dts: msm8953: Add support for Meizu M15(m1871)

PanicNotPanic:


lk2nd/device/dts/msm8953/rules.mk
```diff
@@ -26,6 +26,7 @@ ADTBS += \
 	$(LOCAL_DIR)/sdm632-fairphone-fp3.dtb \
 	$(LOCAL_DIR)/sdm632-mtp.dtb \
 	$(LOCAL_DIR)/sdm632-mtp-3.dtb \
+	$(LOCAL_DIR)/msm8953-meizu-m1871.dtb \
```
TravMurav: this is still not sorted properly

PR #629, opened 2026-01-08
dts: msm8974: victara: use qcom,ids and rename to msm8974pro-ac-motorola-victara

wonderfulShrineMaidenOfParadise:


TravMurav:
Thanks!

PR #638, opened 2026-01-27
dts: msm8952: Add support for Samsung Galaxy Tab A 8.0 2019 (gtowifi)

Kerem4838:


lk2nd/device/dts/msm8952/sdm429-samsung-gtowifi.dts
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_SDM429 0>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0>;
```
TravMurav: If this dts has MTP ids, it should be called "sdm429-mtp.dts" and your device below should be in a subnode. See how it's done in other mtp/qrd files near this one. You already have match-bootloader so it's not a big problem for you in this case

PR #645, opened 2026-02-06
Add Samsung Galaxy S4 LTE (jflte) support

gAtrium:
Tested, fastboot works, panel works, keys work.

lk2nd-screenshot


alex-min:
Hey, I don't think this PR is needed anymore, I've added jflte support on master.

Let me know if your S4 works fine with master as there's lots of models, if not I could make a PR to adjust the codename.

PR #665, opened 2026-03-20
Fix detection of motorola-titan

daviewales:
Move titan from QCDTBS to ADTBS. This enables the dts to be correctly detected, and prevents the bootloader from just picking the first ADTBS device in the list if no valid options are found. (Previously, it would pick the `lg-lenok` device tree.)

Revert a previous commit documenting the device as quirky, and remove quirkiness comment from device tree.

Remove unneeded `GPIO_PULL_UP` flag from `key-volume-down` device tree specification.


TravMurav:
Thanks!

PR #672, opened 2026-04-03
dts: msm8953: Add panels and some touchscreens for sdm632 Vsmart Joy 3

trihmah:
Add some additional panels and touchscreens
There is no touchscreen driver for the icnl9911s-byd panel yet


TravMurav:
Thanks!

PR #673, opened 2026-04-09
dts: msm8917 & msm8920: add support for the Motorola Moto E4 (perry)

agrecascino:
Another one of the nine billion names of moto, I guess. Packed msm8920 and msm8917 into one commit since they're effectively identical.


TravMurav:
Thanks!

PR #675, opened 2026-04-17
adding support for the Samsung Galaxy S4

alex-min:
I'm adding support for the Samsung Galaxy S4 (I've tested with the device)

I've also added the i9506 version as it's almost identical to the original version, just with a beefier cpu & network.

Thanks in advance.

### DIFF
--- a/lk2nd/device/dts/msm8960/bundle.dts
+++ b/lk2nd/device/dts/msm8960/bundle.dts
@@ -70,6 +70,29 @@
 			};
 		};
 	};
+	samsung-jflte {
+		model = "Samsung Galaxy S4 (GT-I9505)";
+		compatible = "samsung,jflte";
+		lk2nd,match-bootloader = "I9505*";
+		lk2nd,dtb-files = "qcom-apq8064-samsung-jflte";
+
+		gpio-keys {
+			compatible = "gpio-keys";
+
+			volume-up {
+				lk2nd,code = <KEY_VOLUMEUP>;
+				gpios = <&pmic 34 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			};
+			volume-down {
+				lk2nd,code = <KEY_VOLUMEDOWN>;
+				gpios = <&pmic 36 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			};
+			home {
+				lk2nd,code = <KEY_HOME>;
+				gpios = <&pmic 29 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			};
+		};
+	};
 	samsung-expressltexx {
 		model = "Samsung Galaxy Express (GT-I8730)";
 		compatible = "samsung,expressltexx", "qcom,msm8960";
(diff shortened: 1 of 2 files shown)


lk2nd/device/dts/msm8960/bundle.dts
```diff
@@ -70,6 +70,52 @@
...
+				gpios = <&pmic 29 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			};
+		};
+	};
+	samsung-jflte-i9506 {
+		model = "Samsung Galaxy S4 LTE-A (GT-I9506)";
+		compatible = "samsung,jflte", "qcom,apq8064";
+		lk2nd,match-bootloader = "I9506*";
+		lk2nd,dtb-files = "qcom-apq8064-samsung-jflte";
+
```
TravMurav: seems like this device was shipped with

> Qualcomm MSM8974 Snapdragon 800 (28 nm)

and not 

> Qualcomm APQ8064T Snapdragon 600

so I don't think it belongs to this file. If you can't test it, I suggest skipping it.

lk2nd/device/dts/msm8960/bundle.dts
```diff
@@ -70,6 +70,52 @@
 			};
 		};
 	};
+	samsung-jflte {
+		model = "Samsung Galaxy S4 (GT-I9505)";
+		compatible = "samsung,jflte", "qcom,apq8064";
```
TravMurav:
```suggestion
		compatible = "samsung,jflte";
```

Other nodes in this file are slightly incorrect.

lk2nd/device/dts/msm8960/bundle.dts
```diff
@@ -70,6 +70,52 @@
 			};
 		};
 	};
+	samsung-jflte {
+		model = "Samsung Galaxy S4 (GT-I9505)";
```
TravMurav, line 74: Please add to Documentation/devices.md in the same commit

lk2nd/device/dts/msm8960/bundle.dts
alex-min: Msm8960 and apq8064 are very close SOC. Apq8064 is basically the high end version of Msm8960, they have identical bootloaders.

Even Samsung's own stock bootloader on the Galaxy S4 is using msm8960 as a product code.

I did test it on the device

lk2nd/device/dts/msm8960/bundle.dts
alex-min: I shouldn't add the soc here?

No problem to remove it

lk2nd/device/dts/msm8960/bundle.dts
alex-min: Ah sorry I mean test the other S4? 

Not yet no, I'll remove it for now and add it later once I have my hand on one.

TravMurav: Yes, Apq8064 uses a different target in caf LK (and so in lk2nd) so I'd be surprised it would work here.

alex-min: I confirm it works with the apq8064 Galaxy S4 I9505 version, it's the one in the screenshot.

Samsung's own stock I9505 bootloader with Odin is also using msm8960, they didn't bother to create a different one for apq8064.

For the I9506 that's a fair point, I can't prove that it works, I9505 and i9506 share the same rom and are very close so it should work but I can't prove it so I'll remove this one.

alex-min: This chart should be clearer, Samsung models are a mess

alex-min:
I think I've implemented all your feedback, let me know if there's anything else

TravMurav:
Thanks!

PR #676, opened 2026-04-18
dts: msm8953: Add support for Xiaomi Mi 5X (tiffany)

tkw6688:
Tested on physical hardware. no significant issues found.

### DIFF
--- a/lk2nd/device/dts/msm8953/msm8953-xiaomi-common.dts
+++ b/lk2nd/device/dts/msm8953/msm8953-xiaomi-common.dts
@@ -11,6 +11,34 @@
 };
 
 &lk2nd {
+	tiffany {
+		/*
+		 * Xiaomi Mi 5X (tiffany) and Xiaomi Mi A1 (tissot) are hardware-identical.
+		 * tiffany's bootloader matches DAISY, but the panels match tissot.
+		 * Both properties are required to correctly identify tiffany.
+		 */
+		model = "Xiaomi Mi 5X (tiffany)";
+		compatible = "xiaomi,tiffany";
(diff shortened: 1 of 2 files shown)


lk2nd/device/dts/msm8953/msm8953-xiaomi-common.dts
```diff
@@ -11,6 +11,31 @@
 };
 
 &lk2nd {
+	tiffany {
+		model = "Xiaomi Mi 5X (tiffany)";
+		compatible = "xiaomi,tiffany";
+		lk2nd,match-bootloader = "*DAISY*";
+
+		/* Use panel matching to avoid conflict with Xiaomi Redmi 6 Pro (sakura) */
+		lk2nd,match-panel;
```
TravMurav, line 23: Can you reformat a bit to have both match properties after the comment and extend the comment that bootloader matches for daisy/sakura but panels match for tissot, so we need both rules

lk2nd/device/dts/msm8953/msm8953-xiaomi-common.dts
```diff
@@ -11,6 +11,31 @@
...
+
+		lk2nd,dtb-files = "msm8953-xiaomi-tissot";
```
TravMurav, line 25: Maybe worth a comment that they're same/similar hardware

lk2nd/device/dts/msm8953/msm8953-xiaomi-common.dts
tkw6688: Done. I've updated the PR to include both match properties after the comment and expanded the description as requested.

TravMurav:
Thanks!

PR #681, opened 2026-04-26
dts: msm8909: Add support for Lenovo Tab e10 (TB-X104F)

### LINKED ISSUE #680: About Lenovo Tab E10 (TB-X104F)
Nights!,

I was past Friday at the IRC channel seeking info about booting `lk2nd` on the above mentioned device, which I found isn't currently supported.

I was lucky enough to get to talk to such a nice person, that he spent a good couple of hours with me until the device booted, and even then he took his time trying out things in order to send a patch that could be usable.

There were a couple of blanks he couldn't fill, unfortunately, neither I could because of this whole subject being unknown to me to the point I couldn't yet backup my device in order to flash `lk2nd` (was booting it from RAM all the time).

He also informed me about the need of cloning the repository for a patch to be accepted, which I wouldn't have a problem investigating and doing if you require so, even if I'm not used to any of this at all.

I'm leaving here the following links, and if I could be of any help, whether providing feedback (trying out things you might require on the device and reporting back the results of these), or setting up a cloned repository if needed, or any other thing I could do, don't hesitate to let me know.

[Link](https://drive.google.com/file/d/1S-bxhux_tNKQSl1iD6Z6aFJzfQFs7h1d/view?usp=sharing) to the `boot.img` this person helped me extracting from the device.

[Link](https://drive.google.com/file/d/1Bm5jcFU50fndcO8PoxPR6x5RMNUyvu2i/view?usp=sharing) to the `apq8009-mtp.dts` file we created, which for sure has some errors and may be lacking things.

This device's boot image is not in `QCDT` format, so this file was added to the `lk2nd/device/dts/msm8909/rules.mk`'s `ADTBS` variable, and it's all that was required for it to boot.

All the buttons are usable and I could test every menu option does what it says, as well as all `fastboot oem` commands implemented (but `oem read` which always crashes `lk2nd`, and `oem write` which I haven't tried).

By the way, given the hardware and resources this device has are pretty much hopeless for regular use, my goal was to have something small to boot that I could turn into a shell to tinker around with forth, but I find this subject so huge I don't even know where to start from, so I can't help but ask, how do you guys even get close to here? Does all this effort come from previous personal experience in the field, only "helped" by the open source vendors have to release? Because that's quite remarkable.

Where I could start if wanted to support this device's panel so that I could have an on screen keyboard? I guess the registers must be in the kernel, but I'd be more interested first on theory behind how these types of displays are driven, including supporting touch events.

I've seen from within `lk2nd`'s source code you implement a frame buffer in software, which is totally fine with me as I'm not looking for performance but for a way of getting there to tinker with the device live without android in between.

Thanks,

somehandle738:
Pull request to add support for the above mentioned device.

There's the [`boot.img`](https://drive.google.com/file/d/1S-bxhux_tNKQSl1iD6Z6aFJzfQFs7h1d/view?usp=sharing) is available, and the downstream `dts` [file](https://drive.google.com/file/d/1LRVrY-qC5InqXU5kFqUPmQK7jELYOeub/view?usp=drive_link).

Thanks,

### DIFF
--- a/lk2nd/device/dts/msm8909/apq8009-qrd.dts
+++ b/lk2nd/device/dts/msm8909/apq8009-qrd.dts
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+#include <skeleton32.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id  = <QCOM_ID_APQ8009 2>;
+	qcom,board-id = <QCOM_BOARD_ID(QRD, 1, 0) 0>;
+};
+
+&lk2nd {
+	tb-x104f {
+		model = "Lenovo Tab e10 (TB-X104F)";
+		compatible = "lenovo,tb-x104f";
+
+		lk2nd,dtb-files = "msm8909-lenovo-tb-x104f";
+		lk2nd,match-panel;
+
+		panel {
+			compatible = "lenovo,tb-x104f-panel", "lk2nd,panel";
+
+			qcom,mdss_dsi_jd9365_boe_video {
+				compatible = "lenovo,jd9365-boe-video";
+			};
+		};
+	};
+};
--- a/lk2nd/device/dts/msm8909/rules.mk
+++ b/lk2nd/device/dts/msm8909/rules.mk
@@ -12,3 +12,5 @@ QCDTBS += \
 
 ADTBS += \
 	$(LOCAL_DIR)/apq8009w-wtp.dtb \
+	$(LOCAL_DIR)/apq8009-qrd.dtb \
+
--- a/Documentation/devices.md
+++ b/Documentation/devices.md
@@ -84,6 +84,7 @@
 - Nokia 8000 4G
 - Nokia 8110 4G
 - ZTE N818S (sapphire)
+- Lenovo Tab e10 (TB-X104F)
 
 ### lk2nd-msm8952
 


lk2nd/device/dts/msm8909/apq8009-mtp.dts
```diff
@@ -0,0 +1,26 @@
...
+#include <skeleton32.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id   = <QCOM_ID_APQ8009 2>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0x08 0x1000b 0x00>;
+};
+
+&lk2nd {
+	model = "Lenovo Tab e10 (TB-X104F)";
```
TravMurav: Please add the device to the Documentation/devices.md in the same commit.

lk2nd/device/dts/msm8909/apq8009-mtp.dts
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+#include <skeleton32.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id   = <QCOM_ID_APQ8009 2>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0x08 0x1000b 0x00>;
```
TravMurav: Each board-id is two values so this describes two boards - mtp/8 (for which your filename would be appropriate) and qrd 1.0. Please check which one do you actually need and remove another one.

lk2nd/device/dts/msm8909/apq8009-mtp.dts
```diff
@@ -0,0 +1,26 @@
...
+};
+
+&lk2nd {
+	model = "Lenovo Tab e10 (TB-X104F)";
+	compatible = "lenovo,tb-x104f";
+
+	lk2nd,dtb-files = "msm8909-lenovo-tb-x104f";
+	lk2nd,match-panel;
```
TravMurav: Since this device matches a generic msm/board id combo, can you double-check that cmdline you get doesn't include anything more useful to have a better match property? Regardless of that, please put this device into a subnode inside lk2nd node, see other files around.

lk2nd/device/dts/msm8909/apq8009-mtp.dts
somehandle738: I changed it as you mentioned, but now the device doesn't boot into `lk2nd'. I get:

```
Sending 'boot.img' (340 KB)                        OKAY [  0.014s]
Booting                                            FAILED (remote: 'dtb not found')
fastboot: error: Command failed
```

lk2nd/device/dts/msm8909/apq8009-mtp.dts
somehandle738: Added

lk2nd/device/dts/msm8909/apq8009-mtp.dts
somehandle738: Put it inside a subnode, as requested, but I don't know how to proceed with your request regarding the command-line.

This is the output of `fastboot get_staged parsed-cmdline`, if that's what you mean:

```
lk2nd androidboot.bootdevice=7824900.sdhci androidboot.verifiedbootstate=orange androidboot
.veritymode=enforcing androidboot.keymaster=1 androidboot.serialno=3324077d androidboot.sec
ureboot=enabled androidboot.baseband=apq mdss_mdp3.panel=1:dsi:0:qcom,mdss_dsi_jd9365_boe_v
ideo:1:none:cfg:single_dsi
```

TravMurav: Okay, I don't see anything better we can do other than match-panel sadly, so we can keep this part as-is

lk2nd/device/dts/msm8909/apq8009-mtp.dts
TravMurav: what if you only keep the other pair of values?

somehandle738: That made the trick

lk2nd/device/dts/msm8909/apq8009-mtp.dts
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+#include <skeleton32.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id   = <QCOM_ID_APQ8009 2>;
+	qcom,board-id = <0x1000b 0x00>;
```
TravMurav: it's a bit weird in the linux-msm kernel but let's call this file apq8909-qrd for now.

Please use the macro to decode this value to (qrd,1,0)

https://github.com/msm8916-mainline/lk2nd/blob/main/lk2nd/include/dt-bindings/arm/qcom%2Cids.h#L222

Please also squash all commits into one after you verify everything works properly

somehandle738: Was a bit tough, but see if that's what you wanted.

lk2nd/device/dts/msm8909/apq8909-qrd.dts
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+#include <skeleton32.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id   = <QCOM_ID_APQ8009 2>;
+	qcom,board-id = <QCOM_BOARD_ID (QRD, 1, 0) 0>;
```
TravMurav:
```suggestion
	qcom,msm-id  = <QCOM_ID_APQ8009 2>;
	qcom,board-id = <QCOM_BOARD_ID(QRD, 1, 0) 0>;
```

Let's keep the style consistent

lk2nd/device/dts/msm8909/rules.mk
```diff
@@ -12,3 +12,5 @@ QCDTBS += \
 
 ADTBS += \
 	$(LOCAL_DIR)/apq8009w-wtp.dtb \
+	$(LOCAL_DIR)/apq8909-qrd.dtb \
```
TravMurav: sorry, seems like I've mixed soc numbers up when making the suggestion


```suggestion
	$(LOCAL_DIR)/apq8009-qrd.dtb \
```

TravMurav (commented):
A couple of nits left and this will be ready to go

lk2nd/device/dts/msm8909/apq8009-qrd.dts
TravMurav, line 1: it doesn't build because the filename should be dts

lk2nd/device/dts/msm8909/apq8009-qrd.dtb
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+#include <skeleton32.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id   = <QCOM_ID_APQ8009 2>;
```
TravMurav:
```suggestion
	qcom,msm-id = <QCOM_ID_APQ8009 2>;
```

since we have to do another round anyway

TravMurav:
Thanks!

PR #682, opened 2026-05-03
dts: msm8226: motorola,peregrine: set lk2nd,dtb-files

a-andre:


lk2nd/device/dts/msm8226/msm8926-motorola-peregrine.dts
```diff
@@ -11,5 +11,5 @@
 	model = "Motorola Moto G 4G 2013";
 	compatible = "motorola,peregrine";
 
-	//FIXME: lk2nd,dtb-files = "msm8926-motorola-peregrine";
+	lk2nd,dtb-files = "msm8926-motorola-peregrine";
```
daviewales, line 14: Shouldn't this be `qcom-msm8926-motorola-peregrine`?

https://github.com/msm8226-mainline/linux/blob/qcom-msm8226-6.16.y/arch/arm/boot/dts/qcom/qcom-msm8926-motorola-peregrine.dts

a-andre: No, according to https://github.com/msm8916-mainline/lk2nd/blob/main/Documentation/dt-bindings.md#model-and-compatible

> Additionally `lk2nd,dtb-files` property can be present with a list of possible device-tree names for this device. Don't add `qcom/` or `qcom-` prefix and `.dtb` suffix to the values since lk2nd will add them as necessary.

TravMurav:
Thanks!

PR #684, opened 2026-05-04
dts: msm8952: dts for Readboy G500X (g500x)

S-lkno:
IMG_20260504_110327~2

### DIFF
--- a/Documentation/devices.md
+++ b/Documentation/devices.md
@@ -110,6 +110,7 @@
 - Motorola Moto G5S (montana)
 - Motorola Moto G6 Play (jeter)
 - OPPO A57 (A57) (quirky - see comment in `lk2nd/device/dts/msm8952/msm8940-oppo-a57.dts`)
+- Readboy G500X Tablet (msm8937_64)(support lk1st with qtestsign)
 - Redmi 3S (land)
 - Redmi 4 (prada)
 - Redmi 4A (rolex)
--- a/Documentation/devices.md
+++ b/Documentation/devices.md
@@ -110,7 +110,7 @@
 - Motorola Moto G5S (montana)
 - Motorola Moto G6 Play (jeter)
 - OPPO A57 (A57) (quirky - see comment in `lk2nd/device/dts/msm8952/msm8940-oppo-a57.dts`)
-- Readboy G500X Tablet (msm8937_64)(support lk1st with qtestsign)
+- Readboy G500X Tablet
 - Redmi 3S (land)
 - Redmi 4 (prada)
 - Redmi 4A (rolex)
--- a/lk2nd/device/dts/msm8952/msm8937-readboy-g500x.dts
+++ b/lk2nd/device/dts/msm8952/msm8937-readboy-g500x.dts
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8937 0>;
+	qcom,board-id = <0x2000b 0>;
+};
+
+&lk2nd {
+		model = "Readboy G500X Tablet";
+		compatible = "readboy,g500x";
+
+		lk2nd,match-device = "g500x";
+		lk2nd,dtb-files = "msm8937-readboy-g500x";
+
+	};
--- a/lk2nd/device/dts/msm8952/msm8937-readboy-g500x.dts
+++ b/lk2nd/device/dts/msm8952/msm8937-readboy-g500x.dts
@@ -7,14 +7,17 @@
 
 / {
 	qcom,msm-id = <QCOM_ID_MSM8937 0>;
-	qcom,board-id = <0x2000b 0>;
+	qcom,board-id = <QCOM_BOARD_ID(QRD, 2, 0) 0>;
 };
 
 &lk2nd {
-		model = "Readboy G500X Tablet";
-		compatible = "readboy,g500x";
+	/*
+	 * Readboy G500X Tablet
+	 * To build lk1st, use: make TOOLCHAIN_PREFIX=arm-none-eabi- lk1st-msm8952
+	 * (with a qtestsign signature after build)
+	 */
+	model = "Readboy G500X Tablet";
+	compatible = "readboy,g500x";
 
-		lk2nd,match-device = "g500x";
-		lk2nd,dtb-files = "msm8937-readboy-g500x";
-    lk2nd,enable-storage;
-	};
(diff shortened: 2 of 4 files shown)


lk2nd/device/dts/msm8952/rules.mk
```diff
@@ -13,6 +13,7 @@ ADTBS += \
 	$(LOCAL_DIR)/msm8937-nokia-nd1.dtb \
 	$(LOCAL_DIR)/msm8937-nokia-ple.dtb \
 	$(LOCAL_DIR)/msm8937-xiaomi-land.dtb \
+  $(LOCAL_DIR)/msm8937-readboy-g500x.dtb \
```
TravMurav: Please sort alphabetically. Also indentation is broken here and in the dts

lk2nd/device/dts/msm8952/msm8937-readboy-g500x.dts
```diff
@@ -0,0 +1,20 @@
...
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8937 0>;
+	qcom,board-id = <0x2000b 0>;
+};
+
+&lk2nd {
+		model = "Readboy G500X Tablet";
```
TravMurav: Please add to Documentation/devices.md in the same commit

S-lkno: Thanks for the review! I've resolved the issues you pointed out — added the device to Documentation/devices.md and fixed the indentation. Please re-review when you have a chance.

lk2nd/device/dts/msm8952/msm8937-readboy-g500x.dts
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+/dts-v1/;
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8937 0>;
+	qcom,board-id = <0x2000b 0>;
```
TravMurav: seems like I've missed it but this looks like `QCOM_BOARD_ID(QRD, 2, 0)`

https://android.googlesource.com/kernel/msm/+/android-7.1.0_r0.2/arch/arm/boot/dts/qcom/msm8937-pmi8950-qrd-sku1.dts#23

Since you already have match-device, I assume your cmdline has the "g500x" in it, so we can trivially convert this to multi-device file. Please replace the const with a macro as suggested, rename the file to `msm8937-pmi8950-qrd-sku1.dts` and put your device information into a subnode under lk2nd node, i.e. like it's done here:

https://github.com/msm8916-mainline/lk2nd/blob/main/lk2nd/device/dts/msm8952/msm8952-mtp.dts

This way if we have to add other device with the same ID to this file, we will be able to do it trivially and without breaking yours

lk2nd/device/dts/msm8952/msm8937-readboy-g500x.dts
```diff
@@ -0,0 +1,19 @@
...
+};
+
+&lk2nd {
+		model = "Readboy G500X Tablet";
+		compatible = "readboy,g500x";
+
+		lk2nd,match-device = "g500x";
+		lk2nd,dtb-files = "msm8937-readboy-g500x";
+	};
```
TravMurav: the indentation is this file is still weird for some reason...

Documentation/devices.md
```diff
@@ -110,6 +110,7 @@
 - Motorola Moto G5S (montana)
 - Motorola Moto G6 Play (jeter)
 - OPPO A57 (A57) (quirky - see comment in `lk2nd/device/dts/msm8952/msm8940-oppo-a57.dts`)
+- Readboy G500X Tablet (msm8937_64)(support lk1st with qtestsign)
```
TravMurav:
```suggestion
- Readboy G500X Tablet
```

We usually write additional notes as comments in the dts when those comments are needed/useful. You can add a comment inside your device node describing how to build lk1st for it if you want. (No need to add quirky note as with the line above if you device boots lk2nd fine from a generic build)

S-lkno:
Thanks for the review. I'm new to this, and I appreciate the guidance.

The match-device property was added by mistake and has been removed. This dts is for a single device only.

S-lkno:
Thanks. I realize I've been rushing this. I'll take some time to learn more and come back to this later.

lk2nd/device/dts/msm8952/msm8937-readboy-g500x.dts
TravMurav: This comment still stands, we can't really make this into a single-device dts since the OEM reused a generic msm-id, so it's possible some other device exists that has same msm-id, and if someone wants to add it, they won't be able to without possibly breaking yours. To make sure this won't happen, you need to make this into a multi-device dts with appropriate name

S-lkno:
I've tested the subnode format with both match-device and qcom,board-id in the subnode. It consistently fails because the stock bootloader does not pass any androidboot.device parameter in the cmdline, and the board-id matching in subnodes doesn't trigger on this device. The only working configuration is with the properties directly under &lk2nd. I've applied the macro and renamed the file as requested. Given the hardware limitations, this is the only viable format. Please let me know if you have any alternative approach that I can test.

TravMurav:
> I've tested the subnode format with both match-device and qcom,board-id in the subnode. It consistently fails because the stock bootloader does not pass any androidboot.device parameter in the cmdline, and the board-id matching in subnodes doesn't trigger on this device. The only working configuration is with the properties directly under &lk2nd. I've applied the macro and renamed the file as requested. Given the hardware limitations, this is the only viable format. Please let me know if you have any alternative approach that I can test.

If there is nothing useful in cmdline, you can still use the match-panel rule as a last resort, would you mind trying that?

S-lkno:
Of course I'd be willing to give it a try, but I don't know what Match-Panel means? , to recognize the screen? There are several panels in the DTB of this device, I need to ask, if you don't mind

TravMurav:
> Of course I'd be willing to give it a try, but I don't know what Match-Panel means? , to recognize the screen? There are several panels in the DTB of this device, I need to ask, if you don't mind

Please see https://github.com/msm8916-mainline/lk2nd/blob/main/Documentation/dt-bindings.md#match-property

If you can find one of the properties in the list that work instead of match-panel that would be perfect (or if unsure, feel free to share cmdline string that lk2nd sees, you can find it in lk2nd log). If you have to add match-panel anyway, take care if you're using decompiled dtb since some vendors don't delete "default" test panels, you want to include those which match resolution of your device, and not things like simulated panel etc...

PR #686, opened 2026-05-07
Implement support for TP-Link M7350 V3 (MDM9625)

jackmthws:
This PR fixes up support for MDM9625 and supports the M7350 V3 LTE hotspot.

Included in this PR is tree wide fixes for MDM9625, support for ONFI NAND in qpic_nand.c and lk2nd's required pieces.
I don't think this is entirely done, not sure how to handle the tags_offset, or the hash command (we could add crypto support???). MMC clocks are also added but afaik not actually really used. The device does have an SD card slot though so it could be supported. But this chip does not have an SDHCI controller so not sure how much work that is.

The DTS name being msm9625 is not incorrect: https://git.codelinaro.org/clo/la/kernel/msm/-/blob/LNX.LE.3.2.1-35120-9x25/arch/arm/boot/dts/msm9625-v2.1-mtp.dts?ref_type=tags


lk2nd/device/dts/mdm9625/msm9625-v2.1-mtp.dts
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <134 7 0x20001>, <152 7 0x20001>, <149 7 0x20001>,
+		      <150 7 0x20001>, <151 7 0x20001>, <148 7 0x20001>,
+		      <173 7 0x20001>, <174 7 0x20001>, <175 7 0x20001>;
```
TravMurav, line 9: I wonder if those msm-ids (134, 152...) are documented somewhere so we could've added them to qcom,ids.h and used here then

lk2nd/device/dts/mdm9625/msm9625-v2.1-mtp.dts
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
```
TravMurav, line 3: idk if we really tried to keep this consistent honestly but maybe

```suggestion
#include <skeleton32.dtsi>
```

lk2nd/device/dts/mdm9625/rules.mk
```diff
@@ -0,0 +1,4 @@
+LOCAL_DIR := $(GET_LOCAL_DIR)
+
+QCDTBS += \
+	$(LOCAL_DIR)/msm9625-v2.1-mtp.dtb
```
TravMurav, line 4:
```suggestion
	$(LOCAL_DIR)/msm9625-v2.1-mtp.dtb \

```

so future diffs are smaller

lk2nd/fastboot/rules.mk
```diff
@@ -3,7 +3,6 @@ LOCAL_DIR := $(GET_LOCAL_DIR)
 
 OBJS += \
 	$(LOCAL_DIR)/fetch.o \
-	$(LOCAL_DIR)/hash.o \
```
TravMurav, line 6: probably needs some ifneq on some condition for crypto module existing, like for screenshot below, no need to bother implementing crypto for now, this tool was mainly added for testing

TravMurav (commented):
> not sure how to handle the tags_offset

I'd prefer if we find a way to limit this to only this target, as to not change all other ones...

> or the hash command (we could add crypto support???).

Don't see point spending time on this, see below.

> MMC clocks are also added but afaik not actually really used. The device does have an SD card slot though so it could be supported. But this chip does not have an SDHCI controller so not sure how much work that is.

Confused... Does it have some non-sdhci-compliant mmc controller then?

If you think you can bring up sd support in lk2nd (for booting from sd) then great, but unless it's somehow necessary, I suggest dropping mmc clock changes until you figure the controller out, and then add at once with sd card support

lk2nd/device/dts/mdm9625/msm9625-v2.1-mtp.dts
jackmthws: some are definitely documented in downstream but I think I remember some missing

jackmthws:
> > not sure how to handle the tags_offset
> 
> I'd prefer if we find a way to limit this to only this target, as to not change all other ones...

For sure

> > MMC clocks are also added but afaik not actually really used. The device does have an SD card slot though so it could be supported. But this chip does not have an SDHCI controller so not sure how much work that is.
> 
> Confused... Does it have some non-sdhci-compliant mmc controller then?
> 
> If you think you can bring up sd support in lk2nd (for booting from sd) then great, but unless it's somehow necessary, I suggest dropping mmc clock changes until you figure the controller out, and then add at once with sd card support

https://git.codelinaro.org/clo/la/kernel/msm/-/blob/LNX.LE.3.2.1-35120-9x25/arch/arm/boot/dts/msm9625.dtsi?ref_type=tags#L479-541 yeah, I think it uses the non sdhci mmc driver from msm_shared. I think most of lk2nd is set up for the SDHCI controller though... I added the clocks to avoid removing the mmc controller from compilation but that is another solution that could be used

TravMurav:
> ... yeah, I think it uses the non sdhci mmc driver from msm_shared. I think most of lk2nd is set up for the SDHCI controller though... I added the clocks to avoid removing the mmc controller from compilation but that is another solution that could be used

Ah I see, very quickly looking at it, it's a bit annoying to bring up "second" card probably, since the old mmc driver just saves configuration values into a static variable (instead of having an instance struct like later sdhci, wrapper do)... But I think it should work fine in wrapper mode (i.e. if you assume that sdcard is the only mmc in the system) so if mmc init is added to the target, I'd not be surprised that's all that's needed for lk2nd to be able to read the sd card

https://github.com/msm8916-mainline/lk2nd/blob/main/target/msm8960/init.c#L157-L167

PR #687, opened 2026-05-16
display: panel: generated: generate `truly_qhd_video` panel.

erikas9987:
Same usage as `innolux_qhd_video` for sony-eagle. Also including instructions on how to pick the right panel for your device.
This might be happening because the primary bootloader does not initialize the panel, unlike in most cases.


TravMurav:
Thanks!

PR #688, opened 2026-05-18
`/chosen/rng-seed` support

samcday:
CAF LK already had scaffolding to obtain entropy from TZ (via `SVC_CRYPTO`) and shovel it into `/chosen/kaslr-seed`. It just wasn't wired up anywhere (the only thing that was pulling in `ENABLE_KASLRSEED_SUPPORT` was `AndroidBoot.mk`).

So, we piggy back this prior art to introduce `/chosen/rng-seed`. This DT prop is automatically credited into the entropy pool by the kernel: https://github.com/torvalds/linux/blob/6916d5703ddf9a38f1f6c2cc793381a24ee914c6/drivers/of/fdt.c#L1112

Note that I've only tested this on my msm8930 samsung-expressltexx. I don't have any msm8960-family devices to test with.

### DIFF
--- a/lk2nd/rng-seed/rng-seed.c
+++ b/lk2nd/rng-seed/rng-seed.c
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <boot.h>
+#include <debug.h>
+#include <libfdt.h>
+#include <scm.h>
+#include <string.h>
+#include <sys/types.h>
+
+#define RNG_SEED_BYTES 64
+
+/**
+ * lk2nd_rng_seed_dt_update() - write a random seed to /chosen/rng-seed
+ *
+ * scm_random() is used to generate the entropy in a best-effort manner.
+ * failures are logged, but do not fail the boot
+ */
+static int lk2nd_rng_seed_dt_update(void *dtb, const char *cmdline,
+				    enum boot_type boot_type)
+{
+	uintptr_t rngseed[RNG_SEED_BYTES / sizeof(uintptr_t)];
+	unsigned int i;
+	int offset, ret;
+
+	if (boot_type & (BOOT_DOWNSTREAM | BOOT_LK2ND))
+		return 0;
+
+	for (i = 0; i < ARRAY_SIZE(rngseed); i++) {
+		ret = scm_random(&rngseed[i], sizeof(rngseed[i]));
+		if (ret) {
+			dprintf(INFO, "rng-seed: scm_call for random failed: %d\n", ret);
+			return 0;
+		}
+	}
+
+	offset = fdt_path_offset(dtb, "/chosen");
+	if (offset < 0) {
+		dprintf(INFO, "rng-seed: couldn't find /chosen: %d\n", ret);
+		return 0;
+	}
+
+	ret = fdt_setprop(dtb, offset, "rng-seed", rngseed, sizeof(rngseed));
+	if (ret < 0)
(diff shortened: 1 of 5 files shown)


lk2nd/rng-seed/rng-seed.c
```diff
@@ -0,0 +1,44 @@
...
+{
+	uintptr_t rngseed[RNG_SEED_BYTES / sizeof(uintptr_t)];
+	unsigned int i;
+	int offset, ret;
+
+	if (boot_type & (BOOT_DOWNSTREAM | BOOT_LK2ND))
+		return 0;
+
+	for (i = 0; i < ARRAY_SIZE(rngseed); i++) {
+		ret = scm_random(&rngseed[i], sizeof(rngseed[i]));
```
TravMurav, line 29: Can we make this an array of u8 and populate with a single call?

lk2nd/rng-seed/rng-seed.c
```diff
@@ -0,0 +1,44 @@
...
+#include <boot.h>
+#include <debug.h>
+#include <libfdt.h>
+#include <scm.h>
+#include <string.h>
+#include <sys/types.h>
+
+#define RNG_SEED_BYTES 64
+
+// write a random seed to /chosen/rng-seed. scm_random() is used to generate the entropy
```
TravMurav: would you mind reformatting this like kinda-kerneldoc we do in other places? No need to describe params/return, just same text + function name header in the c-style comment

https://github.com/msm8916-mainline/lk2nd/blob/a7f2274eba3287d9bf32620efc86c7806d694af0/lk2nd/boot/extlinux.c#L338-L348

lk2nd/rng-seed/rng-seed.c
```diff
@@ -0,0 +1,44 @@
...
+	unsigned int i;
+	int offset, ret;
+
+	if (boot_type & (BOOT_DOWNSTREAM | BOOT_LK2ND))
+		return 0;
+
+	for (i = 0; i < ARRAY_SIZE(rngseed); i++) {
+		ret = scm_random(&rngseed[i], sizeof(rngseed[i]));
+		if (ret) {
+			dprintf(INFO,"rng-seed: scm_call for random failed: %d\n", ret);
```
TravMurav:
```suggestion
			dprintf(INFO, "rng-seed: scm_call for random failed: %d\n", ret);
```

lk2nd/rng-seed/rng-seed.c
TravMurav: Okay, the api is dumb and i see it now, disregard

lk2nd/rng-seed/rng-seed.c
```diff
@@ -0,0 +1,44 @@
...
+#include <sys/types.h>
+
+#define RNG_SEED_BYTES 64
+
+// write a random seed to /chosen/rng-seed. scm_random() is used to generate the entropy
+// best effort: failures are logged, but should not fail the boot.
+static int lk2nd_rng_seed_dt_update(void *dtb, const char *cmdline,
+				    enum boot_type boot_type)
+{
+	uintptr_t rngseed[RNG_SEED_BYTES / sizeof(uintptr_t)];
```
TravMurav, line 21: does this work?

```suggestion
	uintptr_t rngseed[RNG_SEED_BYTES / sizeof(rngseed[0])];
```

TravMurav: Okay I see, sad :(

Can keep the original version

TravMurav:
Thanks!

PR #693, opened 2026-05-19
project: lk2nd-msm8953: enable KASLR seed support 

barni2000:
Enable KASLR seed support for MSM8953 platform.

Tested on SDM450 Redmi 5 (rosy), I can do more tests on different devices if it is necessary


TravMurav:
Thanks!

PR #696, opened 2026-05-25
lk2nd: display: cont-splash: Trigger display refresh for MDP3/4

ejona86:
This fixes display refresh for the Asus Nexus 7 (flo), which is in command mode with no auto refresh.

The #ifndef RDBK_DATA0 allows msm8610 to set the define when compiling
dma.c, which includes mdp.h. I could have instead moved the mipi_dsi.h
include out of mdp.h, but the pre-existing undef goo in msm8610 made
this seem the best mirror for the current RDBK_DATA0 handling.

@user0-07161, you may want to check if this fixes Samsung Galaxy Ace 3 LTE.

### DIFF
--- a/lk2nd/display/cont-splash/refresh.c
+++ b/lk2nd/display/cont-splash/refresh.c
@@ -17,6 +17,9 @@ static void mdp_refresh(void)
 {
 #if MDP3 || MDP4
 	writel(1, MDP_DMA_P_START);
+#ifdef DSI_CMD_MODE_MDP_SW_TRIGGER
+	writel(1, DSI_CMD_MODE_MDP_SW_TRIGGER);
+#endif
 #elif MDP5
 	writel(1, MDP_CTL_0_BASE + CTL_START);
 #endif
--- a/lk2nd/display/cont-splash/mdp.h
+++ b/lk2nd/display/cont-splash/mdp.h
@@ -4,9 +4,11 @@
 
 #if MDP3
 #include <mdp3.h>
+#include <mipi_dsi.h>
 #elif MDP4
 #include <dev/lcdc.h>
 #include <mdp4.h>
+#include <mipi_dsi.h>
 #elif MDP5
 #include <mdp5.h>
 #else
--- a/platform/msm_shared/include/mipi_dsi.h
+++ b/platform/msm_shared/include/mipi_dsi.h
@@ -79,7 +79,9 @@
 #define DTYPE_GEN_LWRITE 0x29	/* 4th Byte is 0xc0 */
 #define DTYPE_DCS_WRITE1 0x15	/* 4th Byte is 0x80 */
 
+#ifndef RDBK_DATA0
 #define RDBK_DATA0 0x06C
+#endif
 
 #define MIPI_VIDEO_MODE	        1
 #define MIPI_CMD_MODE           2


ejona86:
I've left the RDBK_DATA0 fixup as a separate commit. It's not entirely obvious what approach would be preferred there. I can squash/rework as requested.

user0-07161:
hi i'm gonna test this once i've got some time

user0-07161:
yeah this works on loganrelte

lk2nd/display/cont-splash/refresh.c
```diff
@@ -17,6 +17,9 @@ static void mdp_refresh(void)
 {
 #if MDP3 || MDP4
 	writel(1, MDP_DMA_P_START);
+#ifdef DSI_CMD_MODE_MDP_SW_TRIGGER
+	writel(1, DSI_CMD_MODE_MDP_SW_TRIGGER);
```
TravMurav, line 21: I'm a bit confused on where this comes from, according to the MSM8916 documentation (which is mdp5 but eh close enough?) this should be "Software trigger (for debug purpose only)"

Can you readl the DSI_TRIG_CTRL register and check what's set in it?

Looking at the mipi_dsi code seems like they indeed reconfigure it for the sw trigger instead of the freerunning mode but it'd still be good to know how the hardware is configured...

I'm also worried if there's hardware that's configured in such a way that issuing sw trigger while free running mode is already going would glitch the hardware

https://github.com/msm8916-mainline/lk2nd/blob/b611280115149b26510fbc9025d6282a9c7c460b/platform/msm_shared/mipi_dsi.c#L1009

(ftr msm8916 TRM page 793 and 836)

ejona86: The apq8064 hardware register description doc names the register "DSI Command Mode Display Processor Path Software Trigger Register" and bit 0 as "Software can start an MPD trigger for the Command Mode engine by writing a 1 to this bit".

Adding `dprintf(INFO, "DSI_TRIG_CTRL %x\n", readl(REG_DSI(0x080)));` right after the `MDP_DMA_P_START` prints `DSI_TRIG_CTRL 40`. That means dma trigger didn't stick in `mipi_dsi_panel_initialize()`? The apq8064 doc decodes 0x40 as DMA_TRIGGER_SEL:none and TRIGGER_SEL:"sw trigger".

Is it possible to get here while free running? I thought this code could only be run if `cmd_mode && !auto_refresh` within `mdp_setup_refresh()`.

FWIW, I know little about this other than it worked and that I wish I had documentation. This should be matching reasonable display.c:

https://github.com/msm8916-mainline/lk2nd/blob/3018998b7f9bb7f3cfa10dad78b82b05d8884b0f/platform/msm_shared/display.c#L227-L229

[304 was excluded for msm8610](https://github.com/msm8916-mainline/lk2nd/commit/8d534dd3ff483144b9d60d07898c0fa9d623cbd3). [305 was excluded for msm8909](https://github.com/msm8916-mainline/lk2nd/commit/f9e10c4464fb8432fdc204ea7bc5ec1fd405a842). Both of those have DISPLAY_TYPE_MDSS=1 which disables this code.

TravMurav: изображение

so per my understanding DMA_P_START kicks MDP to read one frame and then trigger MDSS automatically, in "free running" mode, and sw trigger lets you perform those steps manually. and autorefresh mode would've been kicking the dma start on it's own too, so there would've been no need for the manual fb flush.

Perhaps there's some errata or other weirdness by which they decided to manually trigger it, so OEM bl configures it for sw mode... Have you checked which lk targets this change would affect other than 8064?

ejona86: I understand what you mean by free running now.

I think this is just how it worked prior to mdp5; mdss didn't exist with mdp4, right? I think in mdp3/4 the hardware didn't do this for you and in mdp5 it became vestigial, thus the documentation saying it was for debugging.

I pointed out it impacts the same targets as the existing display.c. I was expecting it didn't _really_ impact many other targets simply because the others weren't mdp3/4 in "command mode without auto refresh". loganrelte and huashan (both msm8960) were the only ones I saw that mentioned broken display, and it wasn't clear why huashan needed ENABLE_DISPLAY=0. I'm not sure how to determine what other targets would be _actually_ impacted.

I'd be fine limiting the scope of this down to fewer targets. Earlier I had considered it, but it looked hackier than the current code, which actually seems pretty natural/normal when compared to display.c. I hadn't considered `DSI_TRIG_CTRL` though. Should the code read that to decide when to enable this?

TravMurav: I guess only msm8660 and msm8960 targets should be actually affected by this change, it's probably fine to merge it as-is then, if multiple 8960 devices were broken without this...

Please reorder two commits around though so lk2nd compiles at all commit steps

ejona86:
I squashed the two commits; I had considered that second commit as a fix-up and leaving it separate seemed to make it more obvious for your review. I tried a bit to reorder as you suggested, but it really wasn't needed previously so the commit message was a bit hard to write the reasoning behind the change.

TravMurav:
Thanks!

> but it really wasn't needed previously so the commit message was a bit hard to write

I'd written something like "In preparation for blah blah inclusion to blah which causes the error, make sure the error doesn't happen" but squashing it together is good enough for this change I guess...

PR #697, opened 2026-05-25
dts: msm8960: Add Asus Nexus 7 (flo) support

ejona86:
For the display to actually work, #696 is needed.

### DIFF
--- a/lk2nd/device/dts/msm8960/bundle.dts
+++ b/lk2nd/device/dts/msm8960/bundle.dts
@@ -3,6 +3,25 @@
 #include <lk2nd.dtsi>
 
 &lk2nd {
+	asus-nexus7 {
+		model = "Asus Nexus 7 (flo)";
+		compatible = "asus,nexus7-flo";
+		lk2nd,match-bootloader = "FLO-*";
+		lk2nd,dtb-files = "apq8064-asus-nexus7-flo";
+
+		gpio-keys {
+			compatible = "gpio-keys";
+
+			volume-up {
+				lk2nd,code = <KEY_VOLUMEUP>;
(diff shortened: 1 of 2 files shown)


TravMurav:
Thanks!

lk2nd/device/dts/msm8960/bundle.dts
```diff
@@ -3,6 +3,25 @@
...
+			compatible = "gpio-keys";
+
+			volume-up {
+				lk2nd,code = <KEY_VOLUMEUP>;
+				gpios = <&pmic 3 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			};
+			volume-down {
+				lk2nd,code = <KEY_VOLUMEDOWN>;
+				gpios = <&pmic 37 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+			};
```
a-andre, line 22: May I ask, why are GPIOs 3 and 37 used here, while [upstream](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm/boot/dts/qcom/qcom-apq8064-asus-nexus7-flo.dts?h=v7.1#n51) uses GPIOs 4 and 38, respectively?

TravMurav: From a very quick look seems like linux gpio numbers are 1 based and we pass them through to the caf lk driver as-is so we get them 0 based

https://elixir.bootlin.com/linux/v7.1/source/drivers/pinctrl/qcom/pinctrl-ssbi-gpio.c#L112
https://lxr.trvn.ru/lk2nd/latest/source/dev/pmic/pm8921/include/dev/pm8921.h#L131

In a sense it might be a (minor) lk2nd bug...

PR #698, opened 2026-05-25
lk2nd: boot: Increase max kernel size to 64 MiB

ejona86:
This allows booting generic postmarketOS kernels. As an example, linux-postmarketos-lts is 15 MiB on-disk, but is 37 MiB after self-decompression.


TravMurav:
We can't really do that, the allocation was made from 50MiB because that's what's available on the majority of platforms (and even then we've hit problems because some older platforms have less). so changing this would lead to crashes on some devices where reserved memory etc is placed right after those 50MiB. This situation sucks but we have to deal with what caf LK left us with...

ejona86:
That makes sense. I could tell some devices were quite memory constrained, but I couldn't figure out where the 50 MiB magic number came from.

I've made this a draft and pushed two example approaches that would seem to improve the situation.

 * The first approach allows platforms to opt-in to larger kernel sizes. I couldn't find any platforms that had less than 50 MiB available, but this would also allow _reducing_ the memory available if they can be identified. I do wonder if it'd be good to make dynamic at runtime in case things like msm8952 want to more dynamically adjust, or maybe that is just a good example of why we shouldn't use this method because there can be a lot of variance with TrustZone.
 * The second approach moves the ramdisk as far back in memory as allowed to give more room to the kernel. That does require an accurate definition of how much memory is available; it would break if a platform has less than 50 MiB.

TravMurav:
It's probably not a bad idea to define per-target memory limit and then allocate the memory more dynamically.

I think I agree with the approach that after we know how large our memory block is, we can put initramfs to the end of it (need to double check if linux mandates any alignment requirements there), and let kernel have as much space as possible within the known limit.

I'd be hesitant to change default limit from 50MiB though to instead let people with real devices check if it works before adding it to specific platforms.

I guess I'd expect one to define a 50MiB default limit in lk2nd/boot/rules.mk as a make variable (so it can be overriden via make cmdline defines) and pass it into DEFINES, have extlinux.c refactored per your suggestion to place initramfs at the end of the buffer, then as a separate commit increase the buffer on the platform you can verify this on, by editing the respective lk2nd-blah.mk project file. The new make variable should be documented in building.md and we should probably add a short explaination on how memory is allocated in boot.md, providing a list that would say how much ram each platform has (i.e. a list with two entries for now - platform you've changed and verified, and "the rest - 50MiB")

Organizationally, I'd imagine this to be three commits, i.e.:
- lk2nd: boot: extlinux: Maximize kernel image memory
- lk2nd: boot: extlinux: Allow to configure the amount of boot memory   // documentation added here
- project: lk2nd-blah: Increase boot memory to XXMiB   // platform added to the list in documentation

Then since you've already done some research on which platforms can have increased memory limits, you can make an issue and document which platforms you think should have more memory

ejona86:
@TravMurav, thanks so much. Yeah, I wasn't planning on _actually_ changing the other platforms; I did that to give a better view of what would be possible. It was surprising how msm8960 was the peak, I guess because it was toward the end of armv7 and arm64 has TrustZone.

Since this has changed so much from "lk2nd: boot: Increase max kernel size to 64 MiB", I'm closing it and opened #703 as the first of three PRs.

PR #700, opened 2026-05-27
dts: msm8952: Add support for Huawei Enjoy 7 Plus

QuiXan:
This device is known to have two configuration combinations, corresponding to two board IDs.
The storage options are 3 / 32 and 4 / 64 respectively.

### DIFF
--- a/lk2nd/device/dts/msm8952/msm8940-huawei-trt-tl10.dts
+++ b/lk2nd/device/dts/msm8952/msm8940-huawei-trt-tl10.dts
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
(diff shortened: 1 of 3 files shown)


lk2nd/device/dts/msm8952/msm8940-huawei-trt-tl10.dts
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <0x139 0x00>;
```
TravMurav: Please use defines from https://github.com/msm8916-mainline/lk2nd/blob/main/lk2nd/include/dt-bindings/arm/qcom%2Cids.h

TravMurav:
Thanks!

PR #704, opened 2026-06-01
dts: msm8952: add cat-s22flip

CE1CECL:


CE1CECL:
FYI: I have the Amazon "Unifone" variant of the Cat S22 Flip Phone (Bullitt Group), which has a different ROM/fingerprint/kernel binary than the original T-Mobile branded Cat S22, but it should work on the original too (it just has slight differences but flashing the T-Mobile ROM on the Unifone variant will brick Wi-Fi or Modem unless you have a backup, which I do).
Link: https://www.amazon.com/dp/B0DK99QFDV

lk2nd/device/dts/msm8952/qm215-mtp.dts
```diff
@@ -0,0 +1,28 @@
...
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+       qcom,msm-id = <QCOM_ID_QM215 0>;
+       qcom,board-id = <QCOM_BOARD_ID_MTP 0>;
+};
+
+&lk2nd {
+       cat-s22flip {
```
TravMurav: We use tabs not spaces for indentation in lk2nd, please change that

lk2nd/device/dts/msm8952/qm215-mtp.dts
```diff
@@ -0,0 +1,28 @@
...
+
+               lk2nd,match-panel;
+
+               panel {
+                       compatible = "cat,s22flip-panel";
+
+                       qcom,mdss_dsi_jd9161z_boe_ips_video {
+                               compatible = "cat,s22flip";
+                       };
+               };
```
TravMurav: This is a bit interesting i guess since i'd imagine a device like this would've had an SPI display instead of DSI one... Do you happen to know if the hardware actually uses DSI to drive the display?

I'm assuming you're just using this for device detection, is there nothing better in cmdline to match for?

lk2nd/device/dts/msm8952/qm215-mtp.dts
```diff
@@ -0,0 +1,28 @@
...
+#include <lk2nd.dtsi>
+
+/ {
+       qcom,msm-id = <QCOM_ID_QM215 0>;
+       qcom,board-id = <QCOM_BOARD_ID_MTP 0>;
+};
+
+&lk2nd {
+       cat-s22flip {
+               model = "Cat S22 Flip";
```
TravMurav: Please add to Documentation/devices.md in the same commit

lk2nd/device/dts/msm8952/qm215-mtp.dts
```diff
@@ -0,0 +1,28 @@
...
+               compatible = "cat,s22flip";
+
+               // FIXME: lk2nd,dtb-files = "qm215-cat-s22flip";
+
+               lk2nd,match-panel;
+
+               panel {
+                       compatible = "cat,s22flip-panel";
```
TravMurav: This should have a second "lk2nd,panel" compatible for lk2nd to find the node for dt update logic.


```suggestion
                       compatible = "cat,s22flip-panel", "lk2nd,panel";
```

TravMurav (commented):
Also sorry for copilot jumping into review, seems like GH automatically enabled ai reviews for our repos :(

lk2nd/device/dts/msm8952/qm215-mtp.dts
CE1CECL: I'm using it for detection but there are actually 2 displays (the lid closed one and the open lid one)
This only covers the main / open lid one since the closed lid display isn't enabled/used until the phone boots android, except to show the splash screen at first.
Anyways, how do I tell? Device Info HW doesn't show much about it.
(https://deviceinfohw.ru/devices/item.php?item=123427)

lk2nd/device/dts/msm8952/qm215-mtp.dts
CE1CECL: Added

lk2nd/device/dts/msm8952/qm215-mtp.dts
CE1CECL: Added to Documentation/devices.md along with its codename in `(` `)` 's

lk2nd/device/dts/msm8952/qm215-mtp.dts
CE1CECL: Fixed

lk2nd/device/dts/msm8952/qm215-mtp.dts
CE1CECL: So, the non-touch panel (lid) is actually SPI, but the main touchscreen display is DSI/MIPI, at least according to sysfs

CE1CECL: Forgot to add, but its the st7789v2_qvga_cmd, I tried to add it but lk2nd doesn't pick it up, it stays on the original splash screen, unless dual-display isn't supported?

TravMurav: I won't expect lk2nd to support running multiple screens, yeah. It's probably fine to ignore the second screen as long as the primary one works

lk2nd/device/dts/msm8952/qm215-mtp.dts
```diff
@@ -0,0 +1,28 @@
...
+
+		// FIXME: lk2nd,dtb-files = "qm215-cat-s22flip";
+
+		lk2nd,match-panel;
+
+		panel {
+			compatible = "cat,s22flip-panel", "lk2nd,panel";
+
+			qcom,mdss_dsi_jd9161z_boe_ips_video {
+				compatible = "cat,s22flip";
```
TravMurav: One thing left (should've probably mentioned in the last review round, sorry) but since the panel is actually mipi, please change the compatible here to what you'd expect this panel node to have in mainline (see other devices around for usual patterns), obviously no need to think of it too hard as we can change it later, but currently if lk2nd finds "cat,s22flip-panel" in linux dt, it would replace it with "cat,s22flip" which makes no sense.

CE1CECL: Fixed:
It is now:

`compatible = "cat,s22flip-boe-jd9161z";`

cat = The device's manufacture
s22flip = The device's codename
boe = The display manufacture
jd9161z = The display name

CE1CECL: ~~TODO: Fix in `lk2nd/device/dts/msm8909/apq8009w-wtp.dts` as well (for mobvoi-catfish).~~
Edit: Overlooked, sorry, it is actually acceptable `compatible = "mobvoi,rm67162";` (I thought it was catfish-panel like this one, since this DTS is actually based off of that)

TravMurav:
Thanks!

PR #711, opened 2026-06-08
project: lk2nd-msm8960: Limit samsung,jflte boot memory to 64MiB

alex-min:
Limit jflte to a 64 MiB lk2nd boot memory window. This still leaves more than the stock bootloader layout, while avoiding display artifacts seen with the generic msm8960 128 MiB window.

While jflte still boots with 128 Mib, the display artefacts hints that this could be unsafe.

Photo of the problem with 128 Mib: 

Screenshot_20260608_223331_Gallery

### DIFF
--- a/project/lk2nd-msm8960.mk
+++ b/project/lk2nd-msm8960.mk
@@ -7,4 +7,10 @@ DEFINES += ENABLE_KASLRSEED_SUPPORT=1
 # The mmu_section_table doesn't map beyond 0x88000000, which is still plenty.
 LK2ND_BOOT_MEM_SIZE := 0x08000000
 
+# jflte shows display artifacts with the generic 128 MiB boot window.
+# While booting still works, it's safer to keep initramfs/DTB to 64 MiB
+ifeq ($(BOOTLOADER_OUT),samsung-jflte)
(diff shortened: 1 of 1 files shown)


ejona86:
I think the line at the top is the device tree, and the larger square is the initrd. The screen is 1080p, so assuming 3 bytes per pixel that's just under 6 MiB.  We know there is 2 MiB reserved for the ATAGS/FDT, and the distance between the tops of both artifacts is 1/3 of the screen. Was the ramdisk ~1.5 MiB?

It looks like the framebuffer is at `0x87E00000`. That would probably be confirmed by the base printed at "MDP continuous splash detected" in the logs (e.g., gotten via `fastboot oem log; fastboot get_staged log.txt`).

(My comment in project/lk2nd-msm8960.mk about mmu_section_table stopping at 0x88000000 was wrong because BASE_ADDR is 0x80200000. So mmu_section_table and the boot mem actually end at 0x88200000.)

The framebuffer for the flo is at 0x8f000000, MIPI_FB_ADDR is at 0x89000000. If we set LK2ND_BOOT_MEM_SIZE to 0x7000000 (112 MiB) or 0x6000000 (96 MiB) for all msm8960 devices, we think that'd be unlikely to collide?

PR #712, opened 2026-06-09
project: lk2nd-msm8960: Limit boot memory to 124MiB

alex-min:
Limit jflte to a 124 MiB lk2nd boot memory window. This still leaves more than the stock bootloader layout, while avoiding display artifacts seen with the generic msm8960 128 MiB window.

Photo of the problem with 128 Mib:
Screenshot_20260608_223331_Gallery

### DIFF
--- a/project/lk2nd-msm8960.mk
+++ b/project/lk2nd-msm8960.mk
@@ -5,6 +5,7 @@ LK2ND_BUNDLE_DTB ?= bundle.dtb
 DEFINES += ENABLE_KASLRSEED_SUPPORT=1
 # Constrained by MIPI_FB_ADDR at 0x89000000. ramoops@88d00000 for flo and mako.
 # The mmu_section_table doesn't map beyond 0x88000000, which is still plenty.
-LK2ND_BOOT_MEM_SIZE := 0x08000000
+# samsung-jflte constrained to 124 MiB (0x07c00000) as the framebuffer is at 0x87E00000 
+LK2ND_BOOT_MEM_SIZE := 0x07c00000
 
 include lk2nd/project/lk2nd.mk


project/lk2nd-msm8960.mk
```diff
@@ -7,4 +7,10 @@ DEFINES += ENABLE_KASLRSEED_SUPPORT=1
 # The mmu_section_table doesn't map beyond 0x88000000, which is still plenty.
 LK2ND_BOOT_MEM_SIZE := 0x08000000
 
+# jflte shows display artifacts with the generic 128 MiB boot window
+# as it overwrites the framebuffer, 124 MiB is safer
+ifeq ($(BOOTLOADER_OUT),samsung-jflte)
+LK2ND_BOOT_MEM_SIZE := 0x07c00000
```
ejona86, line 9: I'd still prefer you change the value shared for all msm8960. That way a single lk2nd build works for all msm8960. The 4 MiB lost is insignificant, especially compared to most other platforms.

project/lk2nd-msm8960.mk
```diff
@@ -7,4 +7,10 @@ DEFINES += ENABLE_KASLRSEED_SUPPORT=1
 # The mmu_section_table doesn't map beyond 0x88000000, which is still plenty.
 LK2ND_BOOT_MEM_SIZE := 0x08000000
 
+# jflte shows display artifacts with the generic 128 MiB boot window
```
ejona86: Could the comment document that the runtime address of the framebuffer is 0x87E00000?

alex-min:
thanks for the feedback @ejona86 . I think you are right, 124mb and 128mb are mostly the same, I've limited all msm8960 devices to 124mb and added a comment for jflte.

alex-min:
cc @TravMurav maybe as you reviewed my other PR? It seems this file doesn't automatically request people who could review/merge it. I don't know if you can merge @ejona86

ejona86 (approved):

TravMurav:
Thanks!

PR #714, opened 2026-06-15
lk2nd: device:dts : msm8916 :clean up vivo CDP

Tikily:
Reorganize the vivo CDP platform to further refine device granularity
Add panel orise8012a_fwvga_cmd and hx8389b_qhd_video.h


TravMurav:
The changes here make sense, I've merged changes that add the define for the CDP msm-id, would you mind rebasing your change and rearranging a bit, so you add two panels in two separate commits and then rename the dts in the third one.

When you can, would you mind also putting spaces in commit subjects like other commits around, the wording itself is good, just please add one space after `:` and remove spaces before `:` (i.e. `lk2nd: display: generated:`)

PR #718, opened 2026-06-16
fastboot: fix truncated serial

samcday:
My samsung-gt510 has a 16-char serial number. This revealed two issues:

 * target_serialno() is using eMMC PSN, which is only 32bit (8 hex chars)
 * fastboot serialno buffers were too small

### DIFF
--- a/app/aboot/fastboot.c
+++ b/app/aboot/fastboot.c
@@ -665,7 +665,8 @@ static void fastboot_notify(struct udc_gadget *gadget, unsigned event)
 
 int fastboot_init(void *base, unsigned size)
 {
-	char sn_buf[13];
+	extern char sn_buf[MAX_RSP_SIZE];
+	char serialno[MAX_RSP_SIZE] = "";
 	thread_t *thr;
 	dprintf(INFO, "fastboot_init()\n");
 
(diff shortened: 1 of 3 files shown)


app/aboot/fastboot.c
```diff
@@ -676,7 +676,8 @@ int fastboot_init(void *base, unsigned size)
 	target_fastboot_init();
 
 	/* setup serialno */
-	target_serialno((unsigned char *) sn_buf);
+	if (!sn_buf[0])
+		target_serialno((unsigned char *) sn_buf);
```
samcday: IDK this is probably being too clever - maybe better to keep a local `char fb_sn_buf[MAX_RSP_SIZE]` in this function and `strcpy` from the global in `aboot.c`?

TravMurav: This code is weird... In the past it didn't have side effect of updating the global buffer so I guess it was always just generating psn based serial for itself...

I guess if we know that fastboot init happens strictly after the global sn_buffer is initialized, we can just copy or even just access the extern symbol directly...

samcday: > This code is weird... In the past it didn't have side effect of updating the global buffer so I guess it was always just generating psn based serial for itself...

Right.

> I guess if we know that fastboot init happens strictly after the global sn_buffer is initialized, we can just copy or even just access the extern symbol directly...

Hmm, `fastboot_init` only has two callsites but, I'd rather be defensive here: keep a local buffer in this function, try to strcpy from the likely-already-populated `extern char *sn_buf`, and then fallback to psn as last resort. I'll update the PR now

TravMurav:
Thanks!

PR #720, opened 2026-06-19
lk2nd: hw: gpio: pm8921: Add devicetree flag for GPIO polarity inversion

ejona86:
Allow configuring PMIC GPIO interrupt/input polarity inversion dynamically via a custom devicetree flag `GPIO_PMIC_INVERTED`.

On the Asus Nexus 7 (flo), the primary bootloader configures the volume keys with inversion enabled and Linux does not seem to reconfigure this. When lk2nd configured the buttons, it cleared the inversion causing Linux to believe the buttons were pressed.

-------

I'm not beholden to this particular strategy, but it was simple, shows what I'm trying to achieve, and won't break other devices. I played with trying to preserve the inversion flag, and wasn't happy with it. I don't know if we think this is just a linux bug and it should be fixed there instead. I've only seen upstream linux break; I haven't tried with android. I can try out things like that if it helps.

### DIFF
--- a/lk2nd/device/dts/msm8960/bundle.dts
+++ b/lk2nd/device/dts/msm8960/bundle.dts
@@ -14,11 +14,11 @@
 
 			volume-up {
 				lk2nd,code = <KEY_VOLUMEUP>;
-				gpios = <&pmic 3 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+				gpios = <&pmic 3 (GPIO_ACTIVE_HIGH | GPIO_PMIC_INVERTED)>;
 			};
 			volume-down {
 				lk2nd,code = <KEY_VOLUMEDOWN>;
-				gpios = <&pmic 37 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+				gpios = <&pmic 37 (GPIO_ACTIVE_HIGH | GPIO_PMIC_INVERTED)>;
 			};
 		};
 	};
--- a/lk2nd/hw/gpio/pm8921.c
+++ b/lk2nd/hw/gpio/pm8921.c
@@ -18,6 +18,7 @@ int lk2nd_gpio_pmic_config(uint32_t num, int flags)
 		.function	= !!(flags & GPIOL_FLAGS_ASSERTED),
 		.vin_sel	= PMIC_FLAGS_VIN_SEL(flags),
 		.out_strength	= clamp(GPIOL_CONF_DRVSTR(flags), 0, 3),
+		.inv_int_pol	= !!(flags & PMIC_FLAGS_INVERTED),
 	};
 
 	if (flags & GPIO_PULL_UP)
--- a/lk2nd/hw/gpio/supplier.h
+++ b/lk2nd/hw/gpio/supplier.h
@@ -14,6 +14,7 @@
 
 #define PMIC_FLAGS_VIN_SEL(x)		BITS_SHIFT(num, 26, 24)
 #define PMIC_NON_DEFAULT_VIN_SEL	BIT(27)
+#define PMIC_FLAGS_INVERTED		BIT(28)
 
 /* tlmm.c */
 int lk2nd_gpio_tlmm_config(uint32_t num, int flags);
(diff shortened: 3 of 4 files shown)


TravMurav:
I'm generally a bit concerned with this since you seem to say that linux is broken as it doens't properly initialize the hardware (or properly interpret the signal), yet instead of fixing linux, we're trying to hack stuff into lk2nd...

I think it's worth spending a bit more time here to make sure everything is aligned and linux can stand on it's own.

First of all, do you have access to schematics to know how the keys are wired? I.e. I'd assume they are in fact pulled up switches to ground, where active_low flag would make most sense.

Given (I assume) lk2nd currently handles keys properly, maybe the right solution would be changing linux to reset the inversion flag and then fixing the DT to use correct polarity for gpios? Otherwise it sounds like mainline DT for flo is broken as the flag ACTIVE HIGH contradicts what happens IRL where key line is low when the key is pressed...

ejona86:
I don't have the schematics. Testing with a multimeter, the button pressed does connect to ground which would imply ACTIVE_LOW. I'm not really sure how that adds any information, since the inversion provided by pm8921 is logical-only, right? It seems there is no wiring possible that would imply that inversion should be present.

For a decade Linux has left the inversion in [whatever state the bootloader left it](https://github.com/torvalds/linux/blob/e7ae89a0c97ce2b68b0983cd01eda67cf373517d/drivers/pinctrl/qcom/pinctrl-ssbi-gpio.c#L652). So I guess the device trees for all these gpios are functional but invisibilty potentially conceptually broken.

I looked at the downstream flo kernel and [it did clear the inversion](https://github.com/LineageOS/android_kernel_google_msm/blob/0da5107137f93b8d8f697e949ab0784f5ace450c/arch/arm/mach-msm/asustek/flo/board-flo-pmic.c#L158). I didn't find any upstream reasoning [on the patch set](https://lore.kernel.org/lkml/1436942435-25611-3-git-send-email-bjorn.andersson@sonymobile.com/). It seems that would have been the time to have code clear the flag for any gpio being configured.

Looks like making [a new custom devicetree parameter](https://github.com/torvalds/linux/blob/51512e22efe813d8223de27f6fd02a8a48ea2323/drivers/pinctrl/qcom/pinctrl-ssbi-gpio.c#L99) isn't that bad. I could use that to clear the inversion for these two buttons specifically in the device tree and then mark them ACTIVE_LOW. I guess I can see how that goes.

TravMurav:
> I don't have the schematics. Testing with a multimeter, the button pressed does connect to ground which would imply ACTIVE_LOW. I'm not really sure how that adds any information, since the inversion provided by pm8921 is logical-only, right? It seems there is no wiring possible that would imply that inversion should be present.
> 

I'd assume the inversion is just some feature of gpio block in the pmic (perhaps to ""simplify the software""), but otherwise I think if the button is pressed when the line is connected to GND, it must be defined as ACTIVE_LOW. Consider if there was some sensor that said in it's datasheet "The presence of magical bird is communicated by N_BIRD pin driven low", so you'd expect this reflected in DT the same way.

> For a decade Linux has left the inversion in [whatever state the bootloader left it](https://github.com/torvalds/linux/blob/e7ae89a0c97ce2b68b0983cd01eda67cf373517d/drivers/pinctrl/qcom/pinctrl-ssbi-gpio.c#L652). So I guess the device trees for all these gpios are functional but invisibilty potentially conceptually broken.
> 

Don't know what Bjorn wanted to do there but IMO relying on undefined reset state is a bug, and having DT not represent the real line state is also a bug (as I wrote above).

> I looked at the downstream flo kernel and [it did clear the inversion](https://github.com/LineageOS/android_kernel_google_msm/blob/0da5107137f93b8d8f697e949ab0784f5ace450c/arch/arm/mach-msm/asustek/flo/board-flo-pmic.c#L158). I didn't find any upstream reasoning [on the patch set](https://lore.kernel.org/lkml/1436942435-25611-3-git-send-email-bjorn.andersson@sonymobile.com/). It seems that would have been the time to have code clear the flag for any gpio being configured.
> 
> Looks like making [a new custom devicetree parameter](https://github.com/torvalds/linux/blob/51512e22efe813d8223de27f6fd02a8a48ea2323/drivers/pinctrl/qcom/pinctrl-ssbi-gpio.c#L99) isn't that bad. I could use that to clear the inversion for these two buttons specifically in the device tree and then mark them ACTIVE_LOW. I guess I can see how that goes.

That sounds reasonable I guess, perhaps ideally driver would've always reset the inversion, but if you add the prop and update the flo dt with it, you keep "backwards compatibility" with old broken dt (in which case, booting it with lk2nd will stil be broken but I guess it worked because people were booting it without lk2nd fine)

ejona86:
I've got a change to Linux working. I'll work on upstreaming it instead of changing lk2nd. Thank you!

PR #721, opened 2026-06-21
dts: msm8916: Clean up Vivo CDP

Tikily:

PR #708, opened 2026-06-07
dts: msm8916: Add support for Vivo Y927 (PD1410V)

zhangzqs:
## Changes

- **dts**: Merge Vivo PD1304 and Y927 (PD1410V) into `msm8916-vivo.dts` with subnodes
- **panel**: Add HX8394A 720p video mode panel driver

## Verification

Boot tested on Vivo Y927 (PD1410V).

### lk2nd

```bash
make -j$(nproc) TOOLCHAIN_PREFIX=arm-none-eabi- LK2ND_QCDTBS="msm8916-vivo.dtb" lk2nd-msm8916
```

### lk1st

```bash
make -j$(nproc) TOOLCHAIN_PREFIX=arm-none-eabi- lk1st-msm8916 \
    LK2ND_BUNDLE_DTB=msm8916-vivo.dtb \
    LK2ND_COMPATIBLE=vivo,pd1410v \
    LK2ND_DISPLAY=hx8394a_720p_video
```

PR #722, opened 2026-06-23
lk2nd: display: panel: generated: Add hx8394a_720p_video for vivo,pd1410

zhangzqs:
The `vivo-pd1410` device (Vivo Y27L / Y627 / Y927) added in #721 uses the `hx8394a_720p_video` panel, but the generated panel driver was not included alongside the device node. As a result, the lk1st build command documented in `msm8916-vivo-cdp-1.dts` for this device fails to link:

```
make -j$(nproc) TOOLCHAIN_PREFIX=arm-none-eabi- lk1st-msm8916 \
    LK2ND_DISPLAY=hx8394a_720p_video \
    LK2ND_BUNDLE_DTB=msm8916-vivo-cdp-1.dtb \
    LK2ND_COMPATIBLE=vivo,pd1410
```

```
undefined reference to `panel_hx8394a_720p_video_select'
```

lk2nd's `oem_panel.c` selects the panel via `panel_select(LK2ND_DISPLAY)`, which expands to `panel_hx8394a_720p_video_select()`. This function is provided by the generated panel header, so the header is required — the gcdb `panel_hx8394a_720p_video.h` is filtered out by `lk2nd/display/panel/rules.mk` and does not participate in the build.

This change adds the generated `lk_panel_hx8394a_720p_video.h` driver and registers it in `panels.h`.

### Verification

Boot-tested on Vivo Y927 (PD1410V). The lk1st build command above now links successfully and produces `emmc_appsboot.mbn`. The panel command sequences and `panel_hx8394a_720p_video_select` are confirmed present in the resulting binary.

### Related

- This is the panel driver part of #708; the `vivo-pd1410` device node itself was already merged via #721.
- Fixes the build regression introduced by #721.

### DIFF
--- a/lk2nd/display/panel/generated/lk_panel_hx8394a_720p_video.h
+++ b/lk2nd/display/panel/generated/lk_panel_hx8394a_720p_video.h
@@ -1,0 +1,246 @@
+// SPDX-License-Identifier: GPL-2.0-only
+// Copyright (c) 2026 FIXME
+// Generated with linux-mdss-dsi-panel-driver-generator from vendor device tree:
+//   Copyright (c) 2014, The Linux Foundation. All rights reserved. (FIXME)
+
+#ifndef _PANEL_HX8394A_720P_VIDEO_H_
+#define _PANEL_HX8394A_720P_VIDEO_H_
+
+#include <mipi_dsi.h>
+#include <panel_display.h>
+#include <panel.h>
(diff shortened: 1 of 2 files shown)


zhangzqs:
efe9c580da7de0f900e83a7b7fb497c0

TravMurav:
Thanks!

PR #724, opened 2026-06-24
dts: msm8952: Add tree for QiKU 360 N5

dropout-zzz:
QiKU 360 N5 is a mid-range smartphone released only in China

QiKU is probably related to Coolpad (?) clue is it uses same proprietary firmware package format with `.CPB` extension

> **specs**: https://4pda.to/devdb/qiku_n5
> **device page** (official, in chinese): https://qiku.com/product/n5/index.html
> **stock rom DL** (official, DDLs dead, still has version list and checksums): http://bbs.360.cn/forum.php?mod=romproduct

this PR only adds a bare bone device tree needed to boot mainline, but it also seems to boot TWRP and downstream fine

(this may be a horrible idea as LK2nd doesn't have logics to filter DTB by `qiku,*` properties, but QK1605-A01 DTB happens to be the first, that's probably why it works?)

SDcard isn't detected, log says OCR request timed out, need more investigations

[lk2nd_log.txt](https://github.com/user-attachments/files/29285697/lk2nd_log.txt)

### photos

| LK2nd | mainline booting | test points map |
| --- | --- | --- |
| photo | スクリーンショット_20260624_170843 | スクリーンショット_20260616_192418 |

stock LK checks `qiku,pcb-id` and `qiku,project-id`, I copied that from downstream, if they are missing / wrong, some message prints in UART and device refuses to boot:

```
[580] DTB offset is incorrect, kernel image does not have appended DTB
[580] ERROR: Appended Device Tree Blob not found
[590] ERROR: Could not do normal boot. Reverting to fastboot mode.
```

this device has at least two variants, A01 (in this PR) and A02 (which probably has carrier lock or just rebranded version for promotion stuffs), i have not found any hardware differences between them (by comparing decompiled DTBs, the only difference seems to be `qiku,*` properties)

the bootloader is not unlockable but custom ROMs and recoveries can be flashed and booted

(the device has EDL test point and firehose from unpacked firmware package, but stock bootloader banned most fastboot commands)

### DIFF
--- a/Documentation/devices.md
+++ b/Documentation/devices.md
@@ -118,6 +118,7 @@
 - Motorola Moto G5S (montana)
 - Motorola Moto G6 Play (jeter)
 - OPPO A57 (A57) (quirky - see comment in `lk2nd/device/dts/msm8952/msm8940-oppo-a57.dts`)
+- QiKU 360 N5 (qk1605-a01) (quirky - see comment in `lk2nd/device/dts/msm8952/msm8976-360-qk1605-a01.dts`)
 - Redmi 3S (land)
 - Redmi 4 (prada)
 - Redmi 4A (rolex)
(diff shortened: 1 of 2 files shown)


TravMurav:
Thanks!

PR #725, opened 2026-06-24
dts: msm8916: Add support for CMCC A1 (M623C) (Wingtech WT86621)

affggh:
lk2nd runs on Wingtech wt86621
It worked for me!

### DIFF
--- a/lk2nd/device/dts/msm8916/msm8916-qrd-skul.dts
+++ b/lk2nd/device/dts/msm8916/msm8916-qrd-skul.dts
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8916 0>;
+	qcom,board-id = <QCOM_BOARD_ID(QRD, 1, 1) 16>;
+};
+
+&lk2nd {
+	model = "CMCC A1 (M623C) (Wingtech WT86621)";
+	compatible = "wingtech,wt86621";
+
+	lk2nd,dtb-files = "msm8916-wingtech-wt86621";
+
+	panel {
+		compatible = "wingtech,wt86621-panel", "lk2nd,panel";
+
+		qcom,mdss_dsi_ili9881c_dj_720p_video {
+			compatible = "wingtech,ili9881c-dj";
+		};
+		qcom,mdss_dsi_ili9881c_dj_boe_720p_video {
+			compatible = "wingtech,ili9881c-dj-boe";
+		};
+		qcom,mdss_dsi_ili9881c_dj_cmi_720p_video {
+			compatible = "wingtech,ili9881c-dj-cmi";
+		};
+	};
+};
--- a/lk2nd/device/dts/msm8916/rules.mk
+++ b/lk2nd/device/dts/msm8916/rules.mk
@@ -31,6 +31,7 @@ QCDTBS += \
 	$(LOCAL_DIR)/msm8916-qrd-8.dtb \
 	$(LOCAL_DIR)/msm8916-qrd-9.dtb \
 	$(LOCAL_DIR)/msm8916-qrd-skui.dtb \
+	$(LOCAL_DIR)/msm8916-qrd-skul.dtb \
 	$(LOCAL_DIR)/msm8916-qrd-skut1.dtb \
 	$(LOCAL_DIR)/msm8916-samsung.dtb \
 	$(LOCAL_DIR)/msm8916-vivo-cdp-1.dtb \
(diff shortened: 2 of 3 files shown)


lk2nd/device/dts/msm8916/msm8916-qrd-wt86621.dts
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
TravMurav: BSD-3-clause please unless this is imported from somewhere else.

lk2nd/device/dts/msm8916/msm8916-qrd-skul.dts
```diff
@@ -0,0 +1,33 @@
...
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8916 0>;
+	qcom,board-id = <QCOM_BOARD_ID(QRD, 1, 1) 16>;
+};
+
+&lk2nd {
+	model = "CMCC A1 (M623C) (Wingtech WT86621)";
```
TravMurav, line 12: Please add to Documentation/devices.md

lk2nd/device/dts/msm8916/msm8916-qrd-wt86621.dts
```diff
@@ -0,0 +1,33 @@
...
+	compatible = "wingtech,wt86621";
+
+	lk2nd,dtb-files = "msm8916-wingtech-wt86621";
+
+	panel {
+		compatible = "wingtech,wt86621-panel", "lk2nd,panel";
+
+		qcom,mdss_dsi_ili9881c_dj_boe_720p_video {
+			compatible = "wingtech,ili9881c-dj-boe";
+			touchscreen-compatible = "goodix,gt911";
```
TravMurav: Is it necessary to override touchscreen compatible? Are there actually known multi-sourced touchscreens around?

lk2nd/device/dts/msm8916/rules.mk
```diff
@@ -47,5 +47,6 @@ QCDTBS += \
 	$(LOCAL_DIR)/msm8939-qrd-skuk.dtb \
 	$(LOCAL_DIR)/msm8939-qrd-wt82918.dtb \
 	$(LOCAL_DIR)/msm8939-qrd-wt82918hd.dtb \
+	$(LOCAL_DIR)/msm8916-qrd-wt86621.dtb \
```
TravMurav: This is not sorted alphabetically

lk2nd/device/dts/msm8916/msm8916-qrd-wt86621.dts
affggh: Sorry, Now I changed

lk2nd/device/dts/msm8916/rules.mk
affggh: I noticed it named msm8916-qrd-skul in decompiled device tree blob, so I rename it.

lk2nd/device/dts/msm8916/msm8916-qrd-wt86621.dts
affggh: Sorry, only gt9xx in origin dtb blobs.

lk2nd/device/dts/msm8916/rules.mk
TravMurav: I'm a bit confused since i don't see sku L in the caf msm kernel from a quick look, vendors often do weird mess with the naming (i.e. sometimes they just modify existing file, or sometimes they copy them and leave stuff unchanged. Could you share the lines you've used to figure that out?

In any case, if this is a generic dtb, you'd have to find a match property, like in other generic dtbs then, as it'd imply some other oems could've easily copied it the same way...

affggh: Sorry, It seems only wingtech open source kernel have this dts. And seems only this device use, I will change it back to wt86621.

TravMurav: Would you mind sharing a link to that kernel?

affggh: [msm8916-qrd-skul.dts](https://github.com/MiCode/Xiaomi_Kernel_OpenSource/blob/ido-l-oss/arch/arm/boot/dts/qcom/msm8916-qrd-skul.dts)

It seems from Xiaomi, and another lenovo opened source have fully skul.dts and dtsi.

[lenovo k5 source](https://github.com/Jackeagle/kernel_lenovo_k5plus/tree/master/arch/arm/boot/dts/qcom)

They are all from wingtech, so I guess I better rename it back to wt86621.dts

TravMurav: the second source seems to have "skux" use the id=16, so it's probably a mess... Let's just name it as it's model for now but let's try adding device subnode and a match property so it's easier to reconcile multiple devices on this id later, if the need arises

PR #726, opened 2026-06-25
Revert "dts: msm8952: Add tree for QiKU 360 N5"

dropout-zzz:
This reverts #724 , removing the DTS and device entry in documentation I just added

#### what was wrong:

I did not realize `qiku,pcb-id` property in `/` node can also potentially be `0xb` and `0xc` (found in other appended DTBs in downstream `boot.img`), I had hardcoded `0xa` for my device

image

Files in the above image is obtained from:

```shell
magiskboot unpack boot.img
binwalk --extract kernel_dtb  # separate appended DTBs
```

So I should add suffix to file name and tweak documentation accordingly

But I'm still worried about a thing I mentioned in my previous PR, that is LK2nd doesn't have logic to filter DTB by `qiku,*` props. I was never completely sure this is safe

I don't have plan to thoroughly investigate this and I also don't think it's practical to implement a new DTB-selection quirk just for this random device

#### priority / impact:

I don't think any one except me is messing with this device at the moment.

I'm not aware of any actual problem yet

#### final answer:

I'm sorry, it was too rushed. I no longer think I want to have it in this repo, and I'll keep it in my own fork for now. This is the best solution I ended up with so far


TravMurav:
This works on a real device (you've not faked the screenshot did you?), and you've already marked it as quirky requiring single dtb be included in a custom build. I see no reason this causes issues to other devices. If other skus have to be added, they can be added the same way as quirky, and not included in the main build.

Regardless, now i'm confused whether there's a human in the loop too... :(

dropout-zzz:
> This works on a real device (you've not faked the screenshot did you?),

It works on real device.

image

> and you've already marked it as quirky requiring single dtb be included in a custom build. I see no reason this causes issues to other devices. 

I meant I'm not completely sure even on my device there isn't an edge case it can be a bad idea to boot downstream (TWRP or Android) under current state.

> Regardless, now i'm confused whether there's a human in the loop too... :(

I wrote my stuffs by hand..

TravMurav:
> > and you've already marked it as quirky requiring single dtb be included in a custom build. I see no reason this causes issues to other devices.
> 
> I meant I'm not completely sure even on my device there isn't an edge case it can be a bad idea to boot downstream (TWRP or Android) under current state.
> 

Well, for lk2nd we unfortunately can't guarantee it will always be able to boot downstream images, generally you'd want to use OEM bootloader for that (especially on newer devices where bootloader does a lot of more involved modifications to the DT, which we can't feasibly reimplement)

Otherwise it seems like your device works fine with lk2nd when booting mainline, so I suggest we keep it as-is until some issue is found or someone else appears with a similar device, then we can think how to improve things...

> > Regardless, now i'm confused whether there's a human in the loop too... :(
> 
> I wrote my stuffs by hand..

Sorry, the PR description felt being written like a train of thoughts where it feelt like the author changed their mind twice in the span of three paragraphs, which is, y'know... xD Maybe it's just culture/language thing...

Overall, I suggest keeping the device in lk2nd, given we know it works on real hardware

PR #727, opened 2026-06-27
project: add support for arm-none-eabi- toolchain prefix

ylecuyer:
little improvement to add support to the arm-none-eabi- (like the one used in debian) so that we don't have to specify the toolchain prefix every time we build

### DIFF
--- a/makefile
+++ b/makefile
@@ -75,7 +75,7 @@ ifeq ($(ENABLE_EARLY_ETHERNET),1)
 endif
 
 # setup toolchain prefix
-TOOLCHAIN_PREFIX ?= arm-eabi-
+TOOLCHAIN_PREFIX ?= arm-none-eabi-
 CFLAGS += -fstack-protector-all
 CFLAGS += -fno-strict-overflow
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -75,7 +75,14 @@ ifeq ($(ENABLE_EARLY_ETHERNET),1)
 endif
 
 # setup toolchain prefix
-TOOLCHAIN_PREFIX ?= arm-eabi-
+# default to arm-eabi-, fall back to arm-none-eabi- if not found
+ifeq ($(origin TOOLCHAIN_PREFIX), undefined)
+  ifneq ($(shell which arm-eabi-gcc 2>/dev/null),)
```
TravMurav: It's probably a bit annoying to use shell to check user environment. Perhaps we should just change the default to arm-none-eabi- at this point...

ylecuyer: @TravMurav how would you suggest to proceed? should I just do the change? should we give some time for other members of the project to chime in? up to you as I'm new to the project

TravMurav: Please just update this PR

ylecuyer: @TravMurav done

TravMurav:
Thanks!

PR #728, opened 2026-06-27
dts: add the board-id of Vivo Y51A (PD1510)

clinasama:
dts: add the board-id of Vivo Y51A (PD1510)
There is only have the board-id of Vivo Y51L(PD1510) before this push.
I flash the image,but it can't load lk2nd.
So I check the .dtb file of Android kernel,and I found the board-id of Vivo Y51A(PD1510).
I compiled it and flash it to my Vivo Y51A(PD1510),it's loaded lk2nd successful.

### DIFF
--- a/lk2nd/device/dts/msm8916/msm8916-vivo-pd1510.dts
+++ b/lk2nd/device/dts/msm8916/msm8916-vivo-pd1510.dts
@@ -5,7 +5,8 @@
 
 / {
 	qcom,msm-id = <QCOM_ID_MSM8916 0>;
-	qcom,board-id = <QCOM_BOARD_ID_MTP 0x16>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0x16>,
+                  <0x08 0x12>;
 };
 
 &lk2nd {


lk2nd/device/dts/msm8916/msm8916-vivo-pd1510.dts
```diff
@@ -5,7 +5,8 @@
 
 / {
 	qcom,msm-id = <QCOM_ID_MSM8916 0>;
-	qcom,board-id = <QCOM_BOARD_ID_MTP 0x16>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0x16>,
+                  <0x08 0x12>;
```
TravMurav, line 9: Please use the macro from [qcom,ids.h](github.com/msm8916-mainline/lk2nd/blob/main/lk2nd/include/dt-bindings/arm/qcom%2Cids.h)

The indentation is also broken here, and the commit subject line doesn't conform to the style used in other commits.

TravMurav, line 9: I'm confused, is your device fully compatible with Y51L? Do you know what the difference is?

clinasama: The board-id is different from Vivo Y51A and Vivo Y51L.

TravMurav: But is the hardware different?

clinasama: The hardware is same,but the board-id is different.

PR #731, opened 2026-07-06
dts: msm8952: Merge Medion P10610 and Asus X00i to msm8937-qrd

datenpanne:
This is an update to my previous pull request. Initially, I used a specific 
board ID found in the downstream vendor DTS. However, it turned out that 
this ID was too generic and didn't uniquely identify the device. 

The correct, standardized board ID for this hardware configuration is 
QCOM_BOARD_ID(QRD, 1, 0), which is already used by the Asus Zenfone 4 Max Pro (x00i).

To avoid conflicts and keep the repository clean, I have moved the Medion Lifetab 
P10610 (malata) into the shared `msm8937-qrd.dts` file alongside the Asus device. 
Both boards now share the same DTS file and are properly differentiated at runtime 
using `lk2nd,match-panel` based on their respective panel command lines.


lk2nd/device/dts/msm8952/msm8937-qrd.dts
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8937 0x00>;
```
TravMurav: nit:
```suggestion
	qcom,msm-id = <QCOM_ID_MSM8937 0>;
```

lk2nd/device/dts/msm8952/msm8937-qrd.dts
```diff
@@ -0,0 +1,41 @@
...
+	qcom,board-id = <QCOM_BOARD_ID(QRD, 1, 0) 0>;
```
TravMurav, line 8: I'm confused, you seem to have tested this before and it worked? Could you help me understand what was the issue you had?

TravMurav (commented):
s/Add/Move/ in commit message too please, when you're going to fix it up

lk2nd/device/dts/msm8952/msm8937-qrd.dts
datenpanne: Sorry for the confusion! To answer why I changed the board-id:

In my first attempt, I used `qcom,board-id = <0x0b 0x00>;` from the vendor dump. The tablet did boot fine with it back then because the Asus device wasn't included in the repository yet—it was only merged a few days ago.

The real, factory-coded board-id of my Medion tablet is actually `QCOM_BOARD_ID(QRD, 1, 0)`. I only realized this after doing a fresh build with the recently added Asus X00i included: because both the Medion tablet and the Asus X00i hard-code this exact same specific ID from the factory, lk2nd got confused and mistakenly identified my tablet as the Asus device.

So I changed the ID to the correct `QCOM_BOARD_ID(QRD, 1, 0)` to reflect the actual hardware, deleted the separate DTS files, and combined both devices into the shared `msm8937-qrd.dts`. Now they use the correct ID and are cleanly separated at runtime via `lk2nd,match-panel`.

IMG20260707103442

TravMurav:
Thanks!

PR #733, opened 2026-07-09
dts: msm8953: Add support for Nubia Z11 Mini S (nx549j)

ZhangDongzai:
It works.
19CB1AAAF7CEE95DD1F41C47983169D9


lk2nd/device/dts/msm8953/msm8953-mtp.dts
```diff
@@ -32,4 +32,72 @@
...
+			};
+			qcom,mdss_dsi_tianma_otm1901a_1080p_5p5_video {
+				compatible = "nubia,tianma-otm1901a-video";
+			};
+			qcom,mdss_dsi_sharp_1080p_6p4_video {
+				compatible = "nubia,sharp-video";
+			};
+			qcom,mdss_dsi_samsung_s6e3fa3x01_1080p_5p2_command {
+				compatible = "nubia,samsung-s6e3fa3x01-command";
+			};
```
TravMurav, line 73: that's quite a long list... Seems like there are few 5.2", 5.5" and 6.4" size panels (I'm assuming i.e. 5p2 is glass size)

Your device says 5.2 inch, is this true for the actual device size itself? If so, let's drop all other sizes for now.

Also your device says it uses command mode (_cmd) so maybe we should drop all _video mode panels where _command/_cmd version exists? (maybe worth double checking in lk2nd log that it detects panel being in cmd mode)

ZhangDongzai:
I deleted some needless panels.

TravMurav:
Thanks!

PR #734, opened 2026-07-09
Add basic support for some MDM9607 devices

stephan-gh:
E.g. Quectel EG25-G or SIMCOM SIM7600G-H. See commit log.


target/mdm9607/rules.mk
```diff
@@ -15,7 +15,7 @@ SCRATCH_REGION2_128                     := 0x80000000
 SCRATCH_REGION2_SIZE_128                := 0x02900000  # 41M
 
 KERNEL_REGION                           := 0x80000000
-KERNEL_REGION_SIZE                      := 0x02000000  # 32MB
+KERNEL_REGION_SIZE                      := 0x02900000  # 41MB
```
TravMurav, line 18: Should LK2ND_BOOT_MEM_SIZE be overriden for mdm9607 too?

lk2nd/device/dts/mdm9607/mdm9607-mtp.dts
```diff
@@ -0,0 +1,11 @@
...
+
+#include <skeleton32.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MDM9607 0>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0>;
+};
+
+/* TODO: Add &lk2nd device nodes */
```
TravMurav: nit: I'd just not have any comment here for now.

lk2nd/device/dts/mdm9607/rules.mk
```diff
@@ -0,0 +1,5 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
TravMurav, line 1: nit for commit message: we usually have `dts: mdm9607: Add blah blah`. Debatable whether adding the whole new makefile warrants an explicit prefix though. Your call here.

target/mdm9607/rules.mk
stephan-gh: I would leave that for later, since MDM9607 doesn't have lk2nd/boot support at the moment (it doesn't support NAND flash and it's probably not a good idea to put a ext2 on NAND anyway :sweat_smile:). Someday I would like to try making SD card boot working, but we need the lk2nd device node for that first to point out the sd-mmc.

lk2nd/device/dts/mdm9607/rules.mk
stephan-gh: I would prefer to have the `lk2nd:` prefix for all changes inside the lk2nd/ folder, so I now did `lk2nd: dts: mdm9607: ` :see_no_evil:

target/mdm9607/init.c
```diff
@@ -236,7 +237,8 @@ int get_target_boot_params(const char *cmdline, const char *part, char **buf)
 	struct ptable *ptable;
 	int system_ptn_index = -1;
 	int modem_ptn_index = -1;
-	uint32_t buflen = strlen(UBI_CMDLINE) + strlen(" root=ubi0:rootfs ubi.mtd=") + sizeof(int) + (9 * sizeof(char)) + sizeof(int) + 1; /* 1 byte for null character*/
+	uint32_t buflen = strlen(UBI_CMDLINE) + strlen(" root=ubi0:rootfs") + 2 * UBI_MTD_SIZE + 1; /* 1 byte for null character*/
```
TravMurav, line 240: This is rewritten so much, is there any point having this a separate commit instead of squashing the two together?

target/mdm9607/init.c
```diff
@@ -263,24 +265,28 @@ int get_target_boot_params(const char *cmdline, const char *part, char **buf)
...
 	/*check if cmdline contains "root=" at the beginning of buffer or
 	* " root=" in the middle of buffer.
 	*/
-	if (((!strncmp(cmdline, "root=", strlen("root="))) ||
-		(strstr(cmdline, " root="))))
-		dprintf(DEBUG, "DEBUG: cmdline has root=\n");
-	else
-		snprintf(*buf+strlen(*buf), buflen, " root=ubi0:rootfs ubi.mtd=%d ubi.mtd=%d", system_ptn_index, modem_ptn_index);
-		/*in success case buf will be freed in the calling function of this*/
+	if (((!strncmp(cmdline, "root=", strlen("root="))) || (strstr(cmdline, " root=")))) {
```
TravMurav, line 274: nit: do we need strncmp here given the search will always be bound by first 5 characters? (by rhs being const with null terminator)

target/mdm9607/init.c
stephan-gh: I wanted to maintain the original authorship.

target/mdm9607/init.c
stephan-gh: I think it's needed to make a "prefix" search? Not sure :sweat_smile:

TravMurav: ah makes sense

TravMurav:
Thanks!

PR #735, opened 2026-07-10
lk2nd: project: Move platform-specific properties out of lk2nd-.mk

stephan-gh:
Properties specific to the platform should not be defined in
project/lk2nd-$(TARGET).mk because those properties apply only to the
lk2nd configuration and not the lk1st configuration. In most cases, the
lk1st and lk2nd configuration should be equivalent.

Also drop redundant enabling of dev/pmic/pm8921 in target/msm660/rules.mk, this is in project/msm8x60-late.mk already.


TravMurav:
Thanks!

PR #739, opened 2026-07-13
lk2nd: dts: Drop redundant lk2nd device compatibles

stephan-gh:
The &lk2nd node is supposed to contain only the actual device compatible, not any fallback compatibles representing the SoC or lk2nd,device from the legacy branch.


TravMurav:
Thanks for doing this cleanup!

PR #740, opened 2026-07-13
lk2nd: dts: msm8916: Merge Samsung DTBs into single DTB

stephan-gh:
Various Samsung devices do not pick the right DTB, which is why we have to
duplicate some device definitions, e.g. in msm8929-samsung.dts. It is hard
to keep track of this, so sometimes adding a new DTB will break booting on
unrelated devices if they suddenly pick the new DTB for some reason.

Let's give up trying to separate Samsung devices into multiple DTB and
build one big DTB that has them all. We can still detect them reliably by
looking at the bootloader command line.

Also see commit log


stephan-gh:
@wonderfulShrineMaidenOfParadise Could you give this a try on some of your Samsung devices? Pre-built is here: https://github.com/msm8916-mainline/lk2nd/actions/runs/29277617270/artifacts/8289978529

Ideally, all of them should still boot correctly, but this change is a bit risky :see_no_evil:

wonderfulShrineMaidenOfParadise:
I would like to test, ideally all of them when I have time.

PR #743, opened 2026-07-15
platform: msm_shared: qpic_nand: Skip writing trailing 0xFF pages for EFS

stephan-gh:
At the moment, attempting to restore a backup of the EFS2 partitions for the modem corrupts it, e.g.

	$ fastboot fetch efs2 efs2.img
	$ fasbtoot flash efs2 efs2.img
	$ fastboot reboot
	qcom-q6v5-mss 4080000.remoteproc: fatal error received:
	    fs_logr.c:114:Ran out of good blocks in log-rgn XY

It looks like the mdoem relies on having erased ("good") blocks to write to. The current flash_write() routine blindly writes all pages, so if there was an erased page with all 0xFF in the input, it is still explicitly written rather than leaving the page erased. To write these pages, the modem would need to erase the whole block (consisting of multiple pages) and re-write all pages in the beginning.

We can't reliably distinguish between an explicitly written all-0xFF page and an erased page in the image, so let's just assume that all trailing 0xFF pages at the end of a block were erased and leave them that way when flashing. Apply this behavior only for the "efs2" partition, since most other partitions (e.g. "boot", "aboot", "sbl1" etc) contain actual file images were explicitly writing all-0xFF pages makes more sense.

This is similar to the "nandwrite --skip-all-ffs" option on Linux, except that we only consider trailing all-0xFF pages in a block. This is unlikely to make much difference in practice, but it seems to be closer to how the modem firmware is managing the EFS partition.

### DIFF
--- a/platform/msm_shared/qpic_nand.c
+++ b/platform/msm_shared/qpic_nand.c
@@ -2105,18 +2105,37 @@ flash_ecc_bch_enabled(void)
 	return (flash.ecc_width == NAND_WITH_4_BIT_ECC)? 0 : 1;
 }
 
+static uint32_t find_num_pages_to_write(const uint8_t *data, uint32_t num_pages, uint32_t wsize)
+{
+	data += num_pages * wsize; /* Go to end of block */
+
+	/* Scan backward page-by-page, checking if any byte is not 0xFF */
+	while (num_pages > 0) {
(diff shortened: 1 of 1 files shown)


TravMurav:
Thanks!

PR #744, opened 2026-07-15
dts: msm8226: msm8926-v2-720p-mtp: fix dtb name for Sony Xperia M2

erikas9987:
In mainline Linux it is a common practice for Sony devices to be named after their platform name along with their device codename. Change `lk2nd,dtb-names` property to reflect this so lk2nd can pick the right dtb for our device.

See also: https://lore.kernel.org/linux-arm-msm/178379938607.163099.16596449114579997016.b4-ty@kernel.org/


TravMurav:
Thanks!

PR #745, opened 2026-07-18
dts: msm8909: add mobvoi-catfish_ext

CE1CECL:


CE1CECL:
This is for the "Mobvoi TicWatch Pro 2020 (catfish_ext/WF12106)" with 1 GB of RAM, compared to the "Mobvoi TicWatch Pro 2018 (catfish/WF12096)" with 512 MB of RAM.

Documentation/devices.md
```diff
@@ -95,7 +95,7 @@
 - Alcatel Idol 4 (6055*)
 - Asus Zenfone 4 Max Pro (x00i)
 - BQ X5 Plus (Longcheer L9360)
-- Cat S22 Flip (S22FLIP)
```
TravMurav, line 98: if you want to refactor some other device, please add the change as a separate commit.

Documentation/devices.md
```diff
@@ -83,7 +83,7 @@
 - Lenovo Yoga Tab 3 8 LTE / WIFI
 - Lenovo Yoga Tab 3 10 LTE / WIFI
 - Lenovo Tab 10 (TB-X103F)
-- Mobvoi TicWatch Pro (WF12096)
```
TravMurav, line 86: oftentimes we put a list of possible models for a specific marketing name, but I guess in this case there's only two possible models so the support is exhaustive now?

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
```diff
@@ -6,23 +6,24 @@
 / {
 	model = "Qualcomm APQ8009W"; /* Required by the oem bootloader. */
 	qcom,msm-id = <QCOM_ID_APQ8009W 0>;
-	qcom,board-id = <QCOM_BOARD_ID_MTP 0x105>;
+	qcom,board-id = <QCOM_BOARD_ID_MTP 0x5>,
+			<QCOM_BOARD_ID_MTP 0x105>;
 };
```
TravMurav, line 11: looking at it deeper, I'm now confused about these ids, do you have some kernel source where those are provided in a generic mtp/wtp dts? I can't quickly find a CAF tree which has those ids defined...

CE1CECL: [mobvoi-catfish](https://github.com/aosp-mirror/kernel_msm/blob/android-msm-catshark-3.18-pie-wear-dr/arch/arm/boot/dts/qcom/catfish.dts) & [mobvoi-catfish_ext](https://github.com/aosp-mirror/kernel_msm/blob/android-msm-catshark-3.18-pie-wear-dr/arch/arm/boot/dts/qcom/catfish_1g_ddr.dts)
The devices are too similar that it just required a DTS modification to add, else its stock bootloader would throw "dts not found". In fact, there isn't any other real way to detect between the two devices.

Documentation/devices.md
CE1CECL: While I can add back the marketing model number, there are actually 3 variants of the Mobvoi TicWatch Pro, being `catshark`, `catfish`, and `catfish_ext`.

Documentation/devices.md
CE1CECL: Commit or PR?

Documentation/devices.md
TravMurav: Is "catshark" supported with the state of lk2nd after this pr? if not, it's probably worth keeping some note on which specific models are supported (so it's clear if someone has "catshark" that it won't work without being added)

CE1CECL: Short answer: No
Long answer: catshark is the 4G LTE (MSM, compared to the APQ) model of catfish, and its DTS can be found [here](https://github.com/aosp-mirror/kernel_msm/blob/android-msm-catshark-3.18-pie-wear-dr/arch/arm/boot/dts/qcom/catshark.dts) 
Based of diff'ing the precompiled boot.img between the 3 models, the most changes are actually between catshark and catfish. catshark most likely would needs it own file. All the 3 devices are very hard to find on eBay, so unless someone happens to have one already (when they do go on sale anyways, its usually very expensive), I'd assume it won't work.

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
TravMurav: seems like neither WTP nor CDP this file references in "model" uses ids 5/105... Do you remember what was the reason the original device was added into the wtp file?

Documentation/devices.md
TravMurav: commit in the same PR is fine

Documentation/devices.md
TravMurav: Let's then keep the two models in the list (either adding other WF number or catfish, catfish_ext if it's hard to define via "official" SKU numbers, just so it's clear catshark is not included yet

lk2nd/device/dts/msm8909/apq8009w-wtp.dts
CE1CECL: I don't remember the reasoning for the wtp naming, I think I was told to rename it to wtp before, but I can't find the exact comment to where that was. Original PR: https://github.com/msm8916-mainline/lk2nd/pull/399

TravMurav: yeah I've looked at the pr too and didn't see the reasoning for it (seemingly it was added like that immediately or even imported from somewhere else)

It feels like we should rename the file to "apq8009w-catfish.dts" to me, would you mind doing that?

wonderfulShrineMaidenOfParadise: `<QCOM_BOARD_ID_MTP 0x5>` is `apq8009w-1gb-wtp`

TravMurav: Ah, yeah, seems like I failed to find it, thanks! Then disregard my suggestion

TravMurav:
Thanks!

PR #748, opened 2026-07-23
ext2: allocate cache-line aligned memory for group descriptor buffer

lilmayofuksu:
LLM explanation:

> The read DMA path requires cache-line-aligned buffers, and that requirement is silently unenforced. sdhci_send_command even documents and asserts it:
> 
> 
> /* Assert if the data buffer is not aligned to cache line size for read operations. */
> if (cmd->trans_mode == SDHCI_READ_MODE)          // sdhci.c:771
>     ASSERT(IS_CACHE_LINE_ALIGNED(cmd->data.data_ptr));
> …except SDHCI_READ_MODE is BIT(4) = 16, while every real read sets cmd.trans_mode = SDHCI_MMC_READ = 1 (mmc_sdhci.c:701 et al). So trans_mode == SDHCI_READ_MODE is never true and the alignment assert is dead code — an unaligned read buffer sails straight past the guard the authors put there to reject it (the post-read invalidate two lines down correctly uses == SDHCI_MMC_READ, so it does run).
> 
> ext2 hands it a plain malloc buffer. malloc(sizeof(gd)*count) → heap_alloc(size, 0) returns a pointer-aligned (≤8B) address, so the group-descriptor buffer's first and last cache lines are shared with neighbouring heap data. The 1 KiB GDT read is a single ADMA transfer into that buffer. With a non-coherent DMA destination, doing line-granular clean/invalidate on a line that's shared with adjacent data is unsafe: the shared boundary line can interleave the neighbour's (dirty, or subsequently re-dirtied) bytes with the DMA'd bytes — the line gets written back over freshly-DMA'd data, or a stale fill masks it. It happened to land on the tail here = the last group descriptor (gd[N-1]) reading back as zeros, which broke any file in the final block group (extlinux.conf). Intermittent, because it depends on heap layout and what touches the neighbouring line during the transfer — a retry with a different allocation address boots fine.

### DIFF
--- a/lib/fs/ext2/ext2.c
+++ b/lib/fs/ext2/ext2.c
@@ -23,7 +23,9 @@
 
 #include <string.h>
 #include <stdlib.h>
+#include <malloc.h>
 #include <debug.h>
+#include <arch/defines.h>
 #include <lib/fs.h>
 #include "ext2_priv.h"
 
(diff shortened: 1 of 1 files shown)


stephan-gh:
Can you also fix the other usage of `bio_read()` inside `ext2_mount()`?

lilmayofuksu:
Fixed

TravMurav:
Thanks!

PR #750, opened 2026-07-26
dts: msm8952: add support for Zebra TC56 (ironman)

leandrofriedrich:
Hello.

This Merge Request is for adding support for the Zebra TC56 device into lk2nd.

This includes:
- device tree
- panel driver

What works:
- building and booting lk2nd
- building and booting lk1st
- panel
- buttons
- USB

Not working (yet):
- booting stock Android

image

### DIFF
--- a/lk2nd/device/dts/msm8952/msm8956-zebra-ironman.dts
+++ b/lk2nd/device/dts/msm8952/msm8956-zebra-ironman.dts
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8956 0>;
+	qcom,board-id = <QCOM_BOARD_ID_DRAGONBOARD 0x27>;
+
+	/* Bootloader requires those cpu nodes, or device crashes (900e) */
+	cpus {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		cpu@100 {
+			device_type = "cpu";
+			reg = <0x100>;
+		};
+
+		cpu@101 {
+			device_type = "cpu";
+			reg = <0x101>;
+		};
+
+		cpu@102 {
+			device_type = "cpu";
+			reg = <0x102>;
+		};
+	};
+};
+
+&lk2nd {
+	model = "Zebra TC56 (ironman)";
+	compatible = "zebra,ironman";
+	lk2nd,match-panel;
+
+	lk2nd,dtb-files = "msm8956-zebra-ironman";
+
+	panel {
+		compatible = "zebra,ironman-panel", "lk2nd,panel";
+
+		qcom,mdss_dsi_tc7x_ili9881c_5p0inch_720p_video {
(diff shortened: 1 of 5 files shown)


lk2nd/device/dts/msm8952/msm8956-zebra-ironman.dts
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8956 0>;
+	qcom,board-id = <0x0a 0x27>;
+
+	/* Bootloader requires those cpu nodes, or device crashes (900e)*/
```
TravMurav: nit:

```suggestion
	/* Bootloader requires those cpu nodes, or device crashes (900e) */
```

lk2nd/device/dts/msm8952/msm8956-zebra-ironman.dts
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <skeleton64.dtsi>
+#include <lk2nd.dtsi>
+
+/ {
+	qcom,msm-id = <QCOM_ID_MSM8956 0>;
+	qcom,board-id = <0x0a 0x27>;
```
TravMurav: uhhhhhhhhh but that's probably a coincidence....

https://github.com/msm8916-mainline/lk2nd/blob/6752fb8abe45e079f13ed203c7198d5a93f965ed/lk2nd/include/dt-bindings/arm/qcom%2Cids.h#L227

lk2nd/display/panel/generated/panels.h
```diff
@@ -19,6 +19,8 @@
 #include "lk_panel_r61308_720p_video.h"
 #include "lk_panel_r69431_720p_video.h"
 #include "lk_panel_truly_qhd_video.h"
+#include "lk_panel_tc7x_ili9881c_5p0inch_720p_video.h"
+
```
TravMurav:
```suggestion
```

let's just add one include line

lk2nd/device/dts/msm8952/msm8956-zebra-ironman.dts
TheMorc: It seemingly reports "Dragon eMMC" as the variant but I no have idea really if I should take this + that as a coincidence :trollface:
We took this board-id value from the original DT tho, so no idea what was Zebra doing there..

lk2nd/display/panel/generated/panels.h
TheMorc: Oversight on my part, this is now resolved + I also alphabetically sorted it into the list next to the other panels

lk2nd/device/dts/msm8952/msm8956-zebra-ironman.dts
leandrofriedrich: changed to be in line with the other devices, thanks for pointing it out (i guess that explains why variant is Dragon)

TravMurav:
Thanks!

PR #753, opened 2026-08-06
device: msm8960: add Samsung Galaxy S4 (SGH-M919)

ThePacketWarden:
The T-Mobile Galaxy S4 (SGH-M919, jfltetmo) is an apq8064 device in the jflte family but had no matching lk2nd node, so it fell through to the unknown device with no keymap. It reports bootloader string M919UVSFQA1; match on M919* and reuse the jflte dtb and PMIC key GPIOs.

Tested on hardware: device is now correctly identified and volume-key navigation works.

### DIFF
--- a/lk2nd/device/dts/msm8960/bundle.dts
+++ b/lk2nd/device/dts/msm8960/bundle.dts
@@ -112,6 +112,29 @@
 			};
 		};
 	};
+	samsung-jfltetmo {
+		model = "Samsung Galaxy S4 (SGH-M919)";
+		compatible = "samsung,jfltetmo", "samsung,jflte";
+		lk2nd,match-bootloader = "M919*";
+		lk2nd,dtb-files = "qcom-apq8064-samsung-jflte";
+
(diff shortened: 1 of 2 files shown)


lk2nd/device/dts/msm8960/bundle.dts
```diff
@@ -112,6 +112,29 @@
 			};
 		};
 	};
+	samsung-jfltetmo {
+		model = "Samsung Galaxy S4 (SGH-M919)";
```
TravMurav, line 116: Please add to the Documentation/devices.md in the same commit

ThePacketWarden: Done, added the device to Documentation/devices.md and amended it into the same commit.

ThePacketWarden:
Rebased and addressed the bundle.dts comment in f56d5d2. Anything else needed here?

TravMurav:
Thanks!

Issue #751, opened 2026-08-02
lk1st will not compile when providing a panel after commit "dts: msm8952: Add suppport for the Zebra TC56 (ironman) device"


sonic011gamer:
When compiling a target such as msm8909 with a panel specified in the build command, it will yield the following error:

> In file included from lk2nd/display/panel/generated/panels.h:21,
>                  from lk2nd/display/panel/oem_panel.c:14:
> lk2nd/display/panel/generated/lk_panel_tc7x_ili9881c_5p0inch_720p_video.h:880:10: error: conflicting types for 'panel_regulator_settings'; have 'uint32_t[]' {aka 'unsigned int[]'}
>   880 | uint32_t panel_regulator_settings[] = {
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~
> In file included from lk2nd/display/panel/oem_panel.c:7:
> target/msm8909/include/target/display.h:56:23: note: previous definition of 'panel_regulator_settings' with type 'const uint32_t[7]' {aka 'const unsigned int[7]'}
>    56 | static const uint32_t panel_regulator_settings[] = {
>       |                       ^~~~~~~~~~~~~~~~~~~~~~~~
> compiling lk2nd/hw/i2c/dt.c
> make[1]: *** [arch/arm/compile.mk:5: build-lk1st-msm8909/lk2nd/display/panel/oem_panel.o] Error 1

The issue lies in how msm8952 expects extern `uint32_t panel_regulator_settings[7];`, whereas other platforms have this directly in their display.h as `static const uint32_t panel_regulator_settings[]`.

The fix would be to remove that declaration from the panel header, and it will work fine, however there should be a fix eventually to handle it properly.

PR #756, opened 2026-08-09
Add panel support for hx8394a used in EE Harrier Mini, enabling lk1st support for the platform

halal-beef:
Compiled and tested with

make TOOLCHAIN_PREFIX=arm-none-eabi- LK2ND_BUNDLE_DTB=msm8916-ee-harrier-mini.dtb LK2ND_DISPLAY=ee_hx8394a_720p_video lk1st-msm8916 -j32

Unfortunately though as of the date of this commit, issue https://github.com/msm8916-mainline/lk2nd/issues/751 does apply and you will need to remove the definition of the variable from the affected panel header.

After that, build completes sucessfully and using qhypstub to boot, works.


TravMurav:
Thanks!

PR #757, opened 2026-08-16
platform: msm_shared: mmc: bound MCLK register write wait

Smankusors:
> [!NOTE]
> This is one of a series of PRs to support booting from microSD on the msm8960 platform.

`mmc_mclk_reg_wr_delay()` can hang indefinitely if the controller core clock is not running. This can happen when probing the microSD slot if its clocks were not enabled by the PBL.

Add a 1 ms timeout to the wait, matching the behavior of `msmsdcc_sync_reg_wr()` in the downstream msm_sdcc driver.

References:
- [Sony msm8960t downstream](https://github.com/LineageOS/android_kernel_sony_msm8960t/blob/93319b1e5aa343ec1c1aabcb028c5e88c7df7c01/drivers/mmc/host/msm_sdcc.c#L534-L556)
- [Samsung msm8660 downstream](https://github.com/LineageOS/android_kernel_samsung_msm8660-common/blob/ee785111146f563e47857939ff5f0fc09be3fbd0/drivers/mmc/host/msm_sdcc.c#L380-L402)

Tested on Sony Xperia SP (MSM8960T). Verified that a disabled core clock hits the timeout and prints the warning instead of hanging, while an enabled clock completes the write without hitting the timeout.

### DIFF
--- a/platform/msm_shared/mmc.c
+++ b/platform/msm_shared/mmc.c
@@ -35,6 +35,7 @@
 #include <platform/clock.h>
 #include <platform/iomap.h>
 #include <platform/timer.h>
+#include <platform.h>
 #include <bits.h>
 #include <../../../app/aboot/bootimg.h>
 
@@ -55,6 +56,8 @@
 
 #define MMC_BOOT_DATA_READ     0
 #define MMC_BOOT_DATA_WRITE    1
+
+#define MMC_BOOT_MCLK_WR_TIMEOUT_MS  1
 
 static unsigned int mmc_boot_data_transfer(unsigned int *data_ptr,
 						unsigned int data_len,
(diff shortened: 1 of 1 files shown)


TravMurav:
> This is one of a series of PRs to support booting from microSD on the msm8960 platform.

Can you prepare a single PR with multiple commits that gets to that goal? I assume you have work done already, and it's probably better to review everything in-context

Smankusors:
> Can you prepare a single PR with multiple commits that gets to that goal? I assume you have work done already, and it's probably better to review everything in-context

Hmm, unfortunately, while I did get the phone to boot from the SD card, the rest of the changes are still in a... let's say "vibe-coded and not yet properly reviewed" state 😅

This is the only patch I've thoroughly reviewed and tested so far. If you'd prefer a single PR with the full set of changes, I'd like to properly review and test the rest first before submitting them.

For now, my plan only goes as far as switching the main storage from eMMC to microSD. (This switch might stay on my local though, since forcing all legacy platforms to use only microSD would be bad idea?)

Ideally, I'd like these legacy platforms to support both eMMC and microSD at the same time, like the newer platforms do. But from what I've looked into so far, that would require much more substantial changes. 😵

TravMurav:
> For now, my plan only goes as far as switching the main storage from eMMC to microSD. (This switch might stay on my local though, since forcing all legacy platforms to use only microSD would be bad idea?)

Ack, I see, this change seems fine to me assuming it helps with whatever hacks you have locally.

PR #758, opened 2026-08-21
dts: msm8952: fix asus x00i

ylecuyer:
Did some cleaning for the duplicated asus x00i

before

image

after

image


TravMurav:
it's weird that it changed the id it was picking I guess, but since you're the person who initially added this device I guess this is exactly the same device and this fix is correct...

Thanks!

PR #760, opened 2026-08-31
dts: msm8960: Add support for SM-G730A

cookie99999:
Tested on hardware, keys and commands work


TravMurav:
Manually pushed as 28a1044b0 to main, thanks!

